PR #1, opened 2017-03-01
Cleanup etcd

ereslibre:
We are using etcd-proxy for all nodes, and we are relying in etcd
discovery to raise an etcd master node or a proxy. This code is no
longer used.


ereslibre:
`etcd-proxy` includes `etcd`. I didn't notice that bit, it probably makes sense to keep `etcd` in this case.

ereslibre:
I would reject this PR, at least for now.

inercia:
Then I will close this PR. Please reopen if you consider it necessary.

PR #3, opened 2017-03-01
CaaSP/MicroOS adjustments

inercia:
Adjustments necessary for running on CaaSP

### Dependencies

Depends on kubic-project/terraform#1

## TODOs:

- [x] Directories
    - [x] `/var/lib/kubelet`
- [x] Packages missing
    - [x] `python-m2crypto`
    - [x] `conntrack-tools`
- [x] Repos for CaaSP

### DIFF
--- a/salt/repositories/containers.repo
+++ b/salt/repositories/containers.repo
@@ -4,4 +4,5 @@ type=rpm-md
 gpgcheck=0
 enabled=1
 autorefresh=1
+# we only support Leap and CaaSP, and CaaSP will just not use this at all...
 baseurl=http://download.opensuse.org/repositories/Virtualization:/containers/openSUSE_Leap_{{ grains['osrelease'] }}/
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -26,6 +26,12 @@ kube_group:       'kube'
 # install the addons (ie, DNS)
 addons:           'false'
 
+paths:
+  ca_dir:         '/etc/pki/trust/anchors'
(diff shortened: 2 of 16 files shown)


inercia:
After some testing, this seems to work.

salt/repositories/containers.repo
```diff
@@ -4,4 +4,9 @@ type=rpm-md
 gpgcheck=0
 enabled=1
 autorefresh=1
-baseurl=http://download.opensuse.org/repositories/Virtualization:/containers/openSUSE_Leap_{{ grains['osrelease'] }}/
\ No newline at end of file
+{% if grains['oscodename'].startswith("openSUSE Leap") %}
+baseurl=http://download.opensuse.org/repositories/Virtualization:/containers/openSUSE_Leap_{{ grains['osrelease'] }}/
+{% else %}
+# not really important: it will not be used
```
ereslibre: I don't know even if it's worth taking this into account. Having deployed this new code with `SLE` happened on the transition period only. I would ignore completely the `SLE` case since we aren't supporting it. I'd leave this whole block as `baseurl=http://download.opensuse.org/repositories/Virtualization:/containers/openSUSE_Leap_{{ grains['osrelease'] }}/`

inercia: Yeah, we could do that... Anyway, I think we should add a case for Tumbleweed at some moment, just in case someone want to try it.

inercia: Fixed

PR #5, opened 2017-03-01
Add some extra names from the DNS domain

inercia:
Include something like `kubernetes.default.svc.cluster.local` in the DNS names in the API server certificate.

### DIFF
--- a/salt/kubernetes-master/init.sls
+++ b/salt/kubernetes-master/init.sls
@@ -13,10 +13,16 @@ include:
   {% do ip_addresses.append("IP: " + extra_ip) %}
 {% endfor %}
 
+# add some extra names the API server could have
 {% set extra_names = ["DNS: " + grains['fqdn']] %}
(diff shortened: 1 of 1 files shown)


salt/kubernetes-master/init.sls
```diff
@@ -13,11 +13,17 @@ include:
...
 {% endfor %}
 
+# add some extra names the API server could have
 {% set extra_names = ["DNS: " + grains['fqdn']] %}
 {% for extra_name in pillar['api_server']['extra_names'] %}
   {% do extra_names.append("DNS: " + extra_name) %}
 {% endfor %}
 
+# add some standard extra names from the DNS domain
+{% if salt['pillar.get']('dns:domain', '')|length > 0 %}
```
ereslibre: I think this test can be simplified as `{% if salt['pillar.get']('dns:domain') %}`

PR #7, opened 2017-03-02
reboot: added configuration for the lock holders

mssola:
This is needed by rebootmgr in order to check the maximum amount on lock
holders.

### DIFF
--- a/salt/reboot/init.sls
+++ b/salt/reboot/init.sls
@@ -1,0 +1,46 @@
+##################################################
+# Configuration for the reboot manager
+##################################################
+
+# `max_holders` contains the maximum number of lock holders for the cluster. It
+# must comply with the optimal cluster size as defined here:
+#   https://coreos.com/etcd/docs/latest/v2/admin_guide.html
+{% set max_holders = pillar['etcd']['masters']|int %}
+{% if max_holders > 1 %}
+  # Note that for odd numbers, in python (2 or 3): 7 // 2 = 3. So we comply with
+  # the optimal size.
+  {% set max_holders = max_holders // 2 %}
+{% endif %}
+
+# Cleanup any previous cluster information on /opensuse.org
+opensuseorg_cleanup:
+  pkg.installed:
+    - name: etcdctl
+  cmd.run:
+    - name: etcdctl
+            rm -r /opensuse.org
+    # ignore failures for this
+    - check_cmd:
+      - /bin/true
+
+# Initialize the `mutex` key as expected by the reboot manager.
+set_max_holders_mutex:
+  pkg.installed:
+    - name: curl
+  cmd.run:
+    - name: curl -L -X PUT
+            http://127.0.0.1:2379/v2/keys/{{ pillar['reboot']['directory'] }}/{{ pillar['reboot']['group'] }}/mutex?prevExist=false
+            -d value="0"
+    - require:
+      - cmd: opensuseorg_cleanup
+
+# Initialize the `data` key, which is JSON data with: the maximum number of
+# holders, and a list of current holders.
+set_max_holders_data:
+  pkg.installed:
+    - name: curl
+  cmd.run:
+    - name: >-
+        curl -L -X PUT http://127.0.0.1:2379/v2/keys/{{ pillar['reboot']['directory'] }}/{{ pillar['reboot']['group'] }}/data?prevExist=false -d value='{ "max":"{{ max_holders }}", "holders":[] }'
+    - require:
+      - cmd: set_max_holders_mutex
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -43,3 +43,13 @@ flannel:
   backend:        'host-gw'
   etcd_key:       '/flannel/network'
   iface:          'eth0'
+
+# Configuration for the reboot manager (https://github.com/SUSE/rebootmgr).
+# notes:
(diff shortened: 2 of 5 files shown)


mssola:
I finally named it `reboot` because I also expect to put there configuration files and such for [rebootmgr](https://github.com/SUSE/rebootmgr).

salt/reboot/init.sls
```diff
@@ -0,0 +1,42 @@
...
+      - cmd: opensuseorg_cleanup
+
+# Initialize the `data` key, which is JSON data with: the maximum number of
+# holders, and a list of current holders.
+set_max_holders_data:
+  pkg.installed:
+    - name: curl
+  cmd.run:
+    - name: >-
+        curl -L -X PUT http://127.0.0.1:2379/v2/keys/opensuse.org/rebootmgr/locks/global/data -d value='{ "max":"{{ max_holders }}", "holders":[] }'
```
ereslibre: If we run the orchestration after the cluster is already up, wouldn't we clearing the `holders` here?

salt/reboot/init.sls
```diff
@@ -0,0 +1,42 @@
...
+  {% set max_holders = max_holders - 1 %}
+{% endif %}
+
+# Cleanup any previous cluster information on /opensuse.org
+opensuseorg_cleanup:
+  pkg.installed:
+    - name: etcdctl
+  cmd.run:
+    - name: etcdctl
+            rm -r /opensuse.org
```
MaximilianMeister: wouldn't `rm -rf` be better here? i think it might fail otherwise if there was no such folder

salt/reboot/init.sls
```diff
@@ -0,0 +1,42 @@
...
+    # ignore failures for this
+    - check_cmd:
+      - /bin/true
```
MaximilianMeister: do we know which failures could happen here?

salt/reboot/init.sls
```diff
@@ -0,0 +1,42 @@
+##################################################
+# Configuration for the reboot manager
+##################################################
+
+# `max_holders` contains the maximum number of lock holders for the cluster.
+{% set max_holders = salt['pillar.get']('etcd:masters', '1')|int %}
+{% if max_holders > 1 %}
+  {% set max_holders = max_holders - 1 %}
```
ereslibre: Max holders actually can be a completely different number based on `etcd:masters`.

- `etcd:masters == 1` (shouldn't happen) -> `max_holders: 1`
- `etcd:masters == 3` -> `max_holders: 1`
- `etcd:masters == 5` -> `max_holders: 2`
- `etcd:masters == 7` -> `max_holders: 3`
- `etcd:masters == 9` -> `max_holders: 4`

This is based on https://coreos.com/etcd/docs/latest/v2/admin_guide.html (Fault Tolerance Table).

salt/reboot/init.sls
ereslibre: It's the `etcdctl rm` :), it has no `-f`.

mssola: It's etcd, it doesn't fail in this case 😉

salt/reboot/init.sls
ereslibre: We fail to connect to etcd, or something related. The idea is that salt doesn't consider it a fail, always consider it a success, even if `etcdctl` returned a non-zero status.

mssola: That's something we also do somewhere else. We really don't care if it fails in this case iirc

salt/reboot/init.sls
```diff
@@ -0,0 +1,42 @@
+##################################################
+# Configuration for the reboot manager
+##################################################
+
+# `max_holders` contains the maximum number of lock holders for the cluster.
+{% set max_holders = salt['pillar.get']('etcd:masters', '1')|int %}
```
ereslibre: By the way, since this is a `sls` file you can do `pillar['etcd']['masters']` directly.

mssola: Just playing safe, I guess. I'll update this

salt/reboot/init.sls
mssola: Ok, let me take a general approach then.

salt/reboot/init.sls
MaximilianMeister: thanks for clarification

salt/reboot/init.sls
MaximilianMeister: i see now, thanks

salt/reboot/init.sls
mssola: That being said, I remember having problems when doing that and applying `|int`. Let me check...

mssola:
I'll squash my commits later.

salt/reboot/init.sls
mssola: Fixed

salt/reboot/init.sls
mssola: Fixed. For the base case we'll set `max_holders: 1`. For the general case we are going for `max_holders // 2`, which should give us the right number.

salt/reboot/init.sls
mssola: Yes we would 😓 Let me check how we can fix this.

ereslibre: I think we can use `compareAndSwap` (https://coreos.com/etcd/docs/latest/v2/api.html) for this.

PR #17, opened 2017-03-15
[WIP] add sles12 pause image

jordimassaguerpla:
not sure if this is the place where it should go .... but let's start the discussion

### DIFF
--- a/pillar/kube-minion.sls
+++ b/pillar/kube-minion.sls
@@ -1,4 +1,4 @@
 mine_functions:
   network.ip_addrs: [eth0]
-
+# infra container to use instead of downloading gcr.io/google_containers/pause
 pod_infra_container_image: sles12/pause:1.0.0
--- a/pillar/kube-minion.sls
+++ b/pillar/kube-minion.sls
@@ -1,2 +1,4 @@
 mine_functions:
   network.ip_addrs: [eth0]
+
+pod_infra_container_image: sles12/pause:1.0.0
--- a/salt/kubernetes-minion/kubelet.jinja
+++ b/salt/kubernetes-minion/kubelet.jinja
@@ -26,9 +26,7 @@ KUBELET_ARGS="\
 {% endif -%}
     --node-ip={{ grains['ip4_interfaces']['eth0'][0] }} \
     --config=/etc/kubernetes/manifests \
-{% if salt['pillar.get']('pod_infra_container_image') -%}
     --pod-infra-container-image={{ pillar['pod_infra_container_image'] }} \
-{% endif -%}
 {% if salt['pillar.get']('infrastructure', 'libvirt') == 'aws' -%}
     --cloud-provider=aws \
 {% endif -%}


pillar/kube-minion.sls
```diff
@@ -1,2 +1,4 @@
 mine_functions:
   network.ip_addrs: [eth0]
+
```
inercia: It would be nice to have a comment here explaining what `pod_infra_container_image` is. :-)

inercia: And `container` and `image` in the same name sounds redundant, right?

jordimassaguerpla:
@inercia : I added a comment.

Answering your other question about the name, this comes from the kubelet parameter "--pod-infra-container-image=" .

If this looks good, I can rebase.

@inercia: Does it look good?

jordimassaguerpla:
@inercia : Regarding the usage

https://github.com/kubic-project/salt/blob/master/salt/kubernetes-minion/kubelet.jinja#L30

inercia:
@jordimassaguerpla if `pod_infra_container_image` was used before this PR, where was it defined?

jordimassaguerpla:
We were not using that parameter, thus we were using the default one, which is
downloading gcr.io/google_containers/pause

pillar/kube-minion.sls
```diff
@@ -1,2 +1,4 @@
 mine_functions:
   network.ip_addrs: [eth0]
+# infra container to use instead of downloading gcr.io/google_containers/pause
+pod_infra_container_image: sles12/pause:1.0.0
```
inercia: A PR with just a variable declaration does not make sense. This variable must be used somewhere, otherwise for me it is a spurious declaration.

mssola:
@inercia @jordimassaguerpla how is this supposed to work in opensuse ?

inercia:
@mssola I guess the underlying OS is not so important, right? I mean that there should not be any problem in running a `sles12/pause:1.0.0` container in OpenSuse...

mssola:
@inercia it should be fine to run this image. However `sles12/pause:1.0.0` is not available on Docker hub no ? I think that this image is pre-installed in caasp images (even though it wasn't on the qcow2 image I downloaded on friday), so opensuse will fail here.

inercia:
I don't know about how we will deal with repos and so on in OpenSUSE. I guess we could not really run any of this on OpenSUSE right now, so maybe this should be part of a wider effort to port everything to OpenSUSE...

mssola:
> so maybe this should be part of a wider effort to port everything to OpenSUSE...

We don't need to port anything to opensuse. It was working on opensuse before this commit :) All I'm saying, is that the `if` statement shouldn't have been removed, and that pillar data should only be set in CaaSP. Then everything will work in opensuse as it did before.

inercia:
@mssola Yeah, you're right, it was working in OpenSUSE, but it should still work as long as that image is available in the Docker hub, and I guess it could be pushed. Anyway, we could add some `{% if ... %}` somewhere...

mssola:
Done in PR #22

PR #18, opened 2017-03-17
fix etcd proxy instance failure on restart

PI-Victor:
~~* adds a custom systemd unit file for etcd with a one line if statement that~~
~~checks to see if both member and proxy folders exist in /var/lib/etcd/ and~~
~~removes the member folder so that etcd members in proxy mode can start on next~~
~~reboot.~~
* adds a drop-in systemd unit file for etcd that adds an `ExecStartPre` clause removing the `member` directory if both `proxy` and `member` directories exist so that it can allow etcd to restart.
This does not affect etcd full members. It's also not possible to automatically
promote a proxy to member https://github.com/coreos/etcd/issues/2869#issuecomment-105607322
so this fix doesn't break any functionality.

~~* removes the extra parameters sent to etcd from the default systemd unit file,~~
~~since etcd picks up all the settings from the salt provisioned~~
~~/etc/sysconfig/etcd file.~~
* overriding `ExecStart` would only be possible by updating `Type=OneShot`. So extra parameters will stay for now.  

This should be considered a temporary fix.

@inercia @ereslibre

### DIFF
--- a/salt/etcd-proxy/init.sls
+++ b/salt/etcd-proxy/init.sls
@@ -1,6 +1,11 @@
 include:
   - etcd
 
+/etc/systemd/system/etcd.service.d/etcd.conf:
+  file.managed:
+    - source: salt://etcd-proxy/etcd.conf
+    - makedirs: True
+
 extend:
   /etc/sysconfig/etcd:
     file.managed:
--- a/salt/etcd-proxy/etcd.conf
+++ b/salt/etcd-proxy/etcd.conf
@@ -1,0 +1,2 @@
+[Service]
+ExecStartPre=/bin/bash -c "if [[ -d /var/lib/etcd/proxy ]] && [[ -d /var/lib/etcd/member ]]; then rm -rf /var/lib/etcd/member; fi"


PI-Victor:
since this doesn't work with caasp images, ignore this PR for now.

inercia:
@PI-Victor Maybe you could investigate the use of [drop-in units](https://coreos.com/os/docs/latest/using-systemd-drop-in-units.html). You would have to create a `/etc/systemd/system/etcd.service.d` directory, copy a file there with the `PreExec` stuff and restart etcd.

ereslibre:
We would still need to request that `/etc/systemd/system/etcd.service.d` is RW if we expect to do this at salt level. Other solution would be to provide this drop-in unit, or a modification of the original unit at package level.

inercia:
Oh, I thought the whole `/etc` was RW... Maybe we could put the _drop-in unit_ in a `etcd-caasp.rpm`, and then force the installation of that RPM in the CaaSP machines somehow...

ereslibre:
Sorry, @inercia is right, `/etc` is writable. I can confirm on the image from CaaSP that I have I can actually create `/etc/systemd/system/etcd.service.d` and everything is fine.

Sorry for the noise.

salt/etcd/init.sls
```diff
@@ -59,6 +59,7 @@ etcd:
       - file: /var/lib/etcd
     - watch:
       - file: /etc/sysconfig/etcd
+      - file: /usr/lib/systemd/system/etcd.service
```
inercia: This is no longer required with the new drop-in unit, right?

salt/etcd-proxy/init.sls
```diff
@@ -1,6 +1,15 @@
...
 
+/etc/systemd/system/etcd.service.d:
+  file :
+    - directory
+
+
+/etc/systemd/system/etcd.service.d/etcd.service:
+  file.managed:
+    - source: salt://etcd-proxy/etcd.service
+
```
inercia: (as a completely ignorable suggestion) I think you could write this in a more compact form:

```
/etc/systemd/system/etcd.service.d/etcd.service:
  file.directory:
    - name: /etc/systemd/system/etcd.service.d
  file.managed:
    - source: salt://etcd-proxy/etcd.service
```

PI-Victor: not ignorable at all, will update. thanks.

PI-Victor:
@inercia i've tested this with `caasp` and it works ok. please merge if everything's ok.

PR #15, opened 2017-03-13
Vars cleanups

inercia:
Some cleanups in the Salt's pillar variables

@ereslibre @jimmykarily Maybe you have to change some things on the velum side...

PR #21, opened 2017-03-20
vars cleanups

jimmykarily:
We reverted [this PR](https://github.com/kubic-project/salt/pull/15) [here](https://github.com/kubic-project/salt/pull/20/) because the end to end tests were breaking. We should find out what broke them and fix it.

### DIFF
--- a/pillar/kube-master.sls
+++ b/pillar/kube-master.sls
@@ -1,6 +1,7 @@
 mine_functions:
   network.ip_addrs: [eth0]
 
-api_server:
-  extra_names: []
-  extra_ips:   []
+api:
+  server:
+    extra_names: []
+    extra_ips:   []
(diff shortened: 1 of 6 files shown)


jimmykarily:
The [original PR](https://github.com/kubic-project/salt/pull/15) also [moved the dashboard pillar value](https://github.com/kubic-project/salt/pull/15/files#diff-1b366d73824e60734ecbabc4b7925343R30) from vars.sls to params.sls. The tests were failing because when we setup the development envorinment of velum, we [override the vars.sls file](https://github.com/kubic-project/velum/blob/master/kubernetes/salt-templates/pillar/vars.sls) in order to point all minions to the dashboard. After removing vars.sls from top.sls file that file is no longer used so the minions try to connect to "dashbord" domain name which is set [here](https://github.com/kubic-project/salt/blob/master/pillar/vars.sls#L4). If moving the pillar value from vars.sls is important we can find an other solution for the development environment (and the end to end tests). If not, lets merge this as it is.

PR #23, opened 2017-03-21
Configure some components arguments from the pillar

mssola:

### DIFF
--- a/salt/docker/init.sls
+++ b/salt/docker/init.sls
@@ -15,16 +15,15 @@ docker:
     - require:
       - pkg: docker
 
-{% set docker_opt = "--iptables=false" %}
-{% set docker_reg = pillar.get('docker_registry_mirror', '') %}
+{% set docker_reg = pillar['docker']['registry_mirror'] %}
 
 /etc/sysconfig/docker:
   file.replace:
     - pattern: '^DOCKER_OPTS.*$'
 {% if docker_reg == '' %}
-    - repl: DOCKER_OPTS="{{ docker_opt }}"
+    - repl: DOCKER_OPTS="{{ pillar['docker']['args'] }}"
 {% else %}
-    - repl: DOCKER_OPTS="{{ docker_opt }} --insecure-registry={{ docker_reg }} --registry-mirror=http://{{ docker_reg }}"
+    - repl: DOCKER_OPTS="{{ pillar['docker']['args'] }} --insecure-registry={{ docker_reg }} --registry-mirror=http://{{ docker_reg }}"
 {% endif %}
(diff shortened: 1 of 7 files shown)


mssola:
The default for this flag is 5m, which is unreasonable imho.

inercia:
I would prefer to leave the default values unless you have a strong reason...

mssola:
@inercia waiting 5 minutes to re-schedule a pod from a node it is known by kubernetes to be down is unreasonable I think. I agree that 30 seconds is quite of a magic number. Maybe another way to solve this is to provide variables in the params.sls file for extra arguments for the controller-manager, kubelet, etc. I'm thinking on something like this (in params.sls):

```sls
kubernetes:
  args:
    kubelet: ""
    controller-manager: ""
    ...
```

What do you think ?

inercia:
> What do you think ?

I think it is a good idea to allow users to customize the arguments for those components. However, I would stick with the default set of arguments unless strong reasons exist.

inercia:
In your example, `kubernetes` is probably not a good root for the hierarchy, as everything here is related to kubernetes...

mssola:
> I think it is a good idea to allow users to customize the arguments for those components. However, I would stick with the default set of arguments unless strong reasons exist.

Right, those parameters would be left empty by default, meaning that no extra arguments would be provided.

> In your example, kubernetes is probably not a good root for the hierarchy, as everything here is related to kubernetes...

True 😅

inercia:
>> I think it is a good idea to allow users to customize the arguments for those components. However, I would stick with the default set of arguments unless strong reasons exist.

> Right, those parameters would be left empty by default, meaning that no extra arguments would be provided.

As we are into it, I would also add a variable for extra arguments for the Docker daemon (configured [here](https://github.com/kubic-project/salt/blob/master/salt/docker/init.sls#L18)).

mssola:
@inercia updated. I'll squash commits when it's good to be merged.

salt/docker/init.sls
```diff
@@ -15,16 +15,15 @@ docker:
     - require:
       - pkg: docker
 
-{% set docker_opt = "--iptables=false" %}
 {% set docker_reg = pillar.get('docker_registry_mirror', '') %}
```
inercia: As you are already into this, maybe you could also put `docker_registry_mirror` under `docker` and have 

```
docker:
    # Extra arguments to be passed to the Docker daemon.
    args: '--iptables=false'
    registry_mirror: ''
```

mssola: Sure thing.

mssola:
@inercia done

PR #24, opened 2017-03-27
Fixed the infra container path for CaaSP

mssola:

### DIFF
--- a/salt/kubernetes-minion/kubelet.jinja
+++ b/salt/kubernetes-minion/kubelet.jinja
@@ -26,7 +26,7 @@ KUBELET_ARGS="\
 {% endif -%}
     --node-ip={{ grains['ip4_interfaces']['eth0'][0] }} \
     --config=/etc/kubernetes/manifests \
-{% if salt['pillar.get']('pod_infra_container_image') -%}
+{% if grains['lsb_distrib_id'] == "CAASP" -%}
     --pod-infra-container-image={{ pillar['pod_infra_container_image'] }} \
 {% endif -%}
 {% if salt['pillar.get']('infrastructure', 'libvirt') == 'aws' -%}
--- a/pillar/kube-minion.sls
+++ b/pillar/kube-minion.sls
@@ -1,6 +1,4 @@
 mine_functions:
   network.ip_addrs: [eth0]
-{% if grains['lsb_distrib_id'] == "CAASP" -%}
-# infra container to use instead of downloading gcr.io/google_containers/pause
-pod_infra_container_image: sles12/pause:1.0.0
-{% endif -%}
+  # infra container to use instead of downloading gcr.io/google_containers/pause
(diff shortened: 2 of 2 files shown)


salt/kubernetes-minion/kubelet.jinja
```diff
@@ -26,8 +26,8 @@ KUBELET_ARGS="\
 {% endif -%}
     --node-ip={{ grains['ip4_interfaces']['eth0'][0] }} \
     --config=/etc/kubernetes/manifests \
-{% if salt['pillar.get']('pod_infra_container_image') -%}
-    --pod-infra-container-image={{ pillar['pod_infra_container_image'] }} \
+{% if grains['lsb_distrib_id'] == "CAASP" -%}
+    --pod-infra-container-image=sles12/pause:1.0.0 \
```
inercia: To do the `grains['lsb_distrib_id'] == "CAASP"` check here is good, but why don't we keep the image name in a pillar variable?

mssola:
@inercia I've changed a bit the logic and hopefully now it's like you wanted. Now the value on the pillar is always set, but it's only consumed in caasp. From testing, the previous approach of setting that pillar value only on caasp simply didn't work.

ereslibre:
I am not sure overall about this patch. If we modify our packages for fixing the controller node `--pod-infra-container-image` argument (no salt applies there), then we will also have this fixed on our worker nodes (since the same package is installed).

OTOH, on the development environment we wouldn't even care about it (so no changes would be required in this case).

salt/kubernetes-minion/kubelet.jinja
```diff
@@ -26,7 +26,7 @@ KUBELET_ARGS="\
 {% endif -%}
     --node-ip={{ grains['ip4_interfaces']['eth0'][0] }} \
     --config=/etc/kubernetes/manifests \
-{% if salt['pillar.get']('pod_infra_container_image') -%}
```
inercia: And wasn't your previous solution working? Because I would expect this to work:

`pillar/kube-minion.sls`:
```
{% if grains['lsb_distrib_id'] == "CAASP" -%}
# infra container to use instead of downloading gcr.io/google_containers/pause
pod_infra_container_image: sles12/pause:1.0.0
{% endif -%}
```

`salt/kubernetes-minion/kubelet.jinja`:
```
{% if salt['pillar.get']('pod_infra_container_image', '')|length > 0 -%}
--pod-infra-container-image={{ pillar['pod_infra_container_image'] }} \
{% endif -%}
```

This way we could always overwrite the pause container, with `sles12/pause:1.0.0` as the default image for CaaSP.

mssola: It was not working at least from my side...

inercia:
Ok, let's merge this and go ahead...

@ereslibre The problem with the controller will need to be solved in a different way...

PR #25, opened 2017-03-27
Do not set certificate `CN` if domain was not specified by a grain

ereslibre:
With the YaST installation it's really straightforward to install the dashboard without a domain name. This makes the CA not to fail when there is not a domain name specified, we will just create the CA certificate anyway.

### DIFF
--- a/salt/ca/init.sls
+++ b/salt/ca/init.sls
@@ -54,7 +54,9 @@ salt-minion:
 /etc/pki/ca.crt:
   x509.certificate_managed:
     - signing_private_key: /etc/pki/ca.key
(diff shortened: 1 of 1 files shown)


flavio:
LGTM. Can this have an impact on other parts of the deployment (the lack of a FQDN)?

ereslibre:
This isn't a problem right now on other places. We don't really care about `fqdn == hostname -f`, but more about `domain_name` being blank and setting the `CN` to this (the state will fail to execute in this case -- `fqdn` being the same as `hostname -f` shouldn't be a problem).

PR #27, opened 2017-03-28
Use dashboard name to set the size of the cluster

ereslibre:


jimmykarily:
[This](https://github.com/kubic-project/salt/blob/master/salt/kubernetes-master/init.sls#L125) fails as well.

PR #28, opened 2017-03-28
Make etcd state a requirement for states that need etcd running on lo…

jimmykarily:
…calhost


jimmykarily:
This doesn't seem to solve the issue with etcd. We decided to just enable etcd through activate.sh for now.

PR #30, opened 2017-03-28
Always set `CN`

ereslibre:
Even if no grains are set (because the domain could not be inferred), set the default dns domain from the pillar.

This ensures we also have a value for the `Common Name` value on the
certificate that will be generated for the Certificate Authority.


jordimassaguerpla:
lgtm

PR #31, opened 2017-03-28
Actually use `grains.get` default value

ereslibre:


salt/ca/init.sls
```diff
@@ -54,11 +54,7 @@ salt-minion:
 /etc/pki/ca.crt:
   x509.certificate_managed:
     - signing_private_key: /etc/pki/ca.key
-{% if salt['grains.get']('domain', None) is none %}
-    - CN: {{ pillar['dns']['domain'] }}
-{% else %}
-    - CN: {{ grains['domain'] }}
-{% endif %}
+    - CN: {{ salt['grains.get']('domain', pillar['dns']['domain']) }}
```
inercia: Could it be possible that both values are empty?

ereslibre: No, `pillar['dns']['domain']` shouln't be empty, ever. What I think is that `grains['domain']` instead of being `None` is `""`, and the default value is not applied, thus `CN: ""`, failing again.

ereslibre:
Can you please review again? This will tackle all cases (if `None`, it will be replaced by empty string), then we check for the length (so if originally was empty string, we are also good).

salt/ca/init.sls
```diff
@@ -54,11 +54,12 @@ salt-minion:
...
     - signing_private_key: /etc/pki/ca.key
-{% if salt['grains.get']('domain', None) is none %}
-    - CN: {{ pillar['dns']['domain'] }}
-{% else %}
+{% if salt['grains.get']('domain', '')|length > 0 %}
     - CN: {{ grains['domain'] }}
+{% else %}
+    - CN: {{ pillar['dns']['domain'] }}
 {% endif %}
+    - CN: {{ salt['grains.get']('domain', pillar['dns']['domain']) }}
```
inercia: Isn't this duplicated?

ereslibre: Great catch, thank you! Fixed.

inercia:
What about?

```
{% if salt['grains.get']('domain', '')|length > 0 %}
      - CN: {{ grains['domain'] }}
 {% elif salt['pillar.get']('dns:domain', '')|length > 0 %}
     - CN: {{ pillar['dns']['domain'] }}
 {% else %}
     - CN: "kubernetes"
 {% endif %}
```

ereslibre:
Let's play safe @inercia. Added that, thank you.

salt/ca/init.sls
```diff
@@ -54,10 +54,12 @@ salt-minion:
...
   x509.certificate_managed:
     - signing_private_key: /etc/pki/ca.key
-{% if salt['grains.get']('domain', None) is none %}
+{% if salt['grains.get']('domain', '')|length > 0 %}
+    - CN: {{ grains['domain'] }}
+{% elif salt['pillar.get']('dns:domain', '')|length > 0 %}
     - CN: {{ pillar['dns']['domain'] }}
 {% else %}
-    - CN: {{ grains['domain'] }}
+    - CN: kubernetes
```
mssola: I like @inercia's proposal, but I'd like to make it more obvious that this is a default that we've created (maybe in the future we use another value, and when inspecting an "old" cluster we have this CN are we are clueless about it). So, what about something like: "caasp_default_cn" ?

PR #33, opened 2017-03-29
According to the issue linked, pillar.get can handle unicode characte…

jimmykarily:
…r in pillar values

Since we get those values from the user during setup, they might
contain non-ASCII characters which will make the orchestration break.

https://github.com/saltstack/salt/issues/32975#issuecomment-256018650

### DIFF
--- a/salt/ca/init.sls
+++ b/salt/ca/init.sls
@@ -57,20 +57,20 @@ salt-minion:
 {% if salt['grains.get']('domain', None) is not none %}
     - CN: {{ grains['domain'] }}
 {% endif %}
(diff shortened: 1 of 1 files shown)


MaximilianMeister:
for some reason this ran into a timeout, could you have a look there?

jimmykarily:
@MaximilianMeister I'm getting some errors with this patch on ca minion. They might be related to the timeout. I'm looking into this.

jimmykarily:
This doesn't work. pillar.get() [can't access nested values as salt['pillar.get'] does](https://github.com/saltstack/salt/issues/25005#issuecomment-118113156). Closing this.

PR #36, opened 2017-04-03
prevent non existing repository

### LINKED ISSUE #35: when using casp as an image a non existing repository gets added
it's a mix between opensuse and the casp version:

```
minion0:~ # zypper lr -u
Repository priorities are without effect. All enabled repositories share the same priority.

# | Alias      | Name       | Enabled | GPG Check | Refresh | URI                                                                                    
--+------------+------------+---------+-----------+---------+----------------------------------------------------------------------------------------
1 | containers | containers | Yes     | (  ) No   | Yes     | http://download.opensuse.org/repositories/Virtualization:/containers/openSUSE_Leap_1.0/
```

therefore the comment above https://github.com/kubic-project/salt/blob/master/salt/repositories/containers.repo#L8 is not really true

MaximilianMeister:
Revert "Fix the repo for CaaSP"

This reverts commit 45dc75ece636f7b64bb34aa0879d6d644694c771.

Fixes #35

### DIFF
--- a/salt/repositories/containers.repo
+++ b/salt/repositories/containers.repo
@@ -4,5 +4,4 @@ type=rpm-md
 gpgcheck=0
 enabled=1
 autorefresh=1
-# we only support Leap and CaaSP, and CaaSP will just not use this at all...
 baseurl=http://download.opensuse.org/repositories/Virtualization:/containers/openSUSE_Leap_{{ grains['osrelease'] }}/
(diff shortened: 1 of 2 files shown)


inercia:
Many states depend on `/etc/zypp/repos.d/containers.repo`, so that repo must be present, even when it is not used at all. It is not important that the repo does not make sense as CaaSP does not use repos.

MaximilianMeister:
@inercia still `http://download.opensuse.org/repositories/Virtualization:/containers/openSUSE_Leap_1.0/` on casp needs to be fixed

PR #37, opened 2017-04-04
Bug 1032379 - Must install flanneld on the kubernetes master node

robdaemon:
For `kubectl proxy` to properly proxy to the Kubernetes dashboard, you
must run flanneld and docker on the master node. This will set up the
appropriate host routes.

Additionally, nodes cannot communicate with each other in host-gw mode
in flannel without modifying iptables rules to allow the traffic.
host-gw doesn't scale very well, due to needing to add a route to every
node for every node in the cluster. The iptables rules already allow the
flannel udp encapsulation port through (udp port 8285).

### DIFF
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -43,7 +43,7 @@ etcd:
 
 # the flannel backend ('udp', 'vxlan', 'host-gw', etc)
 flannel:
-  backend:        'host-gw'
+  backend:        'udp'
   etcd_key:       '/flannel/network'
   iface:          'eth0'
 
--- a/salt/top.sls
+++ b/salt/top.sls
@@ -23,6 +23,8 @@ base:
   'roles:kube-master':
     - match: grain
     - kubernetes-master
+    - flannel
+    - docker
     - reboot
   'roles:kube-minion':
     - match: grain


robdaemon:
I tested this by installing a cluster via the kubic-project/terraform project:

```bash
export SALT_DIR=/home/rroland/src/github.com/kubic-project/salt
cd /home/rroland/src/github.com/kubic-project/terraform
./contrib/libvirt/k8s-libvirt.sh apply
export KUBECONFIG=/home/rroland/src/github.com/kubic-project/terraform/cluster-config/kubeconfig
kubectl create -f https://rawgit.com/kubernetes/dashboard/master/src/deploy/kubernetes-dashboard.yaml
kubectl proxy
```

Then visited http://localhost:8001/ui and voila, kube-dash!

pillar/params.sls
```diff
@@ -43,7 +43,7 @@ etcd:
 
 # the flannel backend ('udp', 'vxlan', 'host-gw', etc)
 flannel:
-  backend:        'host-gw'
+  backend:        'udp'
```
inercia: I agree UDP would be a better backend for CaaSP...

inercia: I'm afraid this is the only thing I would like to keep of this PR...

robdaemon:
I'm not ready to give up on the other changes. They're required if we want a good Kubernetes install that can run fluentd and other services properly. They match how upstream clusters are deployed in production environments. We really need them, and a kubelet running with an Unchedulable label on the master. I went through this at HPE and there is no way around it.

robdaemon:
See the CoreOS docs on setting up a master as an example: https://coreos.com/kubernetes/docs/latest/deploy-master.html

salt/top.sls
```diff
@@ -23,6 +23,8 @@ base:
   'roles:kube-master':
     - match: grain
     - kubernetes-master
+    - flannel
+    - docker
```
inercia: And why is docker needed in the master?

inercia:
Ok, this will probably make things easier for running some services, so I guess we could run flannel in the master....

salt/top.sls
robdaemon: We really need to add the kubelet to the master also, since you run fluentd as a DaemonSet. I'll do that in a follow-up. You want to run the kubelet on the master as "Unschedulable" to stop regular workloads from going there.

inercia: We will move to an architecture with kubelet in all the machines at some moment, but please open an issue for discussing this before doing a PR because I'm not sure when that will happen...

PR #38, opened 2017-04-14
Allow the kubelet to run on Kubernetes 1.6

robdaemon:
Removes no longer valid --config flag, matches
kubic-project/terraform#32


PI-Victor:
if you remove this, i'm not sure the manifests will get loaded. i think you want `--pod-manifest-path=` unless we load them somewhere else. @ereslibre  fixed this in velum with https://github.com/kubic-project/velum/pull/114/files

robdaemon:
@PI-Victor okay, just pushed a change to use that flag instead.

PR #40, opened 2017-04-18
Do not assume minion_id is hostname/fqdn

ereslibre:


robdaemon:
Something isn't right, this is outputting the wrong value to the `/etc/kubernetes/kubelet` file.

```
Apr 18 21:20:39 minion0 hyperkube[2684]: E0418 21:20:39.120421    2684 event.go:208] Unable to write event: 'Post https://fqdn:6443/api/v1/namespaces/default/events: dial tcp: lookup fqdn on 10.17.3.1:53: no such host' (may retry after sleeping)
```

And the `/etc/kubernetes/kubelet`:

```
# location of the api-server
KUBELET_API_SERVER="--api-servers=https://fqdn:6443"
```

robdaemon:
That being said, the `/etc/sysconfig/etcd` file is correct on my master node.

ereslibre:
Yes, I added [WIP] tag because we need to mine the fqdn, and use the mine. There's no way to get the fqdn grain for a specific node from the jinja template.

ereslibre:
Everything should be fine now. I took advantage of the changes to refactor the pillar a bit, so we don't repeat ourselves too much.

PR #42, opened 2017-04-19
Cleanup

ereslibre:
We have all this code on our git history if we need to get back to it. When doing some relevant changes it's getting more and more difficult see what's relevant and what not.

I think we should go forward with a cleanup now that we released the Beta 2 and now that we will improve our salt states in a significant way because of other priorities like HA and other services.


salt/top.sls
```diff
@@ -12,10 +9,6 @@ base:
   'roles:ca':
     - match: grain
     - ca
-  'roles:etcd':
```
inercia: I would leave the etcd stuff as we will probably need it in the near future...

ereslibre:
Updated

PR #43, opened 2017-04-24
Some docs

inercia:
Some minimal documentation on what we do with Salt.


README.md
```diff
@@ -1,6 +1,32 @@
...
+A collection of Salt states used to provision a kubernetes cluster
+in the context of CaaSP/MicroOS.
+
+# Salt states and CaaSP architecture
+
+The Salt state creates an architecture that depends on some (already existing) elements. This is the expected setup before running the orchestration:
+
+![](docs/k8s-before-orchestration.svg)
+
+The _salt master_ is installed in the _Controller Node_ and is reponsible for
```
ereslibre: I believe the naming is "Administration Dashboard" instead of "Controller Node".

README.md
```diff
@@ -1,6 +1,32 @@
...
+orchestrating the deployment and controlling all the minions in the cluster,
+including a local minion that is used for generating the CA certificate.
+
+A etcd instance is also used for bootstrapping the other etcd instances in
```
ereslibre: An* etcd

README.md
```diff
@@ -1,6 +1,32 @@
...
+the cluster. It will be configured from Salt with the number of etcd masters.
```
ereslibre: It will be configured from Salt with the number of etcd masters set on the pillar.

README.md
```diff
@@ -1,6 +1,32 @@
...
+
+After running the orchestration, the architecture will look like this:
+
+![](docs/k8s-after-orchestration.svg)
+
+Some notes:
+
+* the _etcd discovery_ server is **not** part of the etcd cluster: it is only
```
ereslibre: I would use etcd slang: "The etcd discovery node" is not a **member** of the etcd cluster

inercia: Too many typefaces maybe? And talking about the etcd _node_ sounds like it belongs to a cluster and it is not.

ereslibre: Sorry, it was some kind of diff in comment: "the _etcd discovery_ node is **not** a member of the etcd cluster" is what I actually meant to be written on the document.

PR #44, opened 2017-04-25
Render /etc/hosts on all nodes

grahamhayes:
Render `/etc/hosts` with the machine IDs for all other hosts in the cluster.


salt/hosts/init.sls
```diff
@@ -1,28 +1,13 @@
...
-{{ fqdn }}-host-entry:
+# Internally we refer to hosts by their machine ID, which is also used as the minion ID.
+{%- for minion_id, addrlist in salt['mine.get']('*', 'network.ip_addrs').items() %}
+{{ minion_id }}-host-entry:
   host.present:
 {% if addrlist is string %}
     - ip: {{ addrlist }}
 {% else %}
     - ip: {{ addrlist|first }}
 {% endif %}
```
ereslibre: Maybe this can be simplified by always passing the `addrlist`? If we get several IP addresses in a machine, do we want to add several entries or only the first one?

inercia: I think you need to split this file for two different cases: the master(s) and the minions. The masters could need all the names for all the machines, but the minions don't need to know about each other, only about the api-server.

grahamhayes: It is much easier to have a consistent /etc/hosts across the cluster. Introducing code that does conditional updates of hosts entries based on role is more likely to cause an issue than a large hosts file.

ereslibre: I was thinking that we could do this in subsequent PR's but if you want to tackle it here it's fine for me.

grahamhayes: Only the first IP will be used - /etc/hosts does not support multiple IPs per hostname, so we are OK to just use the first IP

inercia: It is as easy as

1. copying  `salt/hosts/init.sls` to `salt/hosts-master/init.sls` and `salt/hosts-minion/init.sls`, 
2. modifying `top.sls`
```
base:
  '*':
  ...
  'roles:kube-master':
    - match: grain
    - hosts-minion
    ....
  'roles:kube-minion':
    - match: grain
    - hosts-minion
   ...
```

3. and then using this for `salt/hosts-minion/init.sls`:

```
+{%- for master_id, addrlist in salt['mine.get']('role:kube-master', 'network.ip_addrs').items() %}
+{{ master_id }}-host-entry:
   host.present:
 {% if addrlist is string %}
     - ip: {{ addrlist }}
 {% else %}
     - ip: {{ addrlist|first }}
 {% endif %}
    - names:
        - {{ master_id }}
        - api
        - api.{{ pillar['internal_infra_domain'] }}
{% endfor %}

```

grahamhayes: I am still unclear *why* we need to do this.

inercia: Because:

1. we want to use `api.infra.cluster.local` for referring to the API server, so we could move towards solving [this issue](https://trello.com/c/yCbzToby)
2. we will need to react to IP changes for solving [this](https://trello.com/c/a8dfFSkP), and we will need to use two different states unless we want massive updates...
3. it is not such a big change :-)

inercia: I think this is what it currently does: it reads as _if `addrlist` is just a string (not a list), use `addrlist`, otherwise (it is a list), use the first IP in the list_

ereslibre: Yeah, I completely disregarded that you cannot RR with `/etc/hosts`, my fault.

salt/hosts-master/init.sls
```diff
@@ -0,0 +1,26 @@
...
+{% if addrlist is string %}
+    - ip: {{ addrlist }}
+{% else %}
+    - ip: {{ addrlist|first }}
+{% endif %}
+    - names:
+      - {{ minion_id }}
+      - {{ minion_id }}.{{ pillar['internal_infra_domain'] }}
+{% endfor %}
+{%- for minion_id, addrlist in salt['mine.get']('roles:kube-master', 'network.ip_addrs', 'grain').items() %}
```
inercia: I would replace this block by:

```
api-host-entry:
   host.present:
     - ip: 127.0.0.1
     - names:
       - api
       - api.{{ pillar['internal_infra_domain'] }}
       - {{ minion_id }}
       - {{ minion_id }}.{{ pillar['internal_infra_domain'] }}
```

grahamhayes: We will need each master to know about the other masters - I can swap the api to 127.0.0.1

salt/hosts-minion/init.sls
```diff
@@ -0,0 +1,14 @@
...
+    - ip: {{ addrlist }}
+{% else %}
+    - ip: {{ addrlist|first }}
+{% endif %}
+    - names:
+      - api
+      - api.{{ pillar['internal_infra_domain'] }}
+      - {{ minion_id }}
+      - {{ minion_id }}.{{ pillar['internal_infra_domain'] }}
+{% endfor %}
```
inercia: I think we can improve this. We will access the api servers by the `api` alias and never by the hostname (so we don't care about adding entries like `<hash>.<domain>`) and, at the same time, it would be good to point `api.infra.caasp.local` to one of the API servers, but choosing a random one between all the servers available.

I think we can achieve these things with:

```
{% set api_servers = salt['mine.get']('roles:kube-master', 'network.ip_addrs', 'grain').values() %}

api-host-entry:
  host.present:
{% if api_servers is string %}
    - ip: {{ api_servers }}
{% else %}
    - ip: {{ api_servers|first }}
{% endif %}
    - names:
      - api
      - api.{{ pillar['internal_infra_domain'] }}
{% endfor %}```


Lists in Jinja2 are _supposed_ to be _unordered_, so `api_servers|first` should be a random API server...

Comment (grahamhayes) [inline salt/hosts-minion/init.sls]: When we deploy HAProxy to all the nodes for load balancing / HA for the kube-api we will change `api.infra.caasp.local` to `127.0.0.1` and then use the IDs in the HAProxy upstream config, so keeping these entries will avoid us having to make more changes at a later date.

Comment (grahamhayes) [inline salt/hosts-minion/init.sls]: Updated the `top.sls` file

Comment (inercia) [inline salt/hosts-minion/init.sls]: Then keep also the `<hash>.<domain>` entries...

You'll have to use something like:

```
{% set api_servers = salt['mine.get']('roles:kube-master', 'network.ip_addrs', 'grain') %} 
{% set api_servers_addrs = api_servers.values() %}

api-host-entry:
  host.present:
{% if api_servers_addrs is string %}
    - ip: {{ api_servers_addrs }}
{% else %}
    - ip: {{ api_servers_addrs|first }}
{% endif %}
    - names:
      - api
      - api.{{ pillar['internal_infra_domain'] }}
{% endfor %}

{%- for server_id, addrlist in api_servers.items() %}
{{ server_id }}-host-entry:
   host.present:
{% if addrlist is string %}
     - ip: {{ addrlist }}
{% else %}
     - ip: {{ addrlist|first }}
{% endif %}
     - names:
       - {{ server_id }}
       - {{ server_id }}.{{ pillar['internal_infra_domain'] }}
{% endfor %}
```

Comment (grahamhayes) [inline salt/hosts-minion/init.sls]: This will end up with the exact same output in the hosts file, and is needlessly complex. There is no reason to change it

Comment (inercia) [inline salt/hosts-minion/init.sls]: It does not produce the same output in a multi-master environment. Your solution would produce this file (the same in all the workers):

```
10.10.0.2 api api.infra.caasp.local a0848146a8854c519ce698d28901e824 a0848146a8854c519ce698d28901e824.caasp.infra.local
10.10.0.5 api api.infra.caasp.local f9848a36ae854c519ce398d25907e828 f9848a36ae854c519ce398d25907e828.caasp.infra.local
```

while my solution would produce:

```
10.10.0.2 api api.infra.caasp.local
10.10.0.2 a0848146a8854c519ce698d28901e824 a0848146a8854c519ce698d28901e824.caasp.infra.local
10.10.0.5 f9848a36ae854c519ce398d25907e828 f9848a36ae854c519ce398d25907e828.caasp.infra.local
```

with `api.infra.caasp.local` being `10.10.0.2` **or** `10.0.0.5`

Comment (grahamhayes) [inline salt/hosts-minion/init.sls]: OK, but that will not be used - as soon as we add multiple masters we will be moving the `api` host to `127.0.0.1` and this will not be used. This works fine in a single master environment (which we are in right now), and will need an update regardless when we change that.

When we move to multi master, we will remove the 
```
      - api
      - api.{{ pillar['internal_infra_domain'] }}
```

lines and replace it with 

```
api-host-entry:
  host.present:
    - ip: 127.0.0.1
    - names:
      - api
      - api.{{ pillar['internal_infra_domain'] }}
```

inercia: In that case please clearly state in a comment in this file that this cannot be used under any circumstance with a multi-masters environment. I don't want to DoS our own API servers by mistake.

grahamhayes: That is a massive overstatement - this can be used in a multi master cluster, just only using one API server.

If we have enough worker nodes to DoS our API, other things in the cluster are going to fall over first.

This is not a problem, and will not be a problem for the foreseeable future, at which point a lot of how we do the system set will have changed.

inercia: Well, I think that, with this solution, you could DoS your API servers with an scenario like, for example, 7 API servers and 200 workers, and no _"other things in the cluster are going to fall over first"._

Anyway, not a big deal. I will implement this in another PR later on...

inercia:
@grahamhayes please make sure the e2e tests pass... I think they need to ssh to `minion1.k8s.local`, and that cannot be resolved...

grahamhayes:
@inercia I have no idea why this is failing - it looks completely unrelated. Most of the previous PRs for this repo have been merged with failing CI, so I think it is just broken

MaximilianMeister:
> I have no idea why this is failing - it looks completely unrelated. Most of the previous PRs for this repo have been merged with failing CI, so I think it is just broken

but why only for this repo? the velum PR's are passing and they use the salt repo as well... could you maybe run them locally and see if it fails?

inercia:
@MaximilianMeister @grahamhayes the e2e tests depend on doing an `ssh` to `minion1.k8s.local` (I guess this happens in the dashboard), and this PR changes all the hostnames, so it seems related. I don't know much about those e2e tests, so maybe someone could explain why this is happening...

grahamhayes:
@inercia @MaximilianMeister  https://github.com/kubic-project/e2e-tests/blob/master/spec/features/boostrap_cluster.rb#L116 get the master salt id from the salt grains  (not affected by the PR) and then tries find and replace this string with the master IP address. Velum sets this to be `minion1.k8s.local` in the downloaded kubeconfig file.

The failure seems to be pre-existing - whenever the minion ids were changed in salt, this would have stopped working.

Velum either needs to be updated to give the IP, the actual ID, or the tests need to be hardcoded to replace the `minion1.k8s.local` with the IP

ereslibre:
@grahamhayes this should soon be fixed by https://trello.com/c/UKWFZqMD/91-3-add-main-hostname-of-the-apiserver-to-the-pillar. We will provide the possibility to set the public hostname known by the customer of the kubernetes master.

inercia:
Merging regardless of the e2e results... :-(

PR #45, opened 2017-04-25
Use FQDN as kubelet hostname

FedericoCeratto:
Use FQDN as kubelet hostname

### DIFF
--- a/salt/kubernetes-minion/kubelet.jinja
+++ b/salt/kubernetes-minion/kubelet.jinja
@@ -12,8 +12,8 @@ KUBELET_ADDRESS="--address=0.0.0.0"
 # The port for the info server to serve on
 KUBELET_PORT="--port=10250"
 
-# You may leave this blank to use the actual hostname
-KUBELET_HOSTNAME=""
+# Use <machine_id>.<internal_infra_domain> matching the SSL certificates
+KUBELET_HOSTNAME="--hostname-override={{ grains['id'] }}.{{ pillar['internal_infra_domain'] }}"
 
 # location of the api-server
 KUBELET_API_SERVER="--api-servers={{ ','.join(kube_servers) }}"


salt/kubernetes-minion/kubelet.jinja
```diff
@@ -12,8 +12,8 @@ KUBELET_ADDRESS="--address=0.0.0.0"
 # The port for the info server to serve on
 KUBELET_PORT="--port=10250"
 
-# You may leave this blank to use the actual hostname
-KUBELET_HOSTNAME=""
+# Use machine ID as hostname to ensure uniqueness
+KUBELET_HOSTNAME="--hostname-override={{ grains['id'] }}"
```
ereslibre: Isn't the `KUBELET_HOSTNAME` envvar directly the hostname itself? In this case it would be: `KUBELET_HOSTNAME="{{ grains['id'] }}"`

ereslibre: NVM, just saw this usage in the wild...

grahamhayes:
This will be blocked on #44

salt/kubernetes-minion/kubelet.jinja
FedericoCeratto: The /usr/lib/systemd/system/kubelet.service unit is simply concatenating all the variables from /etc/kubernetes/kubelet into parameters for hyperkube. 
E.g.  KUBELET_PORT="--port=10250"

grahamhayes: can we make it `{{ grains['id'] }}.{{ pillar['internal_infra_domain'] }}` ? This means if we ever decide to have a DNS server serve this info, we can easily route traffic based on the domain.

grahamhayes: Based on the updates to https://trello.com/c/yCbzToby/56-8-regenerate-certificates-when-ip-changes - we should make this just `api.{{ pillar['internal_infra_domain'] }}` - #44 has been updated to support this.

PR #46, opened 2017-04-25
React to IP addresses changes

inercia:
React to changes in IP addresses in minions by running an orchestration in the cluster for updating `/etc/hosts`.

It seems the Salt minion loses connectivity to the master when the IP address changes (as reported in saltstack/salt#41578), so the effect of this PR seems to be limited to the _change-IP-after-a-reboot_ case...

### TODO:

- [x] testing

### How to test it

* start the cluster with TF  and run the orchestration
* identify the `dnsmasq` instance that is serving the network with "ps ax | grep dnsmasq" (for example, if your TF file uses the `caasp` prefix then you will see the `caaspnet` somewhere)
    ```
    $ ps ax | grep dnsmasq
    30038 ?        S      0:00 /usr/sbin/dnsmasq --conf-file=/var/lib/libvirt/dnsmasq/caaspnet.conf --leasefile-ro --dhcp-script=/usr/lib64/libvirt/libvirt_leaseshelper
    30039 ?        S      0:00 /usr/sbin/dnsmasq --conf-file=/var/lib/libvirt/dnsmasq/caaspnet.conf --leasefile-ro --dhcp-script=/usr/lib64/libvirt/libvirt_leaseshelper
    ```
* check the config file used by that `dnsmasq` and look for the DHCP file
    ```
    $ sudo cat /var/lib/libvirt/dnsmasq/caaspnet.conf
    ...
    dhcp-hostsfile=/var/lib/libvirt/dnsmasq/caaspnet.hostsfile
    ```
* edit that file (`/var/lib/libvirt/dnsmasq/caaspnet.hostsfile`) and change one of the minion's IP
* `kill -HUP <dnsmasq>`
* reboot that minion
* check that **all** the `/etc/hosts` in the cluster are being updated with the new IP

### DIFF
--- a/reactor/presence.sls
+++ b/reactor/presence.sls
@@ -1,6 +1,8 @@
+# actions to run when a minion disappears or appears in the master
+
 # remove unused keys for minions that are not present anymore
-{% for minion_id in data['lost'] %}
-remove_unused_keys_for_{{ minion_id }}:
-   wheel.key.delete:
-     - match: {{ minion_id }}
-{% endfor %}
+# TODO: we should evaluate what to do here:
+# - we don't want to remove the key for minions that are suffering
+#   some transient connectivity error
+# - but we should remove keys for old minions that will not
+#   connect anymore
--- a/salt/orch/update-etc-hosts.sls
+++ b/salt/orch/update-etc-hosts.sls
@@ -1,0 +1,33 @@
+update_grains:
+  salt.function:
+    - tgt: 'roles:kube-(master|minion)'
+    - tgt_type: grain_pcre
+    - name: saltutil.sync_all
+
+update_mine:
+  salt.function:
+    - tgt: 'roles:kube-(master|minion)'
+    - tgt_type: grain_pcre
+    - name: mine.update
+    - require:
+      - salt: update_grains
+
+etc_hosts_setup:
+  salt.state:
+    - tgt: 'roles:kube-(master|minion)'
+    - tgt_type: grain_pcre
+    - queue: True
+    - sls:
+      - etc-hosts
+    - require:
+      - salt: update_mine
+
+kube_master_setup:
+  salt.state:
+    - tgt: 'roles:kube-master'
+    - tgt_type: grain
+    - queue: True
+    - sls:
+      - kubernetes-master
+    - require:
+      - salt: etc_hosts_setup
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 .idea/
 Makefile
 *.tgz
+
+# you can set a debug log level with "log_level: debug" in this file
+config/master.d/debug.conf
--- a/config/master.d/reactor.conf
+++ b/config/master.d/reactor.conf
@@ -1,3 +1,8 @@
 reactor:
   - 'salt/presence/change':
     - /usr/share/salt/kubernetes/reactor/presence.sls
+    - /usr/share/salt/kubernetes/reactor/etc-hosts.sls
+  - 'salt/beacon/*/network_settings/eth*':
+    - /usr/share/salt/kubernetes/reactor/etc-hosts.sls
+  - 'salt/minion/*/start':
+    - /usr/share/salt/kubernetes/reactor/etc-hosts.sls
--- a/pillar/beacons.sls
+++ b/pillar/beacons.sls
@@ -1,0 +1,7 @@
+beacons:
+  network_settings:
+    interval: 5
+    # monitorize any change in IP addresses in eth0
+    eth0:
+      ipaddr:
+
--- a/pillar/top.sls
+++ b/pillar/top.sls
@@ -6,7 +6,12 @@ base:
     - certificates
     - ip_addrs
     - fqdn
-  'G@roles:kube-master':
+  'roles:kube-(master|minion)':
+    - match: grain_pcre
+    - beacons
+  'roles:kube-master':
+    - match: grain
     - kube-master
-  'G@roles:kube-minion':
+  'roles:kube-minion':
+    - match: grain
     - kube-minion
--- a/reactor/README.md
+++ b/reactor/README.md
@@ -1,0 +1,16 @@
+# Description
+
+Code executed by the [Salt reactor system](https://docs.saltstack.com/en/latest/topics/reactor/).
+Read the official documentation for more details on how it works.
+
+See the [reactor configuration](../config/master.d/reactor.conf) for the list
+of events that trigger actions here.
+
+## Debugging
+
+You can listen for events in the Salt master with:
+
+```
+salt-run state.event pretty=true
+```
+
--- a/reactor/etc-hosts.sls
+++ b/reactor/etc-hosts.sls
@@ -1,0 +1,7 @@
+# actions to run when a host changes/gets/losses connectivity
+
+# update all the /etc/hosts in the cluster
+update_etc_hosts:
+  runner.state.orchestrate:
+    - mods: orch.update-etc-hosts
+
--- a/salt/etcd-proxy/etcd-proxy.conf.jinja
+++ b/salt/etcd-proxy/etcd-proxy.conf.jinja
@@ -6,8 +6,7 @@ ETCD_NAME="{{ grains['id'] }}"
 
 ETCD_LISTEN_CLIENT_URLS="http://0.0.0.0:2379"
 ETCD_LISTEN_PEER_URLS="https://0.0.0.0:2380"
-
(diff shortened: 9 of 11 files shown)


reactor/ip_addr_changed.sls
```diff
@@ -0,0 +1,26 @@
...
+{%- set sender      = data['id'] %}
+
+{% if sender in api_servers %}
+
+# one of the API servers has changed IP:
+# updpate the "<ip> api api.infra.caasp.local" in the minions
+update_hosts_minions:
+  local.state.apply:
+    - tgt: 'G@roles:kube-minion'
+    - expr_form: compound
```
ereslibre: I think `expr_form` can be `grain`, as it's not globbing aside from using grains and tgt could be `roles:kube-minion`?

reactor/ip_addr_changed.sls
```diff
@@ -0,0 +1,26 @@
...
+      - host-minion
+
+{% else %}
+
+# one of the minions has changed IP:
+# updpate the "<ip> <hash> <hash>.infra.caasp.local" in the masters
+update_hosts_masters:
+  local.state.apply:
+    - tgt: 'G@roles:kube-master'
+    - arg:
```
ereslibre: Don't you need to set `expr_form` here?

inercia: Ups, copy&paste'ing is such a really bad thing...

reactor/ip_addr_changed.sls
inercia: I will try both things once we have #44 merged (with `hosts-master` and `hosts-minion`)...

grahamhayes: I used grains in #44 - it worked OK

reactor/ip_addr_changed.sls
```diff
@@ -0,0 +1,27 @@
+{%- set api_servers = salt['pillar.get']('roles:kube-master', '') %}
+{%- set sender      = data['id'] %}
+
+{% if sender in api_servers %}
```
grahamhayes: We should trigger the `hosts-master` in this case as well, as we may have a kubelet on the masters that need to have the api address. In #44 I made the `hosts-master` role render the API address for this reason.

inercia: I think that wouldn't be the right way to solve it: it doesn't make much sense that a local `kubelet` could be talking to a remote API server. And I think the probability of having a working kubelet on the same machine as a dead API server is really low, while the cost of this solution would be that all the `kubelets` in masters _except one_ would be talking to a foreigner API server...

So I think the `/etc/hosts` in the masters should contain just a simple `127.0.0.1 api api.infra.caasp.local` line.

Anyway, we are currently __not__ installing the `kubelet` in the masters...

inercia: See #48 regarding installing the `kubelet` in the masters

grahamhayes: We should still update the list of masters and IDs on the other masters

inercia:
This is blocked until we have a better way to update the Salt code in the CaaSP images...

salt/orch/update-etc-hosts.sls
```diff
@@ -0,0 +1,23 @@
...
+    - tgt_type: grain_pcre
+    - name: mine.update
+    - require:
+      - salt: update_grains
+
+etc_hosts_setup:
+  salt.state:
+    - tgt: 'roles:kube-(master|minion)'
+    - tgt_type: grain_pcre
+    - concurrent: True
```
grahamhayes: this needs to be `queue : True`

salt/orch/update-etc-hosts.sls
```diff
@@ -0,0 +1,23 @@
+update_grains:
+  salt.function:
+    - tgt: 'roles:kube-(master|minion)'
+    - tgt_type: grain_pcre
+    - name: saltutil.refresh_grains
```
grahamhayes: `saltutil.refresh_grains` will not be available until salt version 2016.11.4 - which is not yet available afaik.

reactor/presence.sls
```diff
@@ -1,6 +1,8 @@
+# actions to run when a minion disappears or appears in the master
+
 # remove unused keys for minions that are not present anymore
-{% for minion_id in data['lost'] %}
-remove_unused_keys_for_{{ minion_id }}:
-   wheel.key.delete:
-     - match: {{ minion_id }}
-{% endfor %}
+# {% for minion_id in data['lost'] %}
```
grahamhayes: we should either remove this or uncomment it.

salt/orch/update-etc-hosts.sls
```diff
@@ -0,0 +1,23 @@
...
+    - require:
+      - salt: update_grains
+
+etc_hosts_setup:
+  salt.state:
+    - tgt: 'roles:kube-(master|minion)'
+    - tgt_type: grain_pcre
+    - concurrent: True
+    - sls:
+      - etc-hosts
```
grahamhayes: we also need to add the `kube-apiserver` sls here - https://github.com/kubic-project/salt/blob/f119228e4fad8986b87ed3a932ae19817c087213/salt/kubernetes-master/apiserver.jinja#L24

reactor/presence.sls
inercia: Yeah, probably... I was not sure what do do here. Not sure if we should remove the key when a node is considered as _lost_ from Salt...

salt/orch/update-etc-hosts.sls
inercia: According to [this](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.saltutil.html#salt.modules.saltutil.refresh_grains), _"New in version 2016.3.6,2016.11.4,Nitrogen."_. Doesn't it mean that is available since _2016.3.6_?

salt/orch/update-etc-hosts.sls
inercia: Why are we advertising an IP address? Can't we use a name?

salt/orch/update-etc-hosts.sls
grahamhayes: Yes - it was added to the 2016.3.x branch (after it was forked), and the 2016.11.x branch in 2016.11.4 - if you look at the source in 2016.11.3 that function does not exist.

reactor/presence.sls
grahamhayes: Long Term: We should probably have a UI to remove old node.
Short Term: Auto remove is probably OK - leaving "dead nodes" around will probably cause us issues when doing things like updates

salt/orch/update-etc-hosts.sls
grahamhayes: > --advertise-address ip                                    The IP address on which to advertise the apiserver to members of the cluster. This address must be reachable by the rest of the cluster. If blank, the --bind-address will be used. If --bind-address is unspecified, the host's default interface will be used. 

So, to ensure we are advertising `[eth0]` we need to specify it here.

reactor/presence.sls
inercia: Yeah, but the thing is that a minion that disconnects from the Master for, lets say, 10 minutes, will not be able to join again, and that is not good either...

salt/orch/update-etc-hosts.sls
inercia: Isn't there any alternative to this? I don't think that restarting all the API servers after a Salt minion starts is a good idea... So, do we really need to force the `eth0` address? I think nobody will ever know about master's addresses other than this one, so I'm not sure we really need to advertise it.

grahamhayes: > This address must be reachable by the rest of the cluster

It should only restart if the IP changes on the master - we should have the kube-apiserver watch the apiserver config file on the node.

salt/orch/update-etc-hosts.sls
inercia: Added.

salt/orch/update-etc-hosts.sls
inercia: Yeah, you are right... added.

inercia:
@ereslibre @grahamhayes what do you guys think about this?

salt/orch/update-etc-hosts.sls
grahamhayes: `concurrent: true` needs to go.
We do not want multiple states operating at the same time on the same host.
There is very rarely any reason to use `concurrent: true`

inercia: You are right. I thought it was for running that state concurrently in several minions, but it is on the same minion. Removing this line...

inercia:
Waiting for @ereslibre ;-)

PR #47, opened 2017-04-25
Use a constant name for the API server

inercia:
This should fix [this trello issue](https://trello.com/c/yCbzToby).

### Dependencies

Depends on #44 (rebased on top of it, where we define the local infrastructure domain).

### DIFF
--- a/salt/kubernetes-master/init.sls
+++ b/salt/kubernetes-master/init.sls
@@ -16,7 +16,9 @@ include:
 {% endfor %}
 
 # add some extra names the API server could have
-{% set extra_names = ["DNS: " + grains['fqdn']] %}
+{% set extra_names = ["DNS: " + grains['fqdn'],
+                      "DNS: api",
+                      "DNS: api." + pillar['internal_infra_domain']] %}
 {% for extra_name in pillar['api']['server']['extra_names'] %}
   {% do extra_names.append("DNS: " + extra_name) %}
 {% endfor %}
--- a/salt/kubernetes-minion/kubeconfig.jinja
+++ b/salt/kubernetes-minion/kubeconfig.jinja
@@ -1,13 +1,12 @@
-{% set kube_servers = [] -%}
-{% for fqdn in salt['mine.get']('roles:kube-master', 'fqdn', expr_form='grain').values() -%}
-  {% do kube_servers.append(fqdn) -%}
-{% endfor -%}
+{% set api_server = "api." + pillar['internal_infra_domain']  -%}
+{% set api_ssl_port = salt['pillar.get']('api:ssl_port', '6443') -%}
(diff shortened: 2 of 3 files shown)


salt/kubernetes-minion/kubelet.jinja
```diff
@@ -1,8 +1,7 @@
-{% set kube_servers = [] -%}
+{% set api_server = "api." + salt['pillar.get']('internal_infra_domain','')  -%}
```
ereslibre: Same here...

salt/kubernetes-minion/kubeconfig.jinja
```diff
@@ -1,13 +1,12 @@
-{% set kube_servers = [] -%}
-{% for fqdn in salt['mine.get']('roles:kube-master', 'fqdn', expr_form='grain').values() -%}
-  {% do kube_servers.append(fqdn) -%}
-{% endfor -%}
+{% set api_server = "api." + salt['pillar.get']('internal_infra_domain','')  -%}
```
ereslibre: Not sure about this one, if we happen to fallback to the default value because the pillar value is missing, we end up with `api.`

salt/kubernetes-master/init.sls
```diff
@@ -16,7 +16,9 @@ include:
 {% endfor %}
 
 # add some extra names the API server could have
-{% set extra_names = ["DNS: " + grains['fqdn']] %}
+{% set extra_names = ["DNS: " + grains['fqdn'],
+                      "DNS: api",
```
ereslibre: Do we want to add a raw `api` here? Maybe it makes sense... not sure if we want to ever refer to the apiserver like this.

inercia: I think it doesn't hurt, and it can be nice to be able to `ping api` from the minions...

salt/kubernetes-minion/kubeconfig.jinja
inercia: I know, but I think there is no (easy/clean) way to just fail if the pillar value is empty...

ereslibre: Would it make sense to use `pillar['internal_infra_domain']` and make salt directly fail? (not sure if that would make salt fail right now) -- the idea is to use some mechanism to visibly fail instead of silently set some api server that won't make sense

grahamhayes: In HP Public Cloud many years ago we wrote a macro that did pillar value fallback. I can see if I can dig it out

inercia: Ok, it would be better to use `pillar['something']`

PR #49, opened 2017-04-25
Configure kube-{scheduler/controller-manager} leader elections

robdaemon:
Adds the leader-election flag to kube-scheduler (was already present on
kube-controller-manager)

Also sets the apiserver-count on kube-apiserver

### DIFF
--- a/salt/kubernetes-master/apiserver.jinja
+++ b/salt/kubernetes-master/apiserver.jinja
@@ -22,6 +22,7 @@ KUBE_ADMISSION_CONTROL="--admission-control=NamespaceLifecycle,LimitRanger,Servi
 
 # Add your own!
 KUBE_API_ARGS="--advertise-address={{ grains['ip4_interfaces']['eth0'][0] }} \
+               --apiserver-count={{ salt['mine.get']('roles:kube-master', 'fqdn', expr_form='grain').values()|length }} \
 {% if salt['pillar.get']('infrastructure', 'libvirt') == 'aws' -%}
                --cloud-provider=aws \
 {% endif -%}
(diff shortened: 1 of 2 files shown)


robdaemon:
**do not merge yet** - opening for feedback, since I'm still learning Salt :)

I'm not sure if this would trigger an update when other kube-master nodes are added? I'm going to need some help in testing that.

inercia:
@robdaemon Could we have some context on this? Why are these flags needed?

kiall:
@robdaemon Yes, this should trigger an update when changes are made to the cluster. Though I'd want to check that kubic's use of salt is the same as what I have in my head :)

@inercia https://kubernetes.io/docs/admin/high-availability/ - explains the leader election flag better than I could - "Master elected components" section. For the apiserver-count flag, it's (for whatever reason) not documented on the kubernetes HA page, but it's also essential for running more than 1 kube-apiserver process - https://github.com/kubernetes/kubernetes/issues/24327

Both changes are harmless when there is only 1 master, and required when there is more than 1 - @robdaemon  we should probably target this PR to the `master` rather than `ha` branch?

inercia:
> Yes, this should trigger an update when changes are made to the cluster. Though I'd want to check that kubic's use of salt is the same as what I have in my head, i.e. they they highstate all machines cluster changes, or just the one being change? I assume all, but it's worth checking :)

AFAIK we currently _cannot change the cluster dynamically_, so new machines cannot be added right now... I think that is in our roadmap, but it would require some changes (new machines would have to be shown in the UI, they would have to be accepted, a role assigned, re-orchestrate, etc...)

robdaemon:
@kiall I can retarget this PR. I was erring on the side of staying off master :)

robdaemon:
Verified that this works as expected -

* apiserver-count is 2 when two master nodes are selected
* locks on kube-scheduler and kube-controller-manager are properly held in etcd, and when kube-scheduler or controller-manager are stopped on one node, the lock is released and the other node takes over

This is ready to go.

PR #50, opened 2017-04-26
Update default etcd cluster size to match number of masters

kiall:
Update the etcd role to dynamically choose the cluster size by default, using
the number of masters as the value. This dynamic value may still be overridden
by a pillar value.

We intentially don't make any attempts to prevent "invalid" cluster sizes here,
as doing so would cause issues during master addtion and removal (when that
becomes a suppored operation).

### DIFF
--- a/salt/_modules/etcd_discovery_helpers.py
+++ b/salt/_modules/etcd_discovery_helpers.py
@@ -1,0 +1,43 @@
+from __future__ import absolute_import
+import logging
+
+
+LOG = logging.getLogger(__name__)
+
+
+def __virtual__():
+    return "etcd_discovery_helpers"
+
+
+def get_cluster_size():
+    """
+    Determines the etcd discovery cluster size
+
+    Determines the desired number of cluster members, defaulting to
+    the value supplied in the etcd:masters pillar, falling back to
+    match the number nodes with the kube-master role, and if this is
+    less than 3, it will bump it to 3.
+    """
+    member_count = __pillar__["etcd"]["masters"]
+
+    if member_count is not None:
+        # A value has been set in the pillar, respect the users choice
+        # even it's not a "good" number.
+        member_count = int(member_count)
+
+        if member_count < 3:
+            LOG.warning("etcd member count too low (%d), consider increasing "
+                        "to 3", member_count)
+
+    else:
+        # A value has not been set in the pillar, calculate a "good" number
+        # for the user.
+        member_count = len(__salt__['mine.get'](
+            'roles:kube-master', 'fqdn', expr_form='grain').values())
+
+        if member_count < 3:
+            LOG.warning("etcd member count too low (%d), increasing to 3",
+                        member_count)
+            member_count = 3
+
+    return member_count
--- a/salt/orch/kubernetes.sls
+++ b/salt/orch/kubernetes.sls
@@ -3,6 +3,13 @@ mine_update:
     - name: mine.update
     - tgt: '*'
 
+sync_modules:
+  salt.function:
+    - name: saltutil.sync_modules
+    - tgt: '*'
+    - kwarg:
+        refresh: True
+
 refresh_pillar:
   salt.function:
     - name: saltutil.refresh_pillar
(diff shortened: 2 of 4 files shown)


kiall:
After running a deployment with 2 masters, we now have:

	minion0:~ # etcdctl member list
	3685578e3454c98c: name=b1886fd9783a46d5841cab7ab5636416 peerURLs=https://10.17.3.147:2380,https://minion1.k8s.local:2380 clientURLs=http://10.17.3.147:2379,http://minion1.k8s.local:2379 isLeader=true
	9b9f8b2404816a98: name=af69c76bd28f4792bf957288c14418aa peerURLs=https://10.17.3.163:2380,https://minion0.k8s.local:2380 clientURLs=http://10.17.3.163:2379,http://minion0.k8s.local:2379 isLeader=false

inercia:
Not sure we should do it this way. We need more heuristics for calculating the optimal number of etcd masters.

Maybe we could calculate the number of etcd masters dynamically in the pillar: it seems that pillar variables can get their values [from a renderer](https://docs.saltstack.com/en/latest/ref/renderers/all/salt.renderers.py.html), so you would only need to add a python file with a `run()` function, include that file in the pillar's `top.sls` file and you are done... I have never done this, but it sounds good!

kiall:
@inercia I've updated to use a salt module (salt rendererers are for something else) to encapsulate the logic. I've chosen to not implement any heuristics just yet, but you can probably see they could be slotted in easily enough into the module.

salt/orch/kubernetes.sls
```diff
@@ -3,6 +3,13 @@ mine_update:
     - name: mine.update
     - tgt: '*'
 
+sync_modules:
+  salt.function:
+    - name: saltutil.sync_modules
```
ereslibre: I guess without this the new `etc_discovery_helper` isn't found at all, no?

kiall: Yea, `saltutil.sync_modules` pushes the modules out to the minions's `/var/cache/salt/minion/extmods/` dir where they can be loaded by the minion (and the `refresh=True` forces an an immediate reload of extension modules)

salt/_modules/etcd_discovery_helpers.py
```diff
@@ -0,0 +1,34 @@
...
+    if member_count is not None:
+        member_count = int(member_count)
+    else:
+        member_count = len(__salt__['mine.get'](
+            'roles:kube-master', 'fqdn', expr_form='grain').values())
+
+    if member_count < 3:
+        LOG.warning("etcd member count too low (%d), increasting to 3",
+                    member_count)
+        member_count = 3
```
inercia: Does it make sense to force `member_count = 3` when we have set `pillar["etcd"]["masters"]=1`? Maybe we know what we are doing...

salt/_modules/etcd_discovery_helpers.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+LOG = logging.getLogger(__name__)
+
+
+def __virtual__():
+    return "etcd_discovery_helpers"
+
+
+def get_etcd_cluster_size():
```
inercia: I would rename this function to `get_cluster_size`, as we have to invoke it with  `'etcd_discovery_helpers.get_etcd_cluster_size` and we already know this is related to `etcd`...

kiall: True, will update.

salt/_modules/etcd_discovery_helpers.py
```diff
@@ -0,0 +1,34 @@
...
+    if member_count is not None:
+        member_count = int(member_count)
+    else:
+        member_count = len(__salt__['mine.get'](
+            'roles:kube-master', 'fqdn', expr_form='grain').values())
+
+    if member_count < 3:
+        LOG.warning("etcd member count too low (%d), increasing to 3",
+                    member_count)
+        member_count = 3
```
inercia: Should we force `member_count=3` even if we did set `pillar["etcd"]["masters"]=1`? Maybe we know what we are doing...

kiall: Makes sense, done.

PR #51, opened 2017-04-27
Set hostname on all nodes to match machine ID

grahamhayes:
This pulls in #44 as a dependency - this will go away when it is merged.

### DIFF
--- a/salt/hostname/init.sls
+++ b/salt/hostname/init.sls
@@ -1,0 +1,11 @@
+{% set hostname = grains['id'] + '.' + pillar['internal_infra_domain'] %}
+
+/etc/hostname:
+  file.managed:
+    - contents: {{ hostname }}
+    - backup: false
+
+hostname-static:
+  cmd.run:
+    - name: hostnamectl set-hostname --static --transient {{ hostname }}
+    - unless: [[ $(hostnamectl --transient) == {{ hostname }} ]]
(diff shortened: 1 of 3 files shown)


salt/hostname/init.sls
```diff
@@ -0,0 +1,11 @@
+{% set hostname = grains['id'] + '.' + pillar['internal_infra_domain'] %}
+
+/etc/hostname:
```
inercia: Is this `file.managed` necessary considering that we are also running `hostnamectl --static`? I thought that would also update `/etc/hostname`...

grahamhayes: As it turns out, no, it doesn't. (I thought so too, but it didn't actually do it in my testing)

inercia:
Please see [this code](https://github.com/saltstack-formulas/hostsfile-formula/blob/master/hostsfile/hostname.sls).

grahamhayes:
Sure, I looked at that code, then tested it. That is how I got to this code.

grahamhayes:
That code sets the static hostname, but not the transient hostname, which is used by CaaSP for presentation of the hostname to users.

inercia:
(I should have said that) I'm concerned about this `if`:

```
{%- if grains['os_family'] == 'Suse' %}
/etc/HOSTNAME:
  file.managed:
    - contents: {{ hostname }}
    - backup: false
{% else %}
/etc/hostname:
  file.managed:
    - contents: {{ hostname }}
    - backup: false
{% endif %}
```
Maybe you should modify `/etc/HOSTNAME`... (?)

grahamhayes:
On `caasp` I found that `/etc/HOSTNAME` is symlinked to `/etc/hostname` - I assumed OpenSUSE was the same, but I will check that.

grahamhayes:
From my Leap workstation: 

`lrwxrwxrwx 1 root root 8 Mar 14 17:11 /etc/HOSTNAME -> hostname`

inercia:
Ok, I guess we can assume this will be safe...

PR #53, opened 2017-04-27
Randomize the api.infra.caasp.local IP between all the API servers

inercia:
Our current solution can lead all the workers to the same API server in a multi-master environment. We currently generate something like this `/etc/hosts` file in the minions (the same in all the workers):

```
10.10.0.2 api api.infra.caasp.local a0848146a8854c519ce698d28901e824 a0848146a8854c519ce698d28901e824.caasp.infra.local
10.10.0.5 api api.infra.caasp.local f9848a36ae854c519ce398d25907e828 f9848a36ae854c519ce398d25907e828.caasp.infra.local
```

while it would be better to have something like:

```
10.10.0.2 api api.infra.caasp.local
10.10.0.2 a0848146a8854c519ce698d28901e824 a0848146a8854c519ce698d28901e824.caasp.infra.local
10.10.0.5 f9848a36ae854c519ce398d25907e828 f9848a36ae854c519ce398d25907e828.caasp.infra.local
```

with `api.infra.caasp.local` being `10.10.0.2` **or** `10.0.0.5` randomly on each minion.

### DIFF
--- a/salt/hosts-minion/init.sls
+++ b/salt/hosts-minion/init.sls
@@ -1,14 +1,28 @@
-{%- for minion_id, addrlist in salt['mine.get']('roles:kube-master', 'network.ip_addrs', 'grain').items() %}
-{{ minion_id }}-host-entry:
+{% set api_servers = salt['mine.get']('roles:kube-master', 'network.ip_addrs', 'grain') %}
+{% set api_servers_addrs = api_servers.values() %}
+
+# TODO: remove once we have a haproxy between the kubelet and the API server
+api-host-entry:
   host.present:
-{% if addrlist is string %}
-    - ip: {{ addrlist }}
+{% if api_servers_addrs is string %}
+    - ip: {{ api_servers_addrs }}
 {% else %}
-    - ip: {{ addrlist|first }}
+    - ip: {{ api_servers_addrs|first }}
 {% endif %}
     - names:
       - api
       - api.{{ pillar['internal_infra_domain'] }}
-      - {{ minion_id }}
-      - {{ minion_id }}.{{ pillar['internal_infra_domain'] }}
+{% endfor %}
+
+{%- for server_id, addrlist in api_servers.items() %}
+{{ server_id }}-host-entry:
+   host.present:
+{% if addrlist is string %}
+     - ip: {{ addrlist }}
+{% else %}
+     - ip: {{ addrlist|first }}
+{% endif %}
+     - names:
+       - {{ server_id }}
+       - {{ server_id }}.{{ pillar['internal_infra_domain'] }}
 {% endfor %}


grahamhayes:
This is unneeded. When our team adds support for multiple masters, we will deleting this code.

inercia:
> This is unneeded. When our team adds support for multiple masters, we will deleting this code.

And that will be welcome, but until that I think this Salt repo should produce the best cluster we can, including multi-master deployments (otherwise, we should state that we do not recommend to use this code with multiple masters).

grahamhayes:
This repo does not work with multiple masters properly, so highlighting an individual section seems pointless.

This is also relying on the randomization of python dicts, which can be very not random.

The way the hostfile works is it will take the first entry that matches the request, which means if we are relying on pythons randomization, this will also be random, as the `.items()` will have the hosts in a random order, which will make the IP that nodes get for `api.infra.caasp.local` random across the cluster.

inercia:
> This repo does not work with multiple masters properly, so highlighting an individual section seems pointless.

I would prefer not to have 201 if we already have 200 problems...

> The way the hostfile works is it will take the first entry that matches the request, which means if we are relying on pythons randomization, this will also be random, as the .items() will have the hosts in a random order, which will make the IP that nodes get for api.infra.caasp.local random across the cluster.

I think that Python dictionaries are random by definition, so that shouldn't be a problem. However, I learnt many years ago that resolvers do not follow any _definition_ (in fact, those are really random!), so I don't trust that statement about _"it will take the first entry that matches the request"_... I would prefer to force a random IP address for `api.caasp.infra.local` instead of relying in _"the way the hostfile works"_.

grahamhayes:
Python dicts are random in theory - it is an assumption that does not play out well in the real world.

I would rather not add code that we explicitly know will be removed.

robdaemon:
This isn't a good idea.

As soon as we get ha-proxy in place, we will let ha-proxy handle the shifting among nodes. This isn't reliable, adds extra complexity and should not be done. We're just going to remove this code in a matter of weeks.

:-1: from me.

inercia:
Ok, I have added a big `TODO` so you do not forget this must be removed once we have the `haproxy` (if we finally have it). Until then, even if it happens in a couple of weeks, I think it would be a good idea to have this...

robdaemon:
Your current argument is that this would help in a multi-master scenario, but we do not presently have a multi-master scenario, therefore this code will never execute. Adding dead code that adds complexity is still a :-1:

inercia:
> Your current argument is that this would help in a multi-master scenario, but we do not presently have a multi-master scenario, therefore this code will never execute. Adding dead code that adds complexity is still 

All our code considers that we have multiple masters: we do things like `salt['mine.get']('roles:kube-master', ...)` instead of getting one master from the pillar, so all our code assumes we have multiple masters. And maybe we currently have 201 problems for supporting multi-master, but I would prefer to have 200.

And I know that when we have _haproxy_ this will not be necessary, but maybe that will not happen so soon, or it will be done differently or who knows... You can remove this code by then in a another PR and it will be very welcome.

And regarding the added complexity... well guys, this is not the Paxos algorithm... ;)

robdaemon:
DDoSing the kube-apiserver is not an actual problem. If you do send it too many requests, it will throttle you. You can witness this in the kube-apiserver logs. Nor do we have any condition in these salt scripts where that could occur.

The HA team is currently working on adding haproxy. That is happening as I write this.

This is a solution looking for a problem. We've spent more time discussing why this isn't needed in two separate pull requests than it will actually live in the codebase.

inercia:
I'm afraid throttling has never been a solution for avoiding DoS. But I was more concerned about distributing requests between all the API servers in the cluster...

As you said, I've spent too much time with this discussion and HA is on your plate anyway, so I'm closing this PR.

PR #56, opened 2017-05-02
Update etcd member count logic

kiall:
When etcd discovery is used, and the expected member count is larger
than the available pool of nodes, etcd will fail to start. So, we add
some logic to etcd_discovery_helpers to ensure we don't attempt to
form a cluster that is larger than the available pool of nodes.

### DIFF
--- a/salt/_modules/etcd_discovery_helpers.py
+++ b/salt/_modules/etcd_discovery_helpers.py
@@ -3,6 +3,7 @@ import logging
 
 
 LOG = logging.getLogger(__name__)
+DESIRED_MEMBER_COUNT = 3
 
 
 def __virtual__():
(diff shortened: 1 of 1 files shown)


kiall:
What happens without this patch, and only 2 nodes is:

	May 01 18:28:30 minion1 etcd[2636]: found 2 peer(s), waiting for 1 more
	May 01 18:30:00 minion1 systemd[1]: etcd.service: Start operation timed out. Terminating.
	May 01 18:30:00 minion1 systemd[1]: Failed to start Etcd Server.

grahamhayes:
Tested this (along with https://github.com/kubic-project/e2e-tests/pull/36 ) and e2e tests now pass

salt/_modules/etcd_discovery_helpers.py
```diff
@@ -29,15 +30,22 @@ def get_cluster_size():
...
 
         if member_count < 3:
-            LOG.warning("etcd member count too low (%d), increasing to 3",
-                        member_count)
-            member_count = 3
+            # Attempt to increase the member count to 3, however, if we don't
+            # have 3 nodes in total, then match the number of nodes we have.
+            increased_member_count = len(__salt__['mine.get'](
+                'roles:kube-*', 'fqdn', expr_form='grain').values())
+            increased_member_count = min(3, increased_member_count)
```
inercia: Could it be possible to refactor this `3` to a constant on the top?

kiall: Done

PR #57, opened 2017-05-02
[HA] - Add haproxy

robdaemon:
This adds haproxy as a frontend to the kube-apiserver on all nodes

Runs the kubelet on all nodes

Fixes warnings with the TLS certificates on the etcd nodes

Restarts applicable services when the TLS certificates change

Tested off the master branch of velum, and the ha branch (so, a single master or multiple master nodes)


robdaemon:
Note: I'll rebase / reword the commits when this is approved

Also: need to make a SUSE-based haproxy image, this currently uses the haproxy image from Docker Hub.

salt/etcd/init.sls
```diff
@@ -32,10 +32,6 @@ etcd:
       - etcd
     - require:
       - file: /etc/zypp/repos.d/containers.repo
-  cmd.run:
-    - name: rm -rf /var/lib/etcd/*
-    - prereq:
-      - service: etcd
```
robdaemon: I intentionally took this out - this is destructive, and can lead to a situation where your etcd cluster is destroyed if the initial orchestration run happens multiple times.

salt/haproxy/haproxy.manifest
```diff
@@ -0,0 +1,26 @@
...
+  labels:
+    name: haproxy
+spec:
+  containers:
+  - name: haproxy
+    resources:
+      requests:
+        memory: 128Mi
+      limits:
+        memory: 128Mi
```
robdaemon: this will need testing to know if these values are too big or too small - I just made a guess to start with.

salt/haproxy/haproxy.manifest
```diff
@@ -0,0 +1,26 @@
...
+    image: haproxy:1.7.5
+    volumeMounts:
+    - mountPath: /usr/local/etc/haproxy
+      name: haproxy-cfg
+  volumes:
+  - hostPath:
+      path: /etc/haproxy
```
PI-Victor: just one question, is `/etc/haproxy` a RW path?

robdaemon: Only Salt should be writing to that - it's intention is to contain the haproxy configuration only.

robdaemon:
I do have a TODO here - HUP / restart haproxy when the configuration. Will likely need to use `docker kill -s HUP <haproxy container>` since Kubernetes does not have a primitive for sending signals at this time. See https://github.com/kubernetes/kubernetes/issues/24957

salt/haproxy/haproxy.manifest
PI-Victor: i understand that, what i don't understand is if, under MicroOS, it can write to the path at all, or it's RO. sorry, i might be missing something.

salt/haproxy/haproxy.cfg.jinja
```diff
@@ -0,0 +1,28 @@
...
+        balance roundrobin
+
+{%- for minion_id, addrlist in salt['mine.get']('roles:kube-master', 'network.ip_addrs', 'grain').items() %}
+{% if addrlist is string -%}
+{% set addr = addrlist -%}
+{% else -%}
+{% set addr = addrlist|first -%}
+{% endif -%}
+        server master-{{ minion_id }} {{ addr }}:6443 check
+{% endfor -%}
```
mssola: Something to think in the future: what happens if I add a new node and I want it to become a master, will this work ? In the past we've been using [confd](https://github.com/kelseyhightower/confd) to handle this, but maybe this is something that salt can handle as well ?

salt/haproxy/haproxy.manifest
ereslibre: `/etc` should be RW in MicroOS, so this shouldn't be a problem...

salt/etcd/init.sls
kiall: Yea, that scares me. Maybe it should be split off and proposed separately to master?

salt/haproxy/haproxy.cfg.jinja
```diff
@@ -0,0 +1,28 @@
+global
+        log /dev/log    local0
+        log /dev/log    local1 notice
```
kiall: Where do these logs end up?

salt/haproxy/haproxy.cfg.jinja
kiall: If a new master is added (or removed), salt needs to be ran across all hosts and this file will be regenerated. Rob noted a TODO of actually triggering the haproxy config to be reloaded.

salt/haproxy/init.sls
```diff
@@ -0,0 +1,29 @@
+---
+include:
+  - repositories
+
+haproxy:
+  pkg.installed:
+  - pkgs:
+    - kubernetes-node
+  - require:
+    - file: /etc/zypp/repos.d/containers.repo
```
kiall: Is this still necessary? It feels odd to be asking for kubernetes-node to be installed here

salt/haproxy/haproxy.cfg.jinja
```diff
@@ -0,0 +1,28 @@
...
+        log     global
+        mode    tcp
+        option  tcplog
+        option  dontlognull
+        timeout connect 5000
+        timeout client 50000
+        timeout server 50000
+
+listen kubernetes-master
+        bind 127.0.0.1:6444
```
kiall: Ideally, we should listen on the standard kube-apiserver port here so any pre-existing configs don't need a port update once this lands, and the default port is more obvious for people when thinking it through later.

On the masters, we should probably also bind to 0.0.0.0, so end users will enter through the same LB.

salt/haproxy/haproxy.cfg.jinja
```diff
@@ -0,0 +1,28 @@
...
+        default-server inter 10s fall 3
+        balance roundrobin
+
+{%- for minion_id, addrlist in salt['mine.get']('roles:kube-master', 'network.ip_addrs', 'grain').items() %}
+{% if addrlist is string -%}
+{% set addr = addrlist -%}
+{% else -%}
+{% set addr = addrlist|first -%}
+{% endif -%}
+        server master-{{ minion_id }} {{ addr }}:6443 check
```
grahamhayes: We should make the connection go to `<minion-id>.infra.caasp.local` - means less HAProxy changes / reloads if an IP changes.

salt/haproxy/init.sls
grahamhayes: is it not needed for `kubelet` ?

salt/kubeconfig/kubeconfig.jinja
```diff
@@ -1,12 +1,8 @@
-{% set api_server = "api." + pillar['internal_infra_domain']  -%}
-{% set api_ssl_port = salt['pillar.get']('api:ssl_port', '6443') -%}
-{% set api_server_url = 'https://' + api_server + ':' + api_ssl_port -%}
-
 apiVersion: v1
 clusters:
 - cluster:
     certificate-authority: {{ pillar['paths']['ca_dir'] }}/{{ pillar['paths']['ca_filename'] }}
-    server: {{ api_server_url }}
+    server: https://api.{{ pillar['internal_infra_domain'] }}:6444
```
grahamhayes: We should probably leave this file as is, and update the pillar values

salt/kubelet/kubelet.jinja
```diff
@@ -15,7 +11,7 @@ KUBELET_PORT="--port=10250"
 KUBELET_HOSTNAME="--hostname-override={{ grains['id'] }}.{{ pillar['internal_infra_domain'] }}"
 
 # location of the api-server
-KUBELET_API_SERVER="--api-servers={{ api_server_url }}"
+KUBELET_API_SERVER="--api-servers=https://api.{{ pillar['internal_infra_domain'] }}:6444"
```
grahamhayes: same as `salt/kubeconfig/kubeconfig.jinja` - we should leave this file as is, and update pillars.

salt/kubernetes-master/config.jinja
```diff
@@ -22,5 +22,5 @@ KUBE_ALLOW_PRIV="--allow-privileged=true"
 KUBE_ENABLE_DAEMONSETS=true
 
 # How the controller-manager, scheduler, and proxy find the apiserver
-# note: we do not use SSL for local connections
-KUBE_MASTER="--master=http://127.0.0.1:8080"
\ No newline at end of file
+# note: uses haproxy and TLS for HA and resiliency
+KUBE_MASTER="--master=https://api.{{ pillar['internal_infra_domain'] }}:6444"
```
grahamhayes: we should look at the same logic as other places we use the `{{ api_server_url }}`

salt/haproxy/haproxy.manifest
robdaemon: @PI-Victor ahh, sorry, I misunderstood the why on your question :)

salt/haproxy/haproxy.cfg.jinja
robdaemon: They don't, really. haproxy cannot log to stdout. The "solutions" to this are not really solutions: https://github.com/dockerfile/haproxy/issues/3

salt/haproxy/init.sls
robdaemon: yeah, I probably have these backwards. Oops. Will reverse.

salt/kubeconfig/kubeconfig.jinja
robdaemon: This isn't really coming from the pillar.

```
{% set api_ssl_port = salt['pillar.get']('api:ssl_port', '6443') -%}
{% set api_server_url = 'https://' + api_server + ':' + api_ssl_port -%}
```

robdaemon:
all feedback has been addressed, please review the latest commit.

salt/haproxy/haproxy.manifest
```diff
@@ -0,0 +1,26 @@
...
+    name: haproxy
+spec:
+  containers:
+  - name: haproxy
+    resources:
+      requests:
+        memory: 128Mi
+      limits:
+        memory: 128Mi
+    image: haproxy:1.7.5
```
inercia: I think this image should be changed for something based on SLE

salt/etcd/init.sls
inercia: It is already known that we are currently not prepared for running the orchestration several times (in fact I think there is no way that orchestration can be triggered again once the cluster has been created). We should study how to solve these problems in the (near) future...

salt/kubernetes-master/init.sls
```diff
@@ -19,21 +21,31 @@ include:
...
-{% set api_ssl_port = salt['pillar.get']('api:ssl_port', '6443') %}
+      - subjectAltName: "{{ ", ".join(ip_addresses) }}"
+  /etc/kubernetes/kubelet:
+    file.managed:
+      - context:
+        schedulable: "false"
+  /etc/haproxy/haproxy.cfg:
+    file.managed:
+      - context:
+        bind_ip: "0.0.0.0"
```
inercia: I guess this will make haproxy listen on all interfaces, right? In that case, what is the reason?

salt/kubernetes-master/init.sls
```diff
@@ -65,7 +77,8 @@ kube-apiserver:
     - match:      state
     - connstate:  NEW
     - dports:
-        - {{ api_ssl_port }}
+        - {{ salt['pillar.get']('api:ssl_port', '6443') }}
+        - {{ salt['pillar.get']('api:lb_ssl_port', '6443') }}
```
inercia: Is it the same port for `lb_ssl_port` and `ssl_port`?

salt/kubernetes-master/init.sls
```diff
@@ -99,6 +116,9 @@ kube-scheduler:
     - watch:
       - file:     /etc/kubernetes/config
       - file:     kube-scheduler
+      - file: /etc/pki/minion.crt
+      - file: /etc/pki/minion.key
+      - file: {{ pillar['paths']['ca_dir'] }}/{{ pillar['paths']['ca_filename'] }}
```
inercia: Please align with the previous lines.

salt/kubernetes-master/init.sls
```diff
@@ -114,6 +134,9 @@ kube-controller-manager:
     - watch:
       - file:     /etc/kubernetes/config
       - file:     kube-controller-manager
+      - file: /etc/pki/minion.crt
+      - file: /etc/pki/minion.key
+      - file: {{ pillar['paths']['ca_dir'] }}/{{ pillar['paths']['ca_filename'] }}
```
inercia: ditto

inercia:
From my understanding of the code, I think you are running `haproxy` in the masters for client connections. Is that right? Don't you think that is an overkill? Is this just for the case that the local api server is dead while the local haproxy is alive?. Don't you think this is a quite improbable case?

salt/kubernetes-master/init.sls
```diff
@@ -19,21 +21,31 @@ include:
...
+  {% do ip_addresses.append("DNS: kubernetes.default.svc." + pillar['dns']['domain']) %}
 {% endif %}
 
 extend:
   /etc/pki/minion.crt:
     x509.certificate_managed:
-      - subjectAltName: "{{ ", ".join(extra_names + ip_addresses) }}"
-
-{% set api_ssl_port = salt['pillar.get']('api:ssl_port', '6443') %}
+      - subjectAltName: "{{ ", ".join(ip_addresses) }}"
```
ereslibre: `extra_names` and `ip_addresses` were meant to correspond to `DNS:` and `IP:` respectively in our SAN certificates. Fine from my point of view if we want to merge them.

salt/kubernetes-minion/init.sls
```diff
@@ -3,6 +3,23 @@
...
+# etcd needs proper certs on the minions too
+{% from 'cert/init.sls' import ip_addresses %}
+
+{% for _, interface_addresses in grains['ip4_interfaces'].items() %}
+  {% for interface_address in interface_addresses %}
+    {% do ip_addresses.append("IP: " + interface_address) %}
+  {% endfor %}
+{% endfor %}
+# add some extra names the API server could have
+{% do ip_addresses.append("DNS: " + grains['fqdn']) %}
```
ereslibre: We are mixing names and ip addresses here again. Maybe we can rename the `ip_addresses` var to something like `subject_alt_names` or something like that (everywhere, if we are going to mix DNS and IP entries on our SAN cert request). It's a minor thing, but when debugging can lead to mistakes...

salt/etcd/init.sls
inercia: We are currently tracking this issue in [this card](https://trello.com/c/2WxhX4hX)

salt/kubernetes-master/init.sls
inercia: Ok, I see, thanks for the clarification.

salt/haproxy/haproxy.manifest
inercia: I understand.

robdaemon:
Okay, I've addressed all the feedback and added the HUP on haproxy when the configuration changes. I'm pretty sure I'm done with this one, unless someone has more feedback? @kiall @grahamhayes @ereslibre ?

salt/haproxy/init.sls
```diff
@@ -0,0 +1,35 @@
...
+# send a HUP to haproxy when the config changes
+haproxy_restart:
+  cmd.run:
+    # we use a for loop here because this may run before haproxy starts, which is harmless, because
+    # we only care about sending the HUP when the configuration changes after the initial deployment
+    - name: |-
+            for i in $(docker ps -a | grep haproxy-k8-api | awk '{print $1}')
+            do 
+              docker kill -HUP $i 
+            done
```
inercia: I think you wouldn't need this loop if you use the [`onlyif`](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.cmd.html#salt.states.cmd.run) parameter...

robdaemon:
Thanks for the approval, @inercia - I'll go ahead and rebase this all into one commit.

salt/haproxy/init.sls
robdaemon: Ahh okay, that's good info for next time, thanks. This is the largest amount of work I've done in a SaltStack codebase ever.

ereslibre: Maybe a `onchanges` can help to only execute this HUP signal sending when the relevant config file has changed. A `watch` should also help to make it reactive.

inercia: @ereslibre There is already a `watch` for `file: /etc/haproxy/haproxy.cfg`, so this state `cmd.run`s when the file being `watch`ed changes. But it would be better if we could do that `onlyif` haproxy is running (instead of looping). As @ereslibre also mentioned, maybe `cmd.run` and `onchanges` would be a [better solution](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.cmd.html#should-i-use-cmd-run-or-cmd-wait)...

PR #60, opened 2017-05-09
Adding AZs to bare metal machines

robdaemon:
Adding two parameters - region and zone, to use the Kubernetes
scheduler's awareness of zones / regions during scheduling (for affinity
and anti-affinity rules) for HA

### DIFF
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -95,3 +95,7 @@ components:
   proxy:
     # Extra arguments to be passed to kube-proxy.
     args: ''
+
+availability_zone:
+  region: 'unknown'
+  zone: 'unknown'
--- a/salt/kubelet/kubelet.jinja
+++ b/salt/kubelet/kubelet.jinja
@@ -32,5 +32,4 @@ KUBELET_ARGS="\
     --cloud-provider=aws \
 {% endif -%}
     {{ pillar['components']['kubelet']['args'] }} \
-    --register-schedulable={{ schedulable }} \
     --kubeconfig={{ pillar['paths']['kubeconfig'] }}"
--- a/pillar/kubelet.sls
+++ b/pillar/kubelet.sls
@@ -1,2 +1,3 @@
 # infra container to use instead of downloading gcr.io/google_containers/pause
 pod_infra_container_image: sles12/pause:1.0.0
+node_labels: []
--- a/salt/haproxy/init.sls
+++ b/salt/haproxy/init.sls
@@ -29,7 +29,7 @@ haproxy_restart:
     - name: |-
             for i in $(docker ps -a | grep haproxy-k8-api | awk '{print $1}')
             do 
-              docker kill -HUP $i 
+              docker kill --signal HUP $i 
             done
     - watch:
       - file: /etc/haproxy/haproxy.cfg
--- a/salt/kubelet/init.sls
+++ b/salt/kubelet/init.sls
@@ -4,6 +4,34 @@
 include:
   - repositories
 
+{% set node_labels = salt['pillar.get']('node_labels', []) %}
+{% set region = salt['pillar.get']('availability_zone:region', '') %}
+{% set zone = salt['pillar.get']('availability_zone:zone', '') %}
+{% if region != '' -%}
+{% do node_labels.append('failure-domain.beta.kubernetes.io/region=' + region) -%}
+{% endif -%}
+{% if zone != '' -%}
+{% do node_labels.append('failure-domain.beta.kubernetes.io/zone=' + zone) -%}
+{% endif -%}
+
+/etc/kubernetes/kubelet-initial:
+  file.managed:
+    - name: /etc/kubernetes/kubelet-initial
+    - source: salt://kubelet/kubelet-initial.jinja
+    - template: jinja
+    - defaults:
+      node_labels: {{ node_labels }}
+      schedulable: "true"
+
+az_labels:
+  cmd.run:
+    - name: kubectl --kubeconfig=/var/lib/kubelet/kubeconfig label node --overwrite {{ grains['id'] }}.{{ pillar['internal_infra_domain'] }} {{ node_labels|join(' ') }}
+    # don't bother if kubectl get node doesn't return this node, it'll take affect when the node starts
+    - onlyif: 
+      - kubectl --kubeconfig=/var/lib/kubelet/kubeconfig get node {{ grains['id'] }}.{{ pillar['internal_infra_domain'] }}
+    - onchanges:
+      - file: /etc/kubernetes/kubelet-initial
+
 kubelet:
   pkg.installed:
     - pkgs:
@@ -13,17 +41,18 @@ kubelet:
       - file: /etc/zypp/repos.d/containers.repo
 
   file.managed:
-    - name:     /etc/kubernetes/kubelet
-    - source:   salt://kubelet/kubelet.jinja
+    - name: /etc/kubernetes/kubelet
+    - source: salt://kubelet/kubelet.jinja
     - template: jinja
-    - defaults:
-      schedulable: "true"
+
   service.running:
     - enable:   True
     - watch:
-      - file:   /etc/kubernetes/config
-      - file:   {{ pillar['paths']['kubeconfig'] }}
-      - file:   kubelet
+      - file: /etc/kubernetes/config
+      - file: {{ pillar['paths']['kubeconfig'] }}
+      - file: kubelet
       - file: /etc/pki/minion.crt
       - file: /etc/pki/minion.key
       - file: {{ pillar['paths']['ca_dir'] }}/{{ pillar['paths']['ca_filename'] }}
+    - require:
+      - file: /etc/kubernetes/kubelet-initial
--- a/salt/kubelet/kubelet-initial.jinja
+++ b/salt/kubelet/kubelet-initial.jinja
@@ -1,0 +1,2 @@
+KUBELET_INITIAL_ARGS="--node-labels={{ node_labels|join(",") }} \
+    --register-schedulable={{ schedulable }}"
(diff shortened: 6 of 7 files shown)


salt/kubelet/kubelet.jinja
```diff
@@ -32,5 +44,6 @@ KUBELET_ARGS="\
     --cloud-provider=aws \
 {% endif -%}
     {{ pillar['components']['kubelet']['args'] }} \
+    --node-labels={{ node_labels|join(',') }} \
```
inercia: What if `node_labels = []`? Can we provide `--node-labels=` to the kubelet args?

pillar/top.sls
```diff
@@ -6,5 +6,6 @@ base:
     - ip_addrs
     - fqdn
     - kubelet
+    - availability_zone
```
inercia: It seems this is only used in the kube minions... could we use this?

```
   'G@roles:kube-minion':
      - availability-zone
```

As a minor comment, it seems our code convention for modules is that we do not use underscores, so it would be better to use `availability-zone`.

salt/kubelet/kubelet.jinja
kiall: I think the way it's setup, there was always be values.. However, wrapping it in a `{% if node_labels %} .. {% endif %}` is harmless and prevents it breaking if `--node-labels=` would otherwise cause an issue..

Also, this will only set the labels once, on first startup (well, during node registration). Any post install kubelet start changes won't be applied. And, if the labels change, we'll unnecessarily restart the kubelet daemon. We hit this in the HCP ansible port, and solved it using a mix of two things:

1. Render the labels args into a separate file, e.g. `/etc/kuebrnetes/kubelet-initial`, do not watch this file from any services. Include this file into this one, and append the included `$KUBELET_ARGS` to `KUBELET_ARGS`. This means labels get set on the initial startup, and no unnecessary restarts will happen.

2. To handle future label updates, we had to do something like this for each label: `kubectl --kubeconfig /srv/kubernetes/kubeconfig label --overwrite node {{ inventory_hostname }} {{ item.key }}={{ item.value }}` (we had a TODO to make this a ansible module rather than a shell out, maybe salt has a module ready to use?)

pillar/availability_zone.sls
```diff
@@ -0,0 +1,3 @@
+availability_zone:
+  region: 'region1'
+  zone: 'zone1'
```
kiall: I think these belong in `pillar/params.sls`, as we expect users to override them. Should we maybe default to "unknown" for both rather than potentially telling end-users we know something about the topology?

(Also - Total guess that all user settable pillars are in `params.sls`, It just looks like they are..

pillar/top.sls
kiall: I think the way it's setup, there was always be values.. However, wrapping it in a `{% if node_labels %} .. {% endif %}` is harmless and prevents it breaking if `--node-labels=` would otherwise cause an issue..

salt/kubelet/kubelet.jinja
kiall: (The ansible we used is here as a reference if needed https://gitlab.suse.de/caasp-ha/hcp-ansible-poc/tree/master/ansible/roles/kubelet )

kiall: https://docs.saltstack.com/en/latest/ref/states/all/salt.states.k8s.html might work, assuming we have a new enough salt, and it doesn't have the same issue the ansible module did - no ability to use client certs.

robdaemon: Providing `--node-labels=` is harmless (as a matter of fact, that's the default)

As far as changing the labels, it's more complicated than just the kubectl call. These labels affect scheduling of containers. Changing the labels will not cause the containers to be rescheduled.

pillar/top.sls
robdaemon: It's used in all machines, including masters.

Also, the convention does not appear to be `_`, as you'll see `ip_addrs` directly above this.

pillar/availability_zone.sls
robdaemon: I can move them, no big deal.

However, the `{% if node_labels %}` isn't necessary (that's why I didn't put it there to begin with!)

robdaemon:
Okay I've updated the PR with some of the requested changes. I'm struggling with how to properly update the labels when they're changed.

```
kubectl label --overwrite -n node_name label1=foo
```

does not replace all labels.

You have to use this syntax to remove a label:

```
kubectl label --overwrite -n node_name label1-
```

To do this properly, I need to be able to diff the original value of `node_labels` and the new value. I'm not sure how to do this in Salt. I don't really want to do multiple calls to `kubectl`, because I'd like this entire operation to be a single transaction, so there's no chance of the scheduler doing something wrong while this state is being applied. I'm stumped on this part of using the pillar. Is the history of a variable available?

pillar/params.sls
```diff
@@ -95,3 +95,7 @@ components:
   proxy:
     # Extra arguments to be passed to kube-proxy.
     args: ''
+
+availability_zone:
+  region: 'unknown'
+  zone: 'unknown'
```
inercia: What about leaving these two parameters empty instead of setting them to `unknown`? Or is `unknown` a key value?

kiall: I'd argue to set to them to "unknown", so that the region and zone labels appear and operators both know they exist, and have an incentive to "fix" them.

inercia:
@robdaemon FYI, regarding labels, maybe you will find this [this Salt module](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.k8s.html) helpful...

robdaemon:
@kiall please review the latest commits. I've added the ability to modify the AZ labels, including the extra file so that we don't unnecessarily restart the kubelet. This seems... complicated.

salt/kubelet/kubelet.service
```diff
@@ -0,0 +1,25 @@
...
+	    $KUBELET_ADDRESS \
+	    $KUBELET_PORT \
+	    $KUBELET_HOSTNAME \
+	    $KUBE_ALLOW_PRIV \
+        $KUBELET_NODE_LABELS \
+	    $KUBELET_ARGS
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target
```
inercia: I understand this is only for testing this works, because is generally a bad idea to overwrite the `kubelet.service` from Salt...

If you want to add `$KUBELET_NODE_LABELS` to the service it would be better to add it in [the service file](https://build.opensuse.org/package/view_file/Virtualization:containers/kubernetes/kubelet.service?expand=1), then add `KUBELET_NODE_LABELS=` in `/etc/kubernetes/kubelet` using the Jinja template we have, and just watch that file (Salt will reload the service automatically, so you will not need that `service.systemctl_reload`)...

kiall: kubelet will read the value of `--node-labels` exactly once, at which point, future changes to the value of kubelets `--node-labels` will be ignored. So, if this value changes, we don't want to restart kubelet .. Because it gains us nothing. However, I agree we shouldn't be changing this file.. Another comment suggests an alternative.

salt/kubelet/node_labels.jinja
```diff
@@ -0,0 +1 @@
+KUBELET_NODE_LABELS="--node-labels={{ node_labels|join(",") }}"
```
kiall: Rather than call this "node_labels", we should make it more generic.. It's an "extension" of `/etc/kubernetes/kublet` for values that are only read once upon initial startup. so `/etc/kubernetes/kublet-initial` makes sense. 

The file can then contain this for now:

    KUBELET_ARGS="--node-labels={{ node_labels|join(",") }}"

and /etc/kubernetes/kubelet should be amended to:

    source /etc/kubernetes/kubelet-initial

    # Add your own!
    KUBELET_ARGS="\
        $KUBELET_ARGS \
        ....

This provides a place for other values only read on the initial startup to be placed here too (e.g. `--register-schedulable`), prevents unnecessary service restarts since `-initial` won't have any watchers, and provides a pattern that can be used for other services with similar "read once on initial startup" variables / flags - e.g. etcd's initial clustering variables)

inercia: I agree with @kiall on the solution: add an `-initial` file and just do not `watch` it. That would be enough.

Just one thing: I remember having problems when trying to concat envars like `KUBELET_ARGS="$KUBELET_ARGS ..."`, so maybe you have to use different names...

robdaemon: That will not work. `/etc/kubernetes/kubelet` cannot source other files. It is not interpreted by a shell. You also cannot do parameter expansion.

https://ask.fedoraproject.org/en/question/39528/systemd-when-using-environmentfile-variables-are-not-expanded/

So, we have to rewrite the entire systemd unit. I'll rename things to -initial though.

robdaemon: I'll have to learn how to submit changes to the rpm packages now, I suppose 😁

kiall: It may be possible to use a systemd drop in unit rather than replace the whole file / edit the package? Combined with something like this? https://gist.github.com/nickjacob/9909574

I really should not have been surprised systemd doesn't "execute" the script the same as other init systems have the past.. Oh well.

Either way, even though this is only 1 small detail of this patch, the pattern will be needed for etcd anyway, as restarting etcd services is a much bigger deal.

inercia: > It may be possible to use a systemd drop in unit rather than replace the whole file / edit the package? Combined with something like this? https://gist.github.com/nickjacob/9909574

I've seen that solution in the past and I think it is too dirty for my taste... I think it would be better to add a envvar to the [kubelet service](https://build.opensuse.org/package/view_file/Virtualization:containers/kubernetes/kubelet.service?expand=1) file...

robdaemon:
Updated - depends on https://build.opensuse.org/request/show/494998 now

salt/kubelet/node_labels.jinja
robdaemon: resolved with a kubelet-initial file.

PR #61, opened 2017-05-10
/etc/hosts, hostnames and etcd fixes

inercia:
- Fix: set the hostname the first thing in the orchestration
- Fix: update the grains after setting the hostname
- Fix: add all the machines in the cluster to /etc/hosts, so etcd discovery is happy
- Fix: we must update /etc/hosts before etcd is started
- Fix: fix an error introduced in #37 where Flannel is setup at the same time as the masters, but it must be setup before.


salt/etcd-proxy/etcd.conf
```diff
@@ -1,2 +1,10 @@
 [Service]
-ExecStartPre=/bin/bash -c "if [[ -d /var/lib/etcd/proxy ]] && [[ -d /var/lib/etcd/member ]]; then rm -rf /var/lib/etcd/member; fi"
+# follow these instructions:
+# https://coreos.com/etcd/docs/latest/etcd-live-cluster-reconfiguration.html#replace-a-failed-etcd-member-on-coreos-container-linux
+
+# remove previously created proxy directory
+ExecStartPre=/usr/bin/rm -rf /var/lib/etcd/proxy
+
+# NOTE: use this option if you would like to re-add broken etcd member into cluster
```
ereslibre: What option is this referring to? It's already being done, no? Why clean up `/var/lib/etcd/proxy` twice?

Issue #62, opened 2017-05-11
increase number of salt worker threads?


PI-Victor:
The bootstrapping process never finishes, it seems i hit this while trying to bootstrap 5 nodes
https://github.com/saltstack/salt/issues/27421

here's the [log dump of the salt master](http://pastebin.suse.de/17171)

ereslibre:
I am hitting this problem with 2 nodes after trying to execute a clean run of the e2e tests, and I already have the 10 `worker_threads` setting.

ereslibre:
NVM, I was referring to this error that I am getting consistently now (not sure if this happened before, but I don't think so):

`Received function _file_find which is unavailable on the master, returning False`

PI-Victor:
yeah, i get that error all the time. i actually found some fix for it, but it's some hacky thing where you add that `_file_find` function just so that you can squash the notification. i remember looking into it some time ago.

PR #63, opened 2017-05-11
bump number of worker threads

PI-Victor:
see https://docs.saltstack.com/en/latest/ref/configuration/master.html#worker-threads
* to avoid minion calls to master timing out
* fixes https://github.com/kubic-project/salt/issues/62


PI-Victor:
the default value is 5, not sure if setting this to ten will fix everything on large bootstraps. like 100-200 nodes, but it does work with nodes up to 10, tested on my server.

PR #66, opened 2017-05-15
Handle proxies in the docker daemon (bsc#1036627)

inercia:
This can be tested by deploying a cluster with [our Terraform scripts](https://github.com/kubic-project/terraform), doing:

1. `k8s-setup ... -Vpillar=proxy.http~http://something.com:8080,proxy.https~https://something.com:443` (requires kubic-project/terraform#66)
2. orchestrating the cluster
3. checking that, in any machine of the cluster, we can run `systemctl show --property=Environment docker` and get something like `Environment=HTTP_PROXY=http://something.com:8080`

### DIFF
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -58,6 +58,11 @@ flannel:
   etcd_key:       '/flannel/network'
   iface:          'eth0'
 
+proxy:
+  http:           ''
+  https:          ''
+  no_proxy:       ''
+
 # Configuration for the reboot manager (https://github.com/SUSE/rebootmgr).
 # notes:
 # - The default group for rebootmgr is "default", so we are simply taking
--- a/salt/docker/init.sls
+++ b/salt/docker/init.sls
@@ -2,18 +2,16 @@ include:
   - repositories
   - flannel
 
-docker:
-  pkg.installed:
-    - install_recommends: False
-    - require:
-      - file: /etc/zypp/repos.d/containers.repo
-  service.running:
-    - enable: True
-    - watch:
-      - service: flannel
-      - file: /etc/sysconfig/docker
-    - require:
-      - pkg: docker
+/etc/systemd/system/docker.service.d/proxy.conf:
+  file.managed:
+    - makedirs: True
+    - contents: |
+        [Service]
+        Environment=HTTP_PROXY={{ salt['pillar.get']('proxy:http', '') | yaml_encode }}
+        Environment=HTTPS_PROXY={{ salt['pillar.get']('proxy:https', '') | yaml_encode }}
+        Environment=NO_PROXY={{ salt['pillar.get']('proxy:no_proxy', '') | yaml_encode }}
+  cmd.run:
+    - name: systemctl daemon-reload
 
 {% set docker_opts = salt['pillar.get']('docker:args', '')%}
 {% set docker_reg  = salt['pillar.get']('docker:registry', '') %}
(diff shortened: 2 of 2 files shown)


pillar/params.sls
```diff
@@ -58,6 +58,10 @@ flannel:
   etcd_key:       '/flannel/network'
   iface:          'eth0'
 
+proxy:
+  http:           ''
+  https:          ''
```
kiall: no_proxy will also be required, as there are many environments where HTTP proxies are needed for external things (e.g. docker hub), but won't work for internal things (e.g. a local registry).

salt/docker/init.sls
```diff
@@ -2,25 +2,35 @@ include:
...
+{% set proxy_https = salt['pillar.get']('proxy:https', '') %}
+{% if proxy_https|length > 0 %}
+/etc/systemd/system/docker.service.d/https-proxy.conf:
+  file.managed:
+    - content: |
+      [Service]
+      Environment="HTTPS_PROXY={{ proxy_https }}"
+    - watch_in:
+      - service: docker
+{% endif %}
```
kiall: These two should be updated to remove the proxy when the pillar is removed. Just removing the if|length should do it, unless systemd doesn't like `Environment="HTTPS_PROXY=`. At that point, both files can also be combined into one.

salt/docker/init.sls
```diff
@@ -2,25 +2,35 @@ include:
...
 
 {% set docker_opts = salt['pillar.get']('docker:args', '')%}
 {% set docker_reg  = salt['pillar.get']('docker:registry', '') %}
 {% if docker_reg|length > 0 %}
   {% set docker_opts = docker_opts + " --insecure-registry={{ docker_reg }} --registry-mirror=http://{{ docker_reg }}" %}
 {% endif %}
 
+{% if docker_opts|length > 0 %}
```
kiall: This change prevents the removal of the optional docker_opts?

salt/docker/init.sls
inercia: Yeah, I think we could combine both files...

salt/docker/init.sls
inercia: What do you mean?

pillar/params.sls
inercia: Yeah, I was about to add a `no_proxy`...

inercia:
Do not merge yet: it seems it doesn't work fine yet...

salt/docker/init.sls
```diff
@@ -2,32 +2,51 @@ include:
...
+  # [inercia] when dockerd was already running and we require the
+  # service to be "service.running", Salt does not think it must
+  # restart it even when we say "watch these files", so we
+  # need this "cmd.run"...
+  cmd.run:
+    - name: systemctl restart docker.service
+    - onlyif: systemctl status docker.service
+    - require:
+      - file: /etc/sysconfig/docker
+      - /etc/systemd/system/docker.service.d/proxy.conf
```
flavio: Sorry to jump on that after the code has been merged... I think this is wrong, it should be:

```
- file: /etc/systemd/system/docker.service.d/proxy.conf
```

salt/docker/init.sls
```diff
@@ -2,32 +2,51 @@ include:
...
+  service.running:
+    - enable: True
+    - watch:
+      - service: flannel
+      - pkg: docker
+      - file: /etc/sysconfig/docker
+      - /etc/systemd/system/docker.service.d/proxy.conf
```
flavio: Same here, isn't it?

Sorry to jump on that after the code has been merged... I think this is wrong, it should be:

```
- file: /etc/systemd/system/docker.service.d/proxy.conf
```

salt/docker/init.sls
inercia: No, it is ok: requiring `/etc/systemd/system/docker.service.d/proxy.conf` means that it requires `file: /etc/systemd/system/docker.service.d/proxy.conf` and `cmd: /etc/systemd/system/docker.service.d/proxy.conf`.

salt/docker/init.sls
inercia: Ditto

PR #71, opened 2017-05-17
Do not create the file if it doesn't exist in other OS than openSUSE Leap

ereslibre:
Fixes: bsc#1039453

### DIFF
--- a/salt/repositories/containers.repo
+++ b/salt/repositories/containers.repo
@@ -4,5 +4,4 @@ type=rpm-md
 gpgcheck=0
 enabled=1
 autorefresh=1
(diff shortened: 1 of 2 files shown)


ereslibre:
You are right, let me find a proper solution

flavio:
Maybe we can use the `create` attribute of the [managed file](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html#salt.states.file.managed):

> create
> : True
>     If set to False, then the file will only be managed if the file already exists on the system.

This could be set to `False` when the os != openSUSE.

Just my 2cents

PR #72, opened 2017-05-17
WIP: Improve bootstrapping performance

ereslibre:
* Do not generate a minion certificate on every machine that needs to
  download the CA certificate.
* Generate certificates per service, not per machine.

The rationale behind this is to generate keys/certificates per services for now,
since they are way easier to handle, and it will happen once at the beginning
of the orchestration, and then the certificates/keys will be deployed to each
machine depending on its role. At the very beginning of the orchestration three
certificates will be generated:

- CA certificate (authority, key kept always private)
- etcd certificate (published on the mine)
- apiserver certificate (published on the mine)

After the last two have been generated and created will be published on the mine,
and we will use it for deploying them on the rest of the minions depending on the
role of the machine.

This also allows us for a much better way of handling certificate expiration. We
would update the certificate, publish the new certificate on the mine, and then
deploy the new certificates like we do normally (and all related services would be
automatically restarted as required).

Split the phase of generation and installation of certificates, since they are
clearly two different steps.

CC: bsc#1038814


robdaemon:
I'm not so sure on this one. You can't revoke a machine without revoking the entire service, causing an outage. I realize that a certificate per service per machine takes a lot of computing power to generate, but it's a more secure configuration. I think @ukbelch needs to chime in on this one, since it's a security thing.

ereslibre:
Yes, that's a very valid point...

pillar/top.sls
```diff
@@ -5,7 +5,7 @@ base:
     - certificates
     - ip_addrs
     - fqdn
-  'G@roles:kube-master':
+  'G@roles:kube-master or ca':
```
ereslibre: This wouldn't be on the final version of course...

kiall:
Agree with Rob here, i see 3 choices for certs (ignoring perf concerns):

* Per server per service (i.e. etcd client cert for server1)
* Per server (cert for server1 is client and a server cert for all services)
* Per service (server1 + server2 share an etcd cert)

I think the first is best, the second is probably reasonable, and the 3rd feels dangerous..

I'm thinking we have better ways around this, and even if we do reduce the number of certs, were still going to hit this issue with some other task that takes a few seconds to complete..

I'm on a cell, so can't type the suggestions out now.. but will do tomorrow.

grahamhayes:
So - the single certs per service are an issue - however, pre generation of certs, in a single step before we start orchestrating workers might help the thundering herd issues.

From what I can tell, all the workers request certs at the same time, which ties up all the worker threads, and as salt uses ZeroMQ, not a real queue, the request gets dropped, and we end up with some workers that fail to get a cert.

inercia:
> From what I can tell, all the workers request certs at the same time, which ties up all the worker threads, and as salt uses ZeroMQ, not a real queue, the request gets dropped, and we end up with some workers that fail to get a cert.

Salt isn't supposed to be so bad scaling with the number of minions, so I don't think we will hit that scenario until we have something bigger than our current deployments. I think the problem must lie somewhere else...

kiall:
Honestly, I think the "quick fix" here is to drop the batch size down to a reasonable number, we'll need this anyway, as it will never be possible to bootstrap say a 1000 machines all at once. The batch size likely needs to be lower than the number of workers by some %, but I'm not sure what %.

This gets us past the initial issue, and making it faster can happen over time.

inercia:
> Honestly, I think the "quick fix" here is to drop the batch size down to a reasonable number, we'll need this anyway, as it will never be possible to bootstrap say a 1000 machines all at once. The batch size likely needs to be lower than the number of workers by some %, but I'm not sure what %.

Not sure if we can tune the batch size for this but, if that is the case, and agree with @kiall...

grahamhayes:
@inercia it is not that bad at scaling - but having all nodes asking for an expensive operation at the same time can cause issues. For short running operations, this is no porblem, as the ZMQ message just sits in the buffer, but when it is a long running thing, it can timeout.

Salt does not do a retry, and as the ZMQ bus is not a real message queue, there is no way to set it as "persistent"

I am testing the batch sizes now, but long term we should look at pre-generation of certs as well.

PR #74, opened 2017-05-18
Certain key/values are interpreted in YAML and have special meaning

ereslibre:
For example, Norway country that we represent with its alpha-2 ISO
code will be `NO`, which will get translated when loading this YAML
into memory as False.

Force literal values and we avoid this problem.

Fixes: bsc#1039730

### DIFF
--- a/salt/cert/init.sls
+++ b/salt/cert/init.sls
@@ -38,14 +38,14 @@ include:
     - signing_policy: minion
     - public_key: /etc/pki/minion.key
     - CN: {{ grains['fqdn'] }}
-    - C: {{ pillar['certificate_information']['subject_properties']['C'] }}
-    - Email: {{ pillar['certificate_information']['subject_properties']['Email'] }}
-    - GN: {{ pillar['certificate_information']['subject_properties']['GN'] }}
-    - L: {{ pillar['certificate_information']['subject_properties']['L'] }}
-    - O: {{ pillar['certificate_information']['subject_properties']['O'] }}
-    - OU: {{ pillar['certificate_information']['subject_properties']['OU'] }}
-    - SN: {{ pillar['certificate_information']['subject_properties']['SN'] }}
-    - ST: {{ pillar['certificate_information']['subject_properties']['ST'] }}
+    - C: {{ pillar['certificate_information']['subject_properties']['C']|yaml_dquote }}
+    - Email: {{ pillar['certificate_information']['subject_properties']['Email']|yaml_dquote }}
+    - GN: {{ pillar['certificate_information']['subject_properties']['GN']|yaml_dquote }}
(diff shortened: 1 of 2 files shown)


flavio:
LGTM, is there any way to cover that with some tests?

salt/cert/init.sls
```diff
@@ -38,14 +38,14 @@ include:
...
-    - SN: {{ pillar['certificate_information']['subject_properties']['SN'] }}
-    - ST: {{ pillar['certificate_information']['subject_properties']['ST'] }}
+    - C: "{{ pillar['certificate_information']['subject_properties']['C'] }}"
+    - Email: "{{ pillar['certificate_information']['subject_properties']['Email'] }}"
+    - GN: "{{ pillar['certificate_information']['subject_properties']['GN'] }}"
+    - L: "{{ pillar['certificate_information']['subject_properties']['L'] }}"
+    - O: "{{ pillar['certificate_information']['subject_properties']['O'] }}"
+    - OU: "{{ pillar['certificate_information']['subject_properties']['OU'] }}"
+    - SN: "{{ pillar['certificate_information']['subject_properties']['SN'] }}"
+    - ST: "{{ pillar['certificate_information']['subject_properties']['ST'] }}"
```
inercia: I think all these pillar values (and, in general, all the things that come from a user input) would need to be "yaml escaped" with something like `yaml_encode` or `yaml_dquote` (see [this](https://docs.saltstack.com/en/latest/topics/jinja/index.html#yaml_dquote)).

ereslibre:
> LGTM, is there any way to cover that with some tests?

We would need to add this to the e2e-tests, since this is how salt interprets the `NO` without quotes.

salt/cert/init.sls
ereslibre: We aren't embedding jinja into jinja... it's the fact that `NO` gets interpreted by the yaml parser as `False`, so quoting is the optimal solution IMHO.

inercia: `yaml_dquote` is not for embeding YAML code but for escaping some text and produce some valid YAML code. From the documentation: `yaml_dquote`: _"Serializes a string into a properly-escaped YAML double-quoted string. This is useful when the contents of a string are unknown and may contain quotes or unicode that needs to be preserved. The resulting string will be emitted with opening and closing double quotes."_

Not strictly necessary in this case (as you said, `"` should be enough), but it also escapes unicode and so...

ereslibre: Indeed, `yaml_dquote` can be used. I wonder if it can fix bsc#1031475

ereslibre: I am not sure if in this case it would detect `NO` and quote it... might be the case. It's worth checking it.

PR #78, opened 2017-05-18
WIP: Breaking the kube roles up into smaller formulas

robdaemon:
Will allow for arbitrary composition of kubernetes roles on hosts


salt/top.sls
```diff
@@ -14,11 +14,17 @@ base:
     - flannel
     - docker
     - haproxy
+    - kube-common
```
kiall: Generally, I would try to avoid applying "boring" roles like kube-common to a set of servers, and instead allow the other that need it (e.g.kube-apiserver) to include it.

This is really just a matter of taste though, I believe the top.sls (or playbooks in ansible..)  should show only the important bits - i.e. what services are getting deployed..

salt/top.sls
```diff
@@ -14,11 +14,17 @@ base:
...
+    - kube-proxy
     - kubelet
     - kubeconfig
   'roles:kube-master':
     - match: grain
     - kubernetes-master
+    - kube-apiserver
+    - kube-controller-manager
+    - kube-scheduler
+    - reboot
```
kiall: Was adding `reboot` intentional? It looks like this role is not idempotent, so it needs to be fixed before it can be ran multiple times safely.

salt/kube-apiserver/init.sls
```diff
@@ -0,0 +1,38 @@
+include:
+  - repositories
```
kiall: It would be better to include kube-common here (and all the other kube-* service roles), and in kube common include `repositories` and `cert`.

salt/kube-controller-manager/init.sls
```diff
@@ -0,0 +1,23 @@
+include:
+  - repositories
```
kiall: The comment above about moving repos and certs -> kube-common fixes this, but.. If this was left as is, and kube-controller-manager was deployed onto it's own server, the certs would be missing :)

salt/kube-proxy/init.sls
```diff
@@ -0,0 +1,24 @@
...
+  - repositories
+
+conntrack-tools:
+  pkg.installed
+
+kube-proxy:
+  pkg.installed:
+    - pkgs:
+      - iptables
+      - conntrack-tools
```
kiall: It looks like we're installing conntrack-tools twice, it's probably best to remove the one at line 4?

robdaemon:
Will add generating a cert per service in a future PR when the salt master scaling issues are resolved.

PR #80, opened 2017-05-22
Fix some problems with Docker when HTTP proxy vars are empty

inercia:


PI-Victor:
i'll leave this here as well, since it's more persistent.
the problem is quoting in the systemd drop-in unit and not a docker issue as i thought.
so all is needed is for the final provisioning of the `proxy.conf` to look like this:
```
[Service]
Environment="HTTP_PROXY="
Environment="HTTPS_PROXY="
Environment="NO_PROXY="
```
instead of what it currently looks like

```
9293820ab0ca4233ad1a5648aa8dfb48:/etc/systemd/system/docker.service.d # cat proxy.conf 
[Service]
Environment=HTTP_PROXY=""
Environment=HTTPS_PROXY=""
Environment=NO_PROXY=""

```

ereslibre:
I think @PI-Victor is a better approach by just fixing quoting... my 2 cents.

inercia:
I think it should be fixed now: it seems that removing que `yaml_encode` and some `"` did it...

inercia:
@PI-Victor Yeah, tests have been failing with those errors since last week... :-(

PR #81, opened 2017-05-22
Fixing broken etcd configuration

robdaemon:
Stray + character was causing this line to not execute, and I ended up with a cluster with both folders present, preventing etcd from starting.


PI-Victor:
bugger! copied the `+` sign :/ thanks @robdaemon

PR #84, opened 2017-05-24
Batch runs of the `cert` state

grahamhayes:
This allows more nodes to be deployed without causing
timeouts and failed runs on the `cert` state.

Also, remove concurrecny from the etcd member and proxy
to ensure members are created before proxies

### DIFF
--- a/salt/orch/kubernetes.sls
+++ b/salt/orch/kubernetes.sls
@@ -67,7 +67,7 @@ etcd_nodes_setup:
     - tgt: 'roles:etcd'
     - tgt_type: grain
     - highstate: True
-    - concurrent: True
+    - batch: 1
     - require:
       - salt: etcd_discovery_setup
 
@@ -77,15 +77,15 @@ etcd_proxy_setup:
     - tgt_type: grain_pcre
     - sls:
       - etcd-proxy
-    - concurrent: True
+    - batch: 5
     - require:
-      - salt: etcd_discovery_setup
+      - salt: etcd_nodes_setup
 
 flannel_setup:
   salt.state:
     - tgt: 'roles:kube-master'
     - tgt_type: grain
-    - concurrent: True
+    - batch: 5
     - sls:
       - flannel-setup
     - require:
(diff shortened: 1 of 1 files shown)


grahamhayes:
We need to wait until there is an image with the `2016.11.4`  version of salt available as this will break otherwise

salt/orch/kubernetes.sls
```diff
@@ -72,9 +72,9 @@ etcd_proxy_setup:
     - tgt_type: grain_pcre
     - sls:
       - etcd-proxy
-    - concurrent: True
+    - batch: 5
```
inercia: Does this `batch` imply that provisioning will happen in parallel in these nodes? Otherwise I think this will not work, as all the masters must be started at the same time so they can create the initial cluster...

salt/orch/kubernetes.sls
```diff
@@ -72,9 +72,9 @@ etcd_proxy_setup:
...
     - require:
-      - salt: etcd_discovery_setup
+      - salt: etcd_nodes_setup
```
grahamhayes: This change *should* mean that the etcd members are started before this run.

grahamhayes:
OK, with salt `2016.11.4` + this patchset 30 hardware nodes installed with no issues

salt/orch/kubernetes.sls
```diff
@@ -67,7 +67,7 @@ etcd_nodes_setup:
     - tgt: 'roles:etcd'
     - tgt_type: grain
     - highstate: True
-    - concurrent: True
+    - batch: 1
```
inercia: Maybe we could make these parameters configurable, instead of hardcoding `1`.

We are currently not deploying _only-etcd_ nodes, but I think this could cause some problems if we create, for example, a 7-nodes etcd cluster. Provisioning would only happen in one machine at a time, and that would mean that masters would fail to find other master...

grahamhayes: When we start doing that we will have to do a lot of updates anyway - this is a known good setting for a large amount of machines.

PR #85, opened 2017-05-24
Allow for clean shutdown of nodes

kiall:
Add a stop SLS for each service we wish to shutdown clearly, doing any
necessary pre-stop actions such as draining and cordoning kubelet.

There are few TODOs inline, any suggestions on the salt way to handle
these would be nice. I'm throwing this up now as it works enough to unblock
other parts of the upgrade, but I don't think it should be merged as-in. 

In the meantime, I'll go looking to see if there's a kubernetes module we can
use in place of shelling out to kubectl.

### DIFF
--- a/salt/docker/stop.sls
+++ b/salt/docker/stop.sls
@@ -1,0 +1,4 @@
+# Stop and disable the docker daemon
+docker:
+  service.dead:
+    - enable: False
--- a/salt/kubernetes-minion/init.sls
+++ b/salt/kubernetes-minion/init.sls
@@ -62,13 +62,18 @@ kubelet:
     - require:
       - service: kubelet
 
-# note: br_netfilter is not available in some kernels
-#       not sure we really need it...
-#
-#      - kmod: br_netfilter
-#
-#br_netfilter:
-#  kmod.present
+  # TODO: This needs to wait for the node to register, which takes a few seconds.
+  # Salt doesn't seem to have a retry mechanism in the version were using, so I'm
+  # doing a horrible hack right now.
+  cmd.run:
+    - name: |
+        ELAPSED=0
+        until kubectl uncordon {{ grains['fqdn'] }} ; do
+            [ $ELAPSED -gt 60 ] && exit 1
+            sleep 1 && ELAPSED=$(( $ELAPSED + 1 ))
+        done
+    - env:
+      - KUBECONFIG: {{ pillar['paths']['kubeconfig'] }}
 
 #######################
 # config files
--- a/salt/kubernetes-minion/stop.sls
+++ b/salt/kubernetes-minion/stop.sls
@@ -1,0 +1,24 @@
+# Stop and disable the Kubernetes minion daemons, ensuring pods have been drained
+# and the node is marked as unschedulable.
+
+# If this fails we should ignore it and proceed anyway as Kubernetes will recover
+drain-kubelet:
+  cmd.run:
+    - name: |
+        kubectl drain {{ grains['fqdn'] }} --ignore-daemonsets --grace-period=300 --timeout=340s
+    - env:
+      - KUBECONFIG: {{ pillar['paths']['kubeconfig'] }}
+    - check_cmd:
+      - /bin/true
+
+kubelet:
+  service.dead:
+    - enable: False
+    - require:
+      - cmd: drain-kubelet
+
+kube-proxy:
+  service.dead:
+    - enable: False
+    - require:
+      - cmd: drain-kubelet
--- a/salt/etcd-proxy/stop.sls
+++ b/salt/etcd-proxy/stop.sls
@@ -1,0 +1,2 @@
+include:
+  - etcd.stop
--- a/salt/etcd/stop.sls
+++ b/salt/etcd/stop.sls
@@ -1,0 +1,4 @@
+# Stop and disable the etcd daemon
+etcd:
+  service.dead:
+    - enable: False
--- a/salt/flannel/stop.sls
+++ b/salt/flannel/stop.sls
@@ -1,0 +1,5 @@
+# Stop and disable the flannel daemon
+flannel:
+  service.dead:
+    - name: flanneld
+    - enable: False
--- a/salt/kubernetes-master/stop.sls
+++ b/salt/kubernetes-master/stop.sls
@@ -1,0 +1,12 @@
+# Stop and disable the Kubernetes master daemons
+kube-apiserver:
+  service.dead:
+    - enable: False
+
+kube-scheduler:
+  service.dead:
+    - enable: False
+
+kube-controller-manager:
+  service.dead:
+    - enable: False


salt/kubernetes-minion/stop.sls
```diff
@@ -0,0 +1,29 @@
...
+# and the node is marked as unschedulable.
+
+# TODO: If this, or the state below, fails we should ignore it and proceed anyway as
+# Kubernetes will recover.
+# TODO: || true is a horrible hack, salt doesn't appear to provide a mechanism beyond
+# exit codes to allow proceeding, even upon failure. Open to better suggestions.
+cordon-kubelet:
+  cmd.run:
+    - name: |
+        kubectl --kubeconfig /var/lib/kubelet/kubeconfig cordon {{ grains['id'] }}.{{ pillar['internal_infra_domain'] }} || :
```
ereslibre: Instead of the `||`, `check_cmd` can be used:

```
- check_cmd:
  - /bin/true
```

kiall: Yea, that's close enough to what I was looking for and should do the trick :)

kiall:
This has a bunch of cmd.run's pointed at kubectl, I wanted to replace those with salt.states.k8s, however, neither salt.states.k8s or salt.modules.k8s supports the tasks we're performing. Rather than try implement these ourselves in a rush, it may be best to leave these as is for now (e.g. I believe much of the drain logic is actually built into kubectl, rather than any of the daemons, so it would be a larger chunk of work than it seems at first glance to implement in salt.modules.k8s)

salt/kubernetes-minion/stop.sls
```diff
@@ -0,0 +1,31 @@
...
+  cmd.run:
+    - name: |
+        kubectl --kubeconfig /var/lib/kubelet/kubeconfig cordon {{ grains['id'] }}.{{ pillar['internal_infra_domain'] }}
+    - check_cmd:
+      - /bin/true
+
+drain-kubelet:
+  cmd.run:
+    - name: |
+        kubectl --kubeconfig /var/lib/kubelet/kubeconfig drain {{ grains['id'] }}.{{ pillar['internal_infra_domain'] }} --ignore-daemonsets --grace-period=300 --timeout=340s
```
inercia: I've always thought `drain` implies a `cordon`. From the [documentation for `drain`](https://kubernetes.io/docs/user-guide/kubectl/v1.6/#drain): _"Drain node in preparation for maintenance.The given node will be marked unschedulable to prevent new pods from arriving."_

salt/kubernetes-minion/init.sls
```diff
@@ -49,14 +49,12 @@ kubelet:
...
-#      - kmod: br_netfilter
-#
-#br_netfilter:
-#  kmod.present
+  # TODO: This needs to wait for the node to register, which takes a few seconds.
+  # Salt doesn't seem to have a retry mechanism in the version were using, so I'm
+  # doing a horrible hack right now and open to suggestions for a better workaround
+  cmd.run:
+    - name: |
+        sleep 30 && kubectl --kubeconfig /etc/kubernetes/config uncordon {{ grains['id'] }}.{{ pillar['internal_infra_domain'] }}
```
inercia: I guess the most elegant solution would be to implement a custom `beacon` that could send a _"node registered"_ event, [wait for that event](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.saltmod.html#salt.states.saltmod.wait_for_event), and then run the `kubectl uncordon`.

Otherwise maybe you can just run a `until ... do ...` command...

salt/docker/stop.sls
```diff
@@ -0,0 +1,4 @@
+# Stop and disable the docker daemon
+docker:
+  service.dead:
+    - enable: False
```
PI-Victor: does this use systemd? if not, if not we need to add containerd as well as part of the shutdown process, otherwise the node will hang because containerd exists before the running containers causing the node to just hang.

kiall: Salt will "pick" the init system to use based on what it sees installed, so should use systemd unless there are some horrible bugs in salt ;)

Should we be stopping containerd before or after docker to avoid this? Ill update to do this explicitly.

salt/kubernetes-minion/stop.sls
kiall: Looking at the current docs, I think your correct. I don't think older versions of Kubernetes did though, either way .. I'll check tomorrow.

salt/docker/stop.sls
kiall: As this is using systemd, and I'm not 100% certain what  the issue is + I'm not seeing any issues.. I've left this out for now.

salt/kubernetes-minion/init.sls
```diff
@@ -62,13 +62,16 @@ kubelet:
...
-#
-#br_netfilter:
-#  kmod.present
+  # TODO: This needs to wait for the node to register, which takes a few seconds.
+  # Salt doesn't seem to have a retry mechanism in the version were using, so I'm
+  # doing a horrible hack right now.
+  cmd.run:
+    - name: |
+        NEXT_WAIT_TIME=0
+        until kubectl --kubeconfig /var/lib/kubelet/kubeconfig uncordon {{ grains['fqdn'] }} || [ $NEXT_WAIT_TIME -eq 10 ]; do
```
grahamhayes: the only issue with this is that it will actually mark the task as successful, even when it has failed, and timed out.

kiall: I'll fix this up.

kiall: Done

salt/kubernetes-minion/init.sls
```diff
@@ -62,13 +62,21 @@ kubelet:
...
+        NEXT_WAIT_TIME=0
+        UNCORDON_RET=1
+
+        until [ $UNCORDON_RET -eq 0 ] || [ $NEXT_WAIT_TIME -eq 10 ]; do
+          kubectl --kubeconfig {{ pillar['paths']['kubeconfig'] }} uncordon {{ grains['fqdn'] }}
+          UNCORDON_RET=$?
+          sleep $(( NEXT_WAIT_TIME++ ))
+        done
+
+        exit $UNCORDON_RET
```
inercia: You could use the `kubeconfig`path we have in the path, and it could be even included in the environment with:

```yaml
    - env:
      - KUBECONFIG: {{ pillar['paths']['kubeconfig'] }}
```

Then you could wait with something like:

```bash
    elapsed=0
    until kubectl uncordon {{ grains['fqdn'] }} ; do
        [ $elapsed -gt 60 ] && exit 1
        sleep 1 && elapsed=$(( $elapsed + 1 ))
     done
```

kiall: Yea, thats cleaner.

kiall: Done + tested it works.. 

First attempt:

	"stderr": "Error from server (NotFound): nodes \"a1efe6040a1646b99188dc60dd52a526.infra.caasp.local\" not found", 

Second (and last) attempt:

	#"stdout": "node \"a1efe6040a1646b99188dc60dd52a526.infra.caasp.local\" already uncordoned"

PR #87, opened 2017-05-25
docker: reload container-feeder after starting docker

mssola:

### DIFF
--- a/salt/top.sls
+++ b/salt/top.sls
@@ -13,6 +13,7 @@ base:
     - etcd-proxy
     - flannel
     - docker
+    - container-feeder
   'roles:kube-master':
     - match: grain
     - kubernetes-master
--- a/salt/container-feeder/init.sls
+++ b/salt/container-feeder/init.sls
@@ -1,0 +1,15 @@
+include:
+  - docker
+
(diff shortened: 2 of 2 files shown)


flavio:
LGTM. One really minor note, maybe we should have a dedicated state for container-feeder to make it easier to be tracked (and removed) later on.

salt/docker/init.sls
```diff
@@ -49,4 +49,3 @@ docker:
       - pkg: docker
       - file: /etc/sysconfig/docker
       - /etc/systemd/system/docker.service.d/proxy.conf
-
```
flavio: This is an accidental change, please revert it

salt/top.sls
```diff
@@ -13,6 +13,7 @@ base:
     - etcd-proxy
     - flannel
     - docker
+    - container-feeder
```
flavio: I think `container-feeder` is needed only on the minions (BTW we should really change the name to worker, but not right now...). Am I wrong?

inercia: We need `container-feeder` in all the machines where `docker` is installed, so I think it is fine as it is.

PR #88, opened 2017-05-26
Minor changes in etcd and kubelet

inercia:
Minor changes in etcd and kubelet:

* do not remove `/var/lib/etcd`
* close some ports we don't really need for etcd, and open the port `10250` for `kubelet`.

### DIFF
--- a/salt/etcd/init.sls
+++ b/salt/etcd/init.sls
@@ -32,10 +32,6 @@ etcd:
       - etcd
     - require:
       - file: /etc/zypp/repos.d/containers.repo
-  cmd.run:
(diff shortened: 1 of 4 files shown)


salt/etcd/init.sls
```diff
@@ -32,10 +32,6 @@ etcd:
       - etcd
     - require:
       - file: /etc/zypp/repos.d/containers.repo
-  cmd.run:
-    - name: rm -rf /var/lib/etcd/*
-    - prereq:
-      - service: etcd
```
grahamhayes: :+1:

PR #89, opened 2017-05-29
Make cert always include `fqdn`

ereslibre:
Make cert always include `fqdn`
    
The only component that was adding `fqdn` to the list of dns names of SAN
certificates is the `kube-master` role.
    
However, depending on the size of the cluster and other possible reasons
it might happen that a etcd member falls in a `kube-minion` instance,
where the certificate is missing local ip addresses, as well as the `fqdn`
of the machine. With this change, we are enforcing `cert` to always
generate this information automatically, while we still allow to extend
it, in case that's still necessary (for example, as kubernetes-master still
requires).
    
Check https://bugzilla.novell.com/show_bug.cgi?id=1039269#c9 for further
information.
    
Fixes: bsc#1039269

### DIFF
--- a/salt/kubernetes-master/init.sls
+++ b/salt/kubernetes-master/init.sls
@@ -3,7 +3,7 @@ include:
   - cert
   - etcd-proxy
 
-{% from 'cert/init.sls' import ip_addresses %}
+{% from 'cert/init.sls' import ip_addresses, extra_names %}
 
 {% do ip_addresses.append("IP: " + pillar['api']['cluster_ip']) %}
 {% for _, interface_addresses in grains['ip4_interfaces'].items() %}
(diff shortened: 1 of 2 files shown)


config/master.d/returner.conf
```diff
@@ -5,7 +5,6 @@ mysql:
   # on a host named 'salt'
   host: 'localhost'
   user: 'root'
-  pass: 'salt'
```
ereslibre: Please, ignore this. I will rebase against `master` before merging.

ereslibre:
This will probably fix https://bugzilla.suse.com/show_bug.cgi?id=1041298 too.

jordimassaguerpla:
This makes sense. I haven't tested it thought and I am not expert on salt, but I'll trust you on this.

jordimassaguerpla:
I meant: LGTM

salt/kubernetes-master/init.sls
```diff
@@ -16,9 +16,8 @@ include:
 {% endfor %}
 
 # add some extra names the API server could have
-{% set extra_names = ["DNS: " + grains['fqdn'],
```
inercia: Why don't we keep `grains['fqdn']`?

ereslibre: We keep it, we are just appending extra names to it.

PR #90, opened 2017-05-29
Initial framework of update orchestration

grahamhayes:


grahamhayes:
Pulling in #85 underneath to test both

salt/orch/update.sls
```diff
@@ -0,0 +1,121 @@
...
+  salt.function:
+    - tgt: {{ master_id }}
+    - name: test.ping
+
+{{ master_id }}-clean-shutdown-kubernetes:
+  salt.state:
+    - tgt: {{ master_id }}
+    - sls:
+      - kubernetes-master.stop
+    - concurrent: False
```
kiall: Any reason this has `concurrent: False` and the others don't? Just curious if theres a reason it's needed here.

salt/orch/update.sls
```diff
@@ -0,0 +1,121 @@
...
+  salt.wait_for_event:
+    - name: salt/minion/*/start
+    - id_list:
+      - {{ master_id }}
+
+# Start services
+{{ master_id }}-start-services:
+  salt.function:
+    - tgt: {{ master_id }}
+    - name: test.ping
```
kiall: We should just be able to run a highstate here, right?

salt/orch/update.sls
kiall: Also, I suspect the 3x clean shutdown's can be combined by just listing the 3x SLS's in the right order. I could be wrong though...

salt/orch/update.sls
```diff
@@ -0,0 +1,121 @@
+# Generic Updates
+update_pillar:
+  salt.function:
+    - tgt: '*'
+    - name: saltutil.refresh_pillar
```
kiall: We'll also want a `saltutil.refresh_grains` pillar

salt/orch/update.sls
```diff
@@ -0,0 +1,121 @@
...
+
+update_modules:
+  salt.function:
+    - name: saltutil.sync_modules
+    - tgt: '*'
+    - kwarg:
+        refresh: True
+
+# Get list of masters needing reboot
+{%- set masters = salt.saltutil.runner('mine.get', tgt='G@roles:kube-master and G@upgrade:needs', fun='network.ip_addrs', tgt_type='compound') %}
```
kiall: It looks like the grain we'll want will be called:

    tx_update_reboot_needed: true

salt/orch/update.sls
grahamhayes: yea - I thought the 2 of them (master + minion) were in sync, will update

salt/orch/update.sls
grahamhayes: `concurrent: False` - No, I suspect copypasta error :) 

Combining them - We should be able to - will test it and see

salt/orch/update.sls
grahamhayes: Ack

salt/kubernetes-minion/stop.sls
```diff
@@ -0,0 +1,31 @@
+# Stop and disable the Kubernetes minion daemons, ensuring pods have been drained
+# and the node is marked as unschedulable.
+
+# If this, or the state below, fails we should ignore it and proceed anyway as
+# Kubernetes will recover.
+cordon-kubelet:
+  cmd.run:
+    - name: |
+        kubectl --kubeconfig /var/lib/kubelet/kubeconfig cordon {{ grains['id'] }}.{{ pillar['internal_infra_domain'] }}
```
ereslibre: I think you can use the `grain['fqdn']` here instead of concatenating `id` and `internal_infra_domain`.

salt/kubernetes-minion/stop.sls
```diff
@@ -0,0 +1,31 @@
...
+    - check_cmd:
+      - /bin/true
+
+drain-kubelet:
+  cmd.run:
+    - name: |
+        kubectl --kubeconfig /var/lib/kubelet/kubeconfig drain {{ grains['id'] }}.{{ pillar['internal_infra_domain'] }} --ignore-daemonsets --grace-period=300 --timeout=340s
```
ereslibre: Same here...

salt/kubernetes-minion/stop.sls
ereslibre: `kubectl drain` should already `cordon` the node.

salt/orch/update.sls
ereslibre: It's good to be safe by doing it (nothing wrong would happen for doing it), but I don't think it's necesary to `refresh_grains`, upon start, `salt-minion` is the very first thing it will do.

salt/kubernetes-minion/stop.sls
kiall: Graham sucked in this code from my PR, I'll update in #85

salt/kubernetes-minion/stop.sls
ereslibre: You explicitly indicated here to `--ignore-daemonsets`, I think they will be ignored anyway by default. Do we want to use `--force` here, so we ensure that we evict all the pods?

https://lukemarsden.github.io/docs/user-guide/kubectl/kubectl_drain/

salt/kubernetes-minion/stop.sls
kiall: Already done in in #85 (Graham sucked this change in from there to make use of it)

salt/kubernetes-minion/stop.sls
kiall: Graham sucked in this code from my PR, I'll update in #85

salt/orch/update.sls
inercia: Not sure but I think `concurrent` doesn't have any effect here as the `tgt` is a particular Salt minion...

salt/orch/update.sls
kiall: The TX update systemd timer will change this flag after the machine has booted and salt has been started. When we notice a machine is out of date, and start this play, we should explicitly refresh all machines so we have as up-to-date info as possible!

salt/orch/update.sls
```diff
@@ -0,0 +1,121 @@
...
+  salt.state:
+    - tgt: {{ master_id }}
+    - sls:
+      - flannel.stop
+
+{{ master_id }}-clean-shutdown-etcd:
+  salt.state:
+    - tgt: {{ master_id }}
+    - sls:
+      - etcd.stop
```
inercia: I would merge some of these states so you can keep a compact form. For example:

```yaml
{{ master_id }}-clean-shutdown-stuff:
  salt.state:
    - tgt: {{ master_id }}
    - sls:
      - flannel.stop
      - etcd.stop
```

salt/kubernetes-minion/stop.sls
kiall: `--ignore-daemonsets` is required, without it, any nodes with a DeamonSet  will just hang on the drain until the timeout hits.

`--force`  would allow for fully removing all pods, however, that's not necessarily what we're going for. Using these flags would result in less pods in the system after the upgrade, i.e. we deleted things which Kubernetes can't resurrect itself.

We're instead migrating anything we can away from this node, but manual pods can't be migrated. For those pods, which are hopefully really really rare, we just want to let it fail/timeout on these, continue anyway, and kubernetes will resurrect them on the same node when it starts back up.

salt/orch/update.sls
grahamhayes: and, it seems this was added in 2016.11.4.

https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.saltutil.html#salt.modules.saltutil.refresh_grains

kiall: Heh. Isn't this the version we're moving to anyway?

grahamhayes: yeah - but we do not have it yet -  so it fails right now with it left in

salt/orch/update.sls
grahamhayes: It shouldn't - it is removed in the next patch

salt/orch/update.sls
```diff
@@ -0,0 +1,117 @@
...
+    - id_list:
+      - {{ master_id }}
+
+# Start services
+{{ master_id }}-start-services:
+  salt.state:
+    - tgt: {{ master_id }}
+    - highstate: True
+
+{{ master_id }}-update-grain:
```
kiall: Does this update `/etc/salt/grains` (as that's the source of this particular grain?), or does it store it somewhere else, potentially permanently overriding the value from `/etc/salt/grains` which transactional-update will set for us?

grahamhayes: https://github.com/saltstack/salt/blob/develop/salt/modules/grains.py#L377 

From what I can see it will overwrite the config file with a new grains file, with this added.

grahamhayes: It makes the file look like this:

```
cat /etc/salt/grains 
roles:
- kube-minion
tx_update_reboot_needed: []
```

kiall: Humm, was the key present in the first place?

Either way - maybe it's best to set it to false instead of removing it?

grahamhayes: No, the key was not there before.

It doesn't really make a difference - the selector is for `tx_update_reboot_needed:true` so in the event it is not present, or an empty array (?) it will not match

grahamhayes: Salt considers the above syntax as "not present" - eg:

```
➜  salt git:(update-orchestration) ✗ docker exec -it $(docker ps | grep salt-master | awk '{print $1}') salt '*' grains.get tx_update_reboot_needed 
f680772aba404d8289422cbcbb219b0e:
967985d9f3f0493da8534ff5fc20bd8b:
da332a66af154892ab8f107e48af61bd:
ca:
    - True
0bf56584c1d74810a28b653c2321672e:
21a0aaef1af2486db7d96bd4fab57b31:
```

kiall: It could easily make a difference to the transactional-update code that updates this file and sets the flag to true! better to have it always true/false, rather than true or "falsy"

salt/orch/update.sls
```diff
@@ -0,0 +1,117 @@
+# Generic Updates
+update_pillar:
+  salt.function:
+    - tgt: '*'
+    - name: saltutil.refresh_pillar
+
+# TODO: When we update to 2016.11.4 uncomment this
```
mssola: Could you also add the reasoning ? Possibly with a link to the bug upstream.

salt/orch/update.sls
grahamhayes: Sure - updated to `setval`

However - whatever is setting the grains **really** should use `salt-call grains.setval tx_update_reboot_needed true` to ensure we salts methods for setting grains

salt/orch/update.sls
```diff
@@ -0,0 +1,117 @@
...
+{%- set masters = salt.saltutil.runner('mine.get', tgt='G@roles:kube-master and G@tx_update_reboot_needed:true', fun='network.ip_addrs', tgt_type='compound') %}
+
+{%- for master_id in masters.keys() %}
+
+{{ master_id }}-clean-shutdown:
+  salt.state:
+    - tgt: {{ master_id }}
+    - sls:
+      - kubernetes-master.stop
+      - flannel.stop
```
inercia: I think docker is running in the masters too... shouldn't we include `docker.stop` here?

grahamhayes: oh - is it? I will add that now.

mssola: If that is the case, then the orchestration of masters becomes the same as the orchestration of workers, no ? I wonder then if it would make sense to simply merge both "branches" (we could handle small differences in each specific salt state or something like t that ...)

salt/orch/update.sls
```diff
@@ -0,0 +1,117 @@
...
+
+{{ worker_id }}-update-grain:
+  salt.function:
+    - tgt: {{ worker_id }}
+    - name: grains.setval
+    - arg:
+      - tx_update_reboot_needed
+      - false
+    - require:
+      - salt: {{ worker_id }}-start-services
```
inercia: Maybe they are not strictly necessary but I would include some `require` for forcing some order... otherwise, you could add a compact form:

```SaltStack
{{ worker_id }}-bootup:
  salt.wait_for_event:
    - name: salt/minion/*/start
    - id_list:
      - {{ worker_id }}
  salt.state:
    - tgt: {{ worker_id }}
    - highstate: True
  salt.function:
    - tgt: {{ worker_id }}
    - name: grains.setval
    - arg:
      - tx_update_reboot_needed
      - false
```

grahamhayes: Actually, we need to add requires, to make sure that the grain is not removed even without a reboot, will add that now

PR #92, opened 2017-06-01
Idempotent salt states

ereslibre:
- Adapt some `cmd.run` to use `onchanges`, so they only execute when their
  `watched` states change.

- Add `stateful: True` to some `cmd.run`s, so following the salt protocol
  for this we ensure that the command didn't change anything in the system
  state.

- Move `ca-cert` to its own SLS, so `cert` will only now generate the
  `/etc/pki/minion.{key,crt}` files.

- The `cert` SLS will now be the only responsible for generating
  certificates depending on the role of the machine. This way we ensure
  that without mattering how this SLS is included it behaves in the same
  way under all conditions. We might want to use a certificate for different
  services, but that will need some extra changes.

- Change some `module.run` to `module.wait` so they only execute when the
  `watched` states change.

- Remove cleanups that make it impossible to have idempotent states.

### DIFF
--- a/salt/docker/init.sls
+++ b/salt/docker/init.sls
@@ -16,6 +16,7 @@ include:
         Environment="NO_PROXY={{ salt['pillar.get']('proxy:no_proxy', '') }}"
   cmd.run:
     - name: systemctl daemon-reload
+    - stateful: True
 
 #######################
 # docker daemon
@@ -40,16 +41,13 @@ docker:
     - append_if_not_found: True
     - require:
       - pkg: docker
-  # [inercia] when dockerd was already running and we require the
-  # service to be "service.running", Salt does not think it must
-  # restart it even when we say "watch these files", so we
-  # need this "cmd.run"...
   cmd.run:
     - name: systemctl restart docker.service
     - onlyif: systemctl status docker.service
+    - onchanges:
+      - /etc/systemd/system/docker.service.d/proxy.conf
     - require:
       - file: /etc/sysconfig/docker
-      - /etc/systemd/system/docker.service.d/proxy.conf
   service.running:
     - enable: True
     - watch:
@@ -57,4 +55,3 @@ docker:
       - pkg: docker
       - file: /etc/sysconfig/docker
       - /etc/systemd/system/docker.service.d/proxy.conf
-
--- a/salt/kubernetes-minion/init.sls
+++ b/salt/kubernetes-minion/init.sls
@@ -68,12 +68,14 @@ kubelet:
   cmd.run:
     - name: |
         ELAPSED=0
-        until kubectl uncordon {{ grains['caasp_fqdn'] }} ; do
+        until output=$(kubectl uncordon {{ grains['caasp_fqdn'] }}) ; do
             [ $ELAPSED -gt 60 ] && exit 1
             sleep 1 && ELAPSED=$(( $ELAPSED + 1 ))
         done
+        echo changed="$(echo $output | grep 'already uncordoned' &> /dev/null && echo no || echo yes)"
     - env:
       - KUBECONFIG: {{ pillar['paths']['kubeconfig'] }}
+    - stateful: True
 
 #######################
 # config files
--- a/salt/reboot/init.sls
+++ b/salt/reboot/init.sls
@@ -2,6 +2,12 @@
 # Configuration for the reboot manager
 ##################################################
 
+include:
+  - etcd-proxy
+
+{% set reboot_uri = "http://127.0.0.1:2379/v2/keys/" + pillar['reboot']['directory'] + "/" +
+         pillar['reboot']['group'] %}
+
 # `max_holders` contains the maximum number of lock holders for the cluster. It
 # must comply with the optimal cluster size as defined here:
 #   https://coreos.com/etcd/docs/latest/v2/admin_guide.html
@@ -12,27 +18,15 @@
   {% set max_holders = max_holders // 2 %}
 {% endif %}
 
-# Cleanup any previous cluster information on /opensuse.org
-opensuseorg_cleanup:
-  pkg.installed:
-    - name: etcdctl
-  cmd.run:
-    - name: etcdctl
-            rm -r /opensuse.org
-    # ignore failures for this
-    - check_cmd:
-      - /bin/true
-
 # Initialize the `mutex` key as expected by the reboot manager.
 set_max_holders_mutex:
   pkg.installed:
     - name: curl
   cmd.run:
-    - name: curl -L -X PUT
-            http://127.0.0.1:2379/v2/keys/{{ pillar['reboot']['directory'] }}/{{ pillar['reboot']['group'] }}/mutex?prevExist=false
-            -d value="0"
+    - name: curl -L -X PUT {{ reboot_uri }}/mutex?prevExist=false -d value="0"
+    - onlyif: curl {{ reboot_uri }}/mutex?prevExist=false | grep -i "key not found"
     - require:
-      - cmd: opensuseorg_cleanup
+      - service: etcd
 
 # Initialize the `data` key, which is JSON data with: the maximum number of
 # holders, and a list of current holders.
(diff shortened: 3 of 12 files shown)


salt/docker/init.sls
```diff
@@ -40,21 +41,16 @@ docker:
...
-  # [inercia] when dockerd was already running and we require the
-  # service to be "service.running", Salt does not think it must
-  # restart it even when we say "watch these files", so we
-  # need this "cmd.run"...
-  cmd.run:
+  cmd.wait:
     - name: systemctl restart docker.service
-    - onlyif: systemctl status docker.service
+    - watch:
+      - file: /etc/systemd/system/docker.service.d/proxy.conf
```
inercia: As the previous comment says, this `watch` will not do what you expect it to do...

ereslibre: Yes, it works with `cmd.wait`, it's double checked.

salt/docker/init.sls
```diff
@@ -16,6 +16,7 @@ include:
         Environment="NO_PROXY={{ salt['pillar.get']('proxy:no_proxy', '') }}"
   cmd.run:
     - name: systemctl daemon-reload
+    - stateful: True
```
inercia: I think these `stateful` arguments will have no effect, unless the commands can provide what [Salt expects](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.cmd.html#using-the-stateful-argument).

ereslibre: This is still a WIP, that's why I didn't ask for review yet :), but in this context it will perfectly work, since no output is considered as "nothing changed" given the protocol definition, what perfectly fits in the context of a `daemon-reload`, IMO.

salt/docker/init.sls
inercia: Last time it tried it didn't. Docker will not be restarted if 1) a fresh installation happens and 2) the docker daemon is enabled on bootup.

In this case, removing a `systemctl status` will not help us make things idempotent and, in fact, I think that is the key for reaching that goal.

ereslibre: With the current code it will be restarted *always* on every orchestration/highstate.

Just check `journalctl -fu docker` and run a highstate on that machine, it is restarted everytime.

salt/docker/init.sls
inercia: So, from my understanding of the protocol, if `output=None` `=>` _no changes happened_ `=>` _watchers will not be notified_, is that right?

salt/docker/init.sls
inercia: Yeah, maybe it was wrong and a new `onlyif` is needed, but the `watch` will not fix it...

salt/docker/init.sls
ereslibre: In other situations I guess this is the case (still would have to check it), but we have no watchers listening to this `daemon-reload`, no?

inercia: The thing is nobody will be notified if, for example,  `salt['proxy:no_proxy']` changes (even when a different file has been written to disk) because of the empty output of `systemctl daemon-reload`.

ereslibre: I believe the `file.managed` and `cmd.run` are different actions, so the `cmd.run` shouldn't stop the `file.managed` from being watched by other states, and shouldn't swallow those events itself. However I will double check that so I can confirm, thank you.

The idea behind the stateful here was to avoid it from appearing as `changed` during a highstate or orchestration, so we can have automated tests that run the orchestration twice and receive no changes on the second run.

ereslibre: I can confirm this worked:

```
/tmp/somefile:
  file.managed:
    - source: /etc/systemd/system/docker.service.d/proxy.conf
    - watch:
      - file: /etc/systemd/system/docker.service.d/proxy.conf
```

When `/etc/systemd/system/docker.service.d/proxy.conf` changed its contents, `/tmp/somefile` did too, even with the `cmd.run` with `stateful: True`.

kiall: So, I think @inercia is right here... Given how salt orders things, this could happen (in theory):

1) We edit the file
2) Docker is restarted - old proxy info is still in use
3) Systemd is reloaded

when what we really want is:

1) We edit the file
2) Systemd is reloaded
3) Docker is restarted - new proxy info is now in use

ereslibre: Oh, yes... that *can* happen right now. Taking a look at the moment.

ereslibre: Just checked it: using `/etc/systemd/system/docker.service.d/proxy.conf` directly instead of `file: /etc/systemd/system/docker.service.d/proxy.conf` ensures that this runs also if the file changes (even with the `cmd.run` on that state), and the order is the correct one on all my tests.

No changes -> nothing is done
Changes -> write changes, daemon-reload, restart docker

inercia: Then what is the effect of `stateful: True`? The real chain of actions in not `write changes -> daemon-reload -> restart docker` but `write changes -> daemon-reload && restart docker`, so I think nobody cares about the changes produced by `daemon-reload`, so we wouldn't have to set `stateful: True`.

ereslibre: Salt will always assume a `cmd.run` will "change something", as it cannot know if something has changed when running that command. With `stateful: True` you acknowledge to follow their protocol, so salt will be able to know if a `cmd.run` did change anything in the system state or not.

Even with `stateful: True` and empty `stdout` that doesn't hide the fact that the `file.managed` `/etc/systemd/system/docker.service.d/proxy.conf` did actually change, so if you watch for `/etc/systemd/system/docker.service.d/proxy.conf` or `file: /etc/systemd/system/docker.service.d/proxy.conf` there will be changes, because at least one of the actions did change (in this case if contents change with `file.contents`).

salt/kubernetes-minion/init.sls
```diff
@@ -68,12 +68,18 @@ kubelet:
...
-        until kubectl uncordon {{ grains['caasp_fqdn'] }} ; do
+        until output=$(kubectl uncordon {{ grains['caasp_fqdn'] }}) ; do
             [ $ELAPSED -gt 60 ] && exit 1
             sleep 1 && ELAPSED=$(( $ELAPSED + 1 ))
         done
+        if echo $output | grep "already uncordoned" &> /dev/null; then
+          echo "changed=no"
+        else
+          echo "changed=yes"
+        fi
```
inercia: What about `echo changed="$(echo $output | grep 'already uncordoned' &> /dev/null && echo no || echo yes)"` ?

salt/reboot/init.sls
```diff
@@ -2,6 +2,9 @@
 # Configuration for the reboot manager
 ##################################################
 
+include:
+  - etcd-proxy
```
inercia: I thought we had decided not to include `etc-proxy` from the reboot setup stage....

salt/docker/init.sls
inercia: Ok, I think you are right.

salt/etc-hosts/init.sls
```diff
@@ -4,4 +4,4 @@
     - marker_end:   "#-- end Salt-CaaSP managed hosts --"
     - source:       salt://etc-hosts/hosts.jinja
     - template:     jinja
-    - append_if_not_found: True
+    - append_if_not_found: True
```
inercia: Spurious change.

salt/kubernetes-minion/init.sls
ereslibre: Yeah, thanks!

salt/reboot/init.sls
ereslibre: That was in a rush when releasing the latest beta version. With all these changes in this PR it's safe to bring this back.

inercia: Then it ok.

ereslibre:
Still, changes are detected when running an orchestration due to:

```
update_pillar:
  salt.function:
    - tgt: '*'
    - name: saltutil.refresh_pillar
    - require:
      - salt: hostname_setup

update_mine:
  salt.function:
    - tgt: '*'
    - name: mine.update
    - require:
      - salt: update_pillar

update_modules:
  salt.function:
    - tgt: '*'
    - name: saltutil.sync_modules
    - kwarg:
        refresh: True
```

But this is something I don't think we cannot fix. If you have any ideas, please speak up :smile:

salt/docker/init.sls
kiall: The use of stateful here still feels a little weird to me, however if it works, its not weird enough to block merge IMO.

PR #93, opened 2017-06-01
Don't rely on fqdn grain

kiall:
The FQDN grain is unrelable, and depending on the node (and it's current
state) will give us different values. It's use needs to be dropped entirely,

Fixes bsc#1041951
Partially Fixes bsc#1041789

### DIFF
--- a/salt/cert/init.sls
+++ b/salt/cert/init.sls
@@ -2,7 +2,7 @@ include:
   - crypto
 
 {% set ip_addresses = [] -%}
-{% set extra_names = ["DNS: " + grains['id'] + "." +  pillar['internal_infra_domain'] ] -%}
+{% set extra_names = ["DNS: " + grains['caasp_fqdn'] ] -%}
 
 {{ pillar['paths']['ca_dir'] }}:
   file.directory:
@@ -38,7 +38,7 @@ include:
     - ca_server: {{ salt['mine.get']('roles:ca', 'x509.get_pem_entries', expr_form='grain').keys()[0] }}
     - signing_policy: minion
     - public_key: /etc/pki/minion.key
-    - CN: {{ grains['id'] }}.{{ pillar['internal_infra_domain'] }}
+    - CN: {{ grains['caasp_fqdn'] }}
     - C: {{ pillar['certificate_information']['subject_properties']['C']|yaml_dquote }}
     - Email: {{ pillar['certificate_information']['subject_properties']['Email']|yaml_dquote }}
     - GN: {{ pillar['certificate_information']['subject_properties']['GN']|yaml_dquote }}
--- a/salt/cert/init.sls
+++ b/salt/cert/init.sls
@@ -2,7 +2,7 @@ include:
   - crypto
 
 {% set ip_addresses = [] -%}
-{% set extra_names = ["DNS: " + grains['fqdn']] -%}
+{% set extra_names = ["DNS: " + grains['id'] + "." +  pillar['internal_infra_domain'] ] -%}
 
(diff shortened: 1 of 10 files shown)


kiall:
Yea, I'm working on it.. This however fixes the P1 issue...

kiall:
Given it was so quick to do.. I just added the commit to this PR.

salt/cert/init.sls
```diff
@@ -2,7 +2,7 @@ include:
   - crypto
 
 {% set ip_addresses = [] -%}
-{% set extra_names = ["DNS: " + grains['fqdn']] -%}
+{% set extra_names = ["DNS: " + grains['caasp_fqdn'] ] -%}
```
ereslibre: We can leave this as empty list if the `CN` is the `caasp_fqdn`. Overlooked it before with the `fqdn` case.

salt/hostname/init.sls
```diff
@@ -1,13 +1,22 @@
...
+  grains.present:
+    - value: {{ grains['id'] }}.{{ pillar['internal_infra_domain'] }}
 
+# Both of the below are due to be removed, and can't use the `caasp_fqdn` grain
+# as it's not available in grains until the next state, so lets leave them as is
+# for the moment
 /etc/hostname:
   file.managed:
-    - contents: {{ hostname }}
+    - contents: {{ grains['id'] }}.{{ pillar['internal_infra_domain'] }}
```
ereslibre: Can't we use `grains['caasp_fqdn']` here?

salt/hostname/init.sls
```diff
@@ -1,13 +1,22 @@
...
     - backup: false
+    - require:
+      - grains: caasp_fqdn
 
 hostname-static:
   cmd.run:
-    - name: hostnamectl set-hostname --static --transient {{ hostname }}
-    - unless: [[ $(hostnamectl --transient) == {{ hostname }} ]]
+    - name: hostnamectl set-hostname --static --transient {{ grains['id'] }}.{{ pillar['internal_infra_domain'] }}
+    - unless: [[ $(hostnamectl --transient) == {{ grains['id'] }}.{{ pillar['internal_infra_domain'] }} ]]
```
ereslibre: Same on both cases, cannot we use `grains['caasp_fqdn']`?

salt/cert/init.sls
kiall: Many TLS implementations now ignore the certs CN when SANs are present, i.e. the CN should always be included as a SAN.

salt/hostname/init.sls
kiall: Nope, it doesn't exist in the grains dict built when this particular state starts, it only exists on the next state .. Given both of these are being removed Real Soon Now, I left them as is.

salt/hostname/init.sls
kiall: Same as above :)

salt/cert/init.sls
ereslibre: Okay, will not harm anyway. Let's be safe then :)

salt/hostname/init.sls
ereslibre: Then I have doubts about the `pillar.fqdn` mining... When we mine it, will it be there?

ereslibre: Oh, right. On the orchestration first comes the hostname sls and then the mine.update. Everything's fine :)

kiall: Yea, I really wanted a "non-runtime" way to set this, but that didn't seem possible

PR #95, opened 2017-06-01
Don't change the system hostname

kiall:
Operators don't want us to change the system hostname, which we previously
did to account for environments which don't provide unique DHCP hostnames.

We'll undo this change, as we have now removed our reliance on the system
default hostname.

Tested with:

* Unique + Resolvable Hostnames
* Unique + NOT Resolvable Hostnames
* NOT Unique + Resolvable Hostnames (hostname resolves to all 3 node IPs)
* NOT Unique + NOT Resolvable Hostnames

Fixes bsc#1041789

### DIFF
--- a/salt/hostname/init.sls
+++ b/salt/hostname/init.sls
@@ -1,22 +1,3 @@
 caasp_fqdn:
   grains.present:
     - value: {{ grains['id'] }}.{{ pillar['internal_infra_domain'] }}
-
-# Both of the below are due to be removed, and can't use the `caasp_fqdn` grain
-# as it's not available in grains until the next state, so lets leave them as is
-# for the moment
-/etc/hostname:
-  file.managed:
-    - contents: {{ grains['id'] }}.{{ pillar['internal_infra_domain'] }}
-    - backup: false
-    - require:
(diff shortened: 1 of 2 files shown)


grahamhayes:
we may need to change https://github.com/kubic-project/salt/blob/master/salt/motd/motd.jinja as well

kiall:
I thought about the MOTD, and figured it was better to leave it (as the generated name is still used by etcd and kubernets), so that provides a way to know whats what...

kiall:
Updated MOTD:

	Welcome!

	Machine ID: 94de83d05b8a47119ca3a40fac5f82e6
	External FQDN: b61b7c64d07e48dfba7303b9c60458a2.infra.caasp.local
	Internal FQDN: 94de83d05b8a47119ca3a40fac5f82e6.infra.caasp.local

	The roles of this node are:
	  - kube-minion

kiall:
And - salt's FQDN grain is even more unreliable than I thought!

https://github.com/saltstack/salt/issues/36619

Removing the "Internal FQDN" line so this can merge, can evaluate the fqdn grain separately.

PR #97, opened 2017-06-02
use git revision in package version

jordimassaguerpla:
this way zypper sees each new commit as an update
Otherwise, using the date, will create a conflict if 2 commits are from
the same day

Signed-off-by: Jordi Massaguer Pla <jmassaguerpla@suse.de>

### DIFF
--- a/packaging/suse/make_spec.sh
+++ b/packaging/suse/make_spec.sh
@@ -12,8 +12,9 @@ cd $(dirname $0)
 
 YEAR=$(date +%Y)
 VERSION=$(cat ../../VERSION)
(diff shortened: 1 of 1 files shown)


robdaemon:
This will get weird on builds made off PRs - you end up with `/` characters in the revision.

jordimassaguerpla:
@robdaemon : Sorry I don't get it. Why would you get a "/" in the revision?

robdaemon:
@jordimassaguerpla sorry, I was confusing the output of `git name-rev --name-only HEAD` with `git rev-parse --short HEAD` - carry on :)

PR #98, opened 2017-06-02
disable reboot manager when orchestration happens

PI-Victor:
this will disable rebootmgr when orchestration starts.
@ereslibre @inercia PTAL.
not sure if i'm missing something.
Initially i wanted to define the service as part of the state (this is just a variation, i did a lot of them):
```
disable_rebootmgr:
  salt.state:
    - service:
      - name: rebootmgr
      - running: False
      - enabled: False
      - reload: True
```
i think that defining a separate `role` just for this is stupid but i couldn't find any other way. suggestions are welcomed.

### DIFF
--- a/salt/rebootmgr/init.sls
+++ b/salt/rebootmgr/init.sls
@@ -1,0 +1,3 @@
+rebootmgr:
+  service.dead:
+    - enable: False
--- a/salt/orch/kubernetes.sls
+++ b/salt/orch/kubernetes.sls
@@ -1,3 +1,9 @@
+disable_rebootmgr:
+  salt.state:
+    - tgt: '*'
+    - sls:
+      - rebootmgr
+
 hostname_setup:
   salt.state:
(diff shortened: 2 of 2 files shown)


salt/rebootmgr/init.sls
```diff
@@ -0,0 +1,4 @@
+rebootmgr:
+  service.dead:
+    - enable: False
+    - reload: True
```
kiall: What does `reload: True` do in the context of a service.dead?

PI-Victor: nvm, i thought it was necessary. but it turns out it's not.

ereslibre: Agree with @kiall here.

ereslibre:
Agree with @kiall here, `highstates` independently should be able to keep state consistent.

PI-Victor:
great, thanks for the merge.

PR #103, opened 2017-06-06
Update transactional-update to use "salt" option

grahamhayes:
This will ensure that the transactional-update code will
write a grain (`tx_update_reboot_needed:true`) on the
node instead of rebooting the node.

This also allows for increasing the frequency of the snapshots
being built

### DIFF
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -75,3 +75,7 @@ proxy:
 reboot:
   group:          'default'
   directory:      'opensuse.org/rebootmgr/locks'
+
+transactional-update:
+  timer:
+    on_calendar: 'daily' # See https://www.freedesktop.org/software/systemd/man/systemd.time.html#Calendar%20Events for syntax
--- a/salt/transactional-update/10-update-rebootmgr-options.conf
+++ b/salt/transactional-update/10-update-rebootmgr-options.conf
@@ -1,0 +1,3 @@
+[Service]
+ExecStart=
+ExecStart=/usr/sbin/transactional-update cleanup dup salt
--- a/salt/transactional-update/init.sls
+++ b/salt/transactional-update/init.sls
@@ -1,0 +1,29 @@
+/etc/systemd/system/transactional-update.service.d/10-update-rebootmgr-options.conf:
+  file.managed:
+    - makedirs: true
+    - source: salt://transactional-update/10-update-rebootmgr-options.conf
+    - user: root
+    - group: root
+    - mode: 644
+
+/etc/systemd/system/transactional-update.timer.d/10-increase-update-speed.conf:
+  file.managed:
+    - makedirs: true
+    - template: jinja
+    - source: salt://transactional-update/10-increase-update-speed.conf.jinja
+    - user: root
+    - group: root
+    - mode: 644
+
+service.systemctl_reload:
+  module.run:
+    - onchanges:
+      - file: /etc/systemd/system/transactional-update.service.d/10-update-rebootmgr-options.conf
+      - file: /etc/systemd/system/transactional-update.timer.d/10-increase-update-speed.conf
+
+transactional-update.timer:
+  service.running:
+    - name: transactional-update.timer
+    - enable: True
+    - watch:
+      - file: /etc/systemd/system/transactional-update.timer.d/10-increase-update-speed.conf
--- a/salt/top.sls
+++ b/salt/top.sls
@@ -15,6 +15,7 @@ base:
     - flannel
     - docker
     - container-feeder
+    - transactional-update
   'roles:kube-master':
     - match: grain
     - kubernetes-master
--- a/salt/transactional-update/10-increase-update-speed.conf.jinja
+++ b/salt/transactional-update/10-increase-update-speed.conf.jinja
@@ -1,0 +1,2 @@
+[Timer]
+OnCalendar={{ salt['pillar.get']('transactional-update:timer:on_calendar', 'OnCalendar=daily') }} 


pillar/params.sls
```diff
@@ -75,3 +75,7 @@ proxy:
 reboot:
   group:          'default'
   directory:      'opensuse.org/rebootmgr/locks'
+
+transactional-update:
+  timer:
+    OnCalendar: '*:0/15' # See https://www.freedesktop.org/software/systemd/man/systemd.time.html#Calendar%20Events for syntax
```
grahamhayes: I just put a holding value here for now - we should decide how often we want to do this check

ereslibre: Would it make sense to take this value from the pillar?

ereslibre: Isn't this expanded to `*-*-* *:00/15:00`? I'm confused by the format TBH.

inercia: Why don't we follow the convention for pillar names? On don't think we have used _CamelCase_ so far...

salt/transactional-update/init.sls
```diff
@@ -0,0 +1,31 @@
...
+      - file: /etc/systemd/system/transactional-update.timer.d/10-increase-update-speed.conf
+
+transactional-update.timer:
+  service.running:
+    - name: transactional-update.timer
+    - enable: True
+    - require:
+      - file: /etc/systemd/system/transactional-update.timer.d/10-increase-update-speed.conf
+    - watch:
+      - file: /etc/systemd/system/transactional-update.timer.d/10-increase-update-speed.conf
```
inercia: I think that a `watch` implies a `require`

salt/transactional-update/init.sls
```diff
@@ -0,0 +1,31 @@
...
+    - mode: 644
+
+/etc/systemd/system/transactional-update.timer.d/10-increase-update-speed.conf:
+  file.managed:
+    - makedirs: true
+    - template: jinja
+    - source: salt://transactional-update/10-increase-update-speed.conf.jinja
+    - user: root
+    - group: root
+    - mode: 644
```
inercia: For such small files, maybe it would be better to embed them here with `contents`...

pillar/params.sls
grahamhayes: That is the option name in the file

grahamhayes: It is in the pillar - and yes - that is what it is expanded to. I can be more explicit if needed - but lets agree on a timescale.

salt/transactional-update/init.sls
grahamhayes: I prefer the readability of separate file, and the clean separation

salt/transactional-update/init.sls
grahamhayes: Yup - it does - removing require

pillar/params.sls
ereslibre: Oh... sorry... /me buys new :eyeglasses:

inercia: This is just a variable you are using in a template that is replaced inside a line that starts with `OnCalendar`. Otherwise, we would be using many different conventions depending on _where_ we are using those replacements...

salt/transactional-update/init.sls
inercia: Fair enough

grahamhayes:
OK - looks like systemd drop ins for timers don't override the `ExecStart` - they add an extra one. :man_facepalming:

pillar/params.sls
grahamhayes: Done

grahamhayes:
OK - looks like you just have to add a blank `ExecStart=` to clear the inherited value

grahamhayes:
Using the steps outlined on the ML earlier, #105 and manually triggering the update orchestration file, successfully created a new snapshot, stopped services, rebooted, and started back up.

pillar/params.sls
```diff
@@ -75,3 +75,7 @@ proxy:
 reboot:
   group:          'default'
   directory:      'opensuse.org/rebootmgr/locks'
+
+transactional-update:
+  timer:
+    on_calendar: '*-*-* 00/12:00:00' # See https://www.freedesktop.org/software/systemd/man/systemd.time.html#Calendar%20Events for syntax
```
inercia: _TODO_: it seems we could be creating too many snapshots and downloading too many times the same packages if we lower the frequency for this timer. @thkukuk please take a looks at [this log](https://gist.github.com/inercia/4592aff3651da259d2ea512339cf5f08)... is this a known behaviour in `Build22.17`?

thkukuk:
On Mon, Jun 12, Alvaro wrote:

> @inercia commented on this pull request.
> 
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> 
> In pillar/params.sls:
> 
> > @@ -75,3 +75,7 @@ proxy:
>  reboot:
>    group:          'default'
>    directory:      'opensuse.org/rebootmgr/locks'
> +
> +transactional-update:
> +  timer:
> +    on_calendar: '*-*-* 00/12:00:00' # See https://www.freedesktop.org/software/systemd/man/systemd.time.html#Calendar%20Events for syntax
> 
> TODO: it seems we could be creating too many snapshots and downloading too many
> times the same packages if we lower the frequency for this timer.

You should not run the transactional-update script more often than you
are able to reboot the machine. Everything else doesn't make any sense.
Once a day is normally enough.

> @thkukuk please take a looks at this log... is this a known behaviour in Build22.17?

Would be very helpfull to know what of the long output you mean ...


-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & CaaSP
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

pillar/params.sls
ereslibre: So maybe we can use `*-*-*`, so this is run daily at 00:00?

thkukuk:
On Tue, Jun 13, Rafael Fernández López wrote:

> @ereslibre commented on this pull request.
> 
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> 
> In pillar/params.sls:
> 
> > @@ -75,3 +75,7 @@ proxy:
>  reboot:
>    group:          'default'
>    directory:      'opensuse.org/rebootmgr/locks'
> +
> +transactional-update:
> +  timer:
> +    on_calendar: '*-*-* 00/12:00:00' # See https://www.freedesktop.org/software/systemd/man/systemd.time.html#Calendar%20Events for syntax
> 
> So maybe we can use *-*-*, so this is run daily at 00:00?

If you want to run it daily at 00:00, there is no need to change the current
configuration for the timer at all.

-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & CaaSP
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

salt/transactional-update/10-update-rebootmgr-options.conf
```diff
@@ -0,0 +1,3 @@
+[Service]
+ExecStart=
+ExecStart=/usr/sbin/transactional-update cleanup dup salt
```
ereslibre: If we are going to fallback to 1 daily run (if we do it at 00:00), wouldn't this be the only required change? (along with the `salt/top.sls` change and `salt/transactional-update/init.sls` change (but only including `/etc/systemd/system/transactional-update.service.d/10-update-rebootmgr-options.conf` state)

grahamhayes: I think we should leave it at 12hrs - we do not currently have a an auto update timer, when we get that we can set the pillar value to whatever that is.

Right now, if we have a 24hr window, and a 24 hour update window, it could 48 hours before an update gets applied.

For normal updates this might be fine, but there is some updates that will need much faster deployments. (libssl and similar)

inercia: I think @grahamhayes is right: the `transactional-updates` should be run at a higher frequency, but it should not have nasty side effects like creating multiple snapshots or downloading the same packages again and again... We should tackle these problems after 1.0.

ereslibre: > For normal updates this might be fine, but there is some updates that will need much faster deployments. (libssl and similar)

Those should be treated as critical and 12 hours can also be a really bad window for something with such an impact.

Maybe we can skip calling `transactional-update` (thus, avoiding an extra snapshot) if we check on the systemd unit that the grain is already there. If it is, we skip calling transactional-update. This would skip intermediate updates since the last snapshot though.

Just wanted to point another possible approach, but I'm okay with your reasoning given where we are.

thkukuk:
On Tue, Jun 13, Alvaro wrote:

> I think @grahamhayes is right: the transactional-updates should be run at a
> higher frequency, but it should not have nasty side effects like creating
> multiple snapshots or downloading the same packages again and again... We
> should tackle these problems after 1.0.

Sorry, but we are speaking here about transactional updates. So we 
clearly not change that after 1.0.

  Thorsten

-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & CaaSP
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

thkukuk:
On Tue, Jun 13, Rafael Fernández López wrote:

> @ereslibre commented on this pull request.
> 
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> 
> In salt/transactional-update/10-update-rebootmgr-options.conf:
> 
> > @@ -0,0 +1,3 @@
> +[Service]
> +ExecStart=
> +ExecStart=/usr/sbin/transactional-update cleanup dup salt
> 
>     For normal updates this might be fine, but there is some updates that will
>     need much faster deployments. (libssl and similar)
> 
> Those should be treated as critical and 12 hours can also be a really bad
> window for something with such an impact.

You can run transactional-update as often as you want a day, but you
need to make sure that you afterwards reboot the machine.
It doesn't make any sense to run transactional-update on a hourly base,
if you only reboot the machine once a day.

  Thorsten

-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & CaaSP
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

inercia:
> You can run transactional-update as often as you want a day, but you
need to make sure that you afterwards reboot the machine.
It doesn't make any sense to run transactional-update on a hourly base,
if you only reboot the machine once a day.

I think that is not what we want to do. We want to run the `transactional-update` and flag the machine as `needs reboot` (by using the Salt grain) and then reboot the machines in the right order in the _update window_.

So I don't think we want to just run `transactional-update up reboot` every 15 mins (as it seems that, in the presence of updated, it creates a new snapshot every time and downloads packages again and again), and we don't want to do that in the update window either (as it would be better to spread the downloads in a random way during the day).

kiall:
Daily seems fine to me, post 1.0 we could add the ability to run it on demand easily enough to cater for the getting the next HeartBleed style CVE fix ASAP.

> If you want to run it daily at 00:00, there is no need to change the current
configuration for the timer at all.

True, but allowing customers to choose how often this runs would be nice - so we should copy the default systemd unit's OnCalendar here, allowing for the timer to be overridden if needed? Or only template this file if it the pillar has non-null value.

---

@thkukuk would it make sense for transactional-update to refuse to create a new snapshot (or to discard new snapshots) if an already pending snapshot would be the same? I imagine this would be useful for anyone using transactional-update to avoid snapshots building up unnecessarily.

thkukuk:
On Tue, Jun 13, Alvaro wrote:

>     You can run transactional-update as often as you want a day, but you
>     need to make sure that you afterwards reboot the machine.
>     It doesn't make any sense to run transactional-update on a hourly base,
>     if you only reboot the machine once a day.
> 
> I think that is not what we want to do. We want to run the transactional-update
> and flag the machine as needs reboot (by using the Salt grain) and then reboot
> the machines in the right order in the update window.

Which means, you should take the start of the update window, subtract the
guessed time the update needs in average, and run transactional-update at
that time. Like we do in the default configuration.

> So I don't think we want to just run transactional-update up reboot every 15
> mins (as it seems that, in the presence of updated, it creates a new snapshot
> every time and downloads packages again and again), and we don't want to do
> that in the update window either (as it would be better to spread the downloads
> in a random way during the day).

So you should run it once a day only nearly before the maintenance window
starts.

  Thorsten

-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & CaaSP
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

grahamhayes:
OK, updated to "daily" which was the initial value.

We can edit the pillar later if we need to modify it.

salt/transactional-update/10-update-rebootmgr-options.conf
```diff
@@ -0,0 +1,3 @@
+[Service]
+ExecStart=
```
PI-Victor: what's with the empty ExecStart?

grahamhayes: To override the whole `ExecStart` instead of appending to it, systemd insists on a blank one.

grahamhayes:
@ereslibre had +A'd - I needed to `git push -f` to trigger CI, so I am taking it as an approval :)

PR #105, opened 2017-06-06
Change update orchestration to use systemctl

grahamhayes:
We were using rebootmgr, but with that service disabled
we now need to use systemctl.

Salt has a bug where it will mark any call to `system.reboot` as
failed when a system actually reboots, so we cannot use the
built-in salt methods.

See saltstack/salt#24297 and saltstack/salt#31557

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -44,7 +44,7 @@ update_modules:
     - tgt: {{ master_id }}
     - name: cmd.run
     - arg:
-      - rebootmgrctl reboot now
+      - systemctl reboot
     - require:
       - salt: {{ master_id }}-clean-shutdown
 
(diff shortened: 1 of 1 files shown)


inercia:
So does `systemctl reboot` exactly the same as `rebootmgrctl reboot now`?

grahamhayes:
@inercia yes - it triggers a reboot straight away - but the minion has enough time to return the status code.

grahamhayes:
@kiall / @flavio - can we merge this with out e2e tests? It will not get used by e2e tests. We are just replacing the call to a salt module with a call to the raw command.

kiall:
Going to see how concourse reacts to a close + reopen.

PR #107, opened 2017-06-09
Option for using the proxy settings system-wide (bsc#1036627)

inercia:
We will use the proxy settings system-wide when `pillar["proxy:systemwide"]` is set to `true` (enabled by default).

### How to test it

* invoke `k8s-setup` with `-Vpillar=proxy:http~http://error.com:8080,proxy:https~https://error.com:443`. This will add some additional Pillar values for setting the proxy.
* create the cluster
* orchestrate
* `ssh` into the machines and verify that `/etc/sysconfig/proxy` has the values we set in the Pillar
* from one of the master/minions, do `curl www.suse.com` and verify it fails (because the proxy is incorrect)
* from one of the master/minions, do `curl <dashboard-IP>` and verify it dumps the HTML code for the Dashboard.
* from one of the master/minions, do `docker pull alpine` and verify it fails.

### DIFF
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -65,6 +65,7 @@ proxy:
   http:           ''
   https:          ''
   no_proxy:       ''
+  systemwide:     'true'
 
 # Configuration for the reboot manager (https://github.com/SUSE/rebootmgr).
 # notes:
--- a/salt/proxy/init.sls
+++ b/salt/proxy/init.sls
@@ -1,0 +1,12 @@
+{% if salt['pillar.get']('proxy:systemwide', '').lower() == 'true' %}
+
+/etc/sysconfig/proxy:
+  file.managed:
+    - makedirs: True
+    - contents: |
+        PROXY_ENABLED="yes"
+        HTTP_PROXY="{{ salt['pillar.get']('proxy:http', '') }}"
+        HTTPS_PROXY="{{ salt['pillar.get']('proxy:https', '') }}"
+        NO_PROXY="{{ pillar['dashboard'] }},{{ salt['pillar.get']('proxy:no_proxy', '') }}"
+
+{% endif %}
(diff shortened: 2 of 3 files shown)


flavio:
Great, this is going to address the concerns I raised [here](https://bugzilla.suse.com/show_bug.cgi?id=1036627#c11).

I know this is still WIP, but shouldn't we avoid the `systemwide` pillar and just rely on having a value for either `proxy:http` or `proxy:https`?

inercia:
> I know this is still WIP, but shouldn't we avoid the systemwide pillar and just rely on having a value for either proxy:http or proxy:https?

I think having `proxy:http` or `proxy:https` does not imply we want those settings systemwide. Having the `systemwide` pillar gives us the flexibility to change the behaviour and just use these proxy settings for Docker (for example, if we consider that autoyast is enough)

flavio:
I still prefer to have salt overwrite what was written with AutoYaST. At the end of the day, once a node joins the cluster, salt must be the only source of truth.

inercia:
> I still prefer to have salt overwrite what was written with AutoYaST. At the end of the day, once a node joins the cluster, salt must be the only source of truth.

Yeah, the settings in Salt are what really matters at the end.

And regarding the `systemwide` pillar, we decided not to enable/disable the proxy setting based on `http_proxy != ""` when we solved the problem for docker because users could decide to reconfigure the cluster and set `http_proxy = ""`, and I think that still applies...

pillar/params.sls
```diff
@@ -65,6 +65,7 @@ proxy:
   http:           ''
   https:          ''
   no_proxy:       ''
+  systemwide:     'true'
```
ereslibre: Can't this value be a real boolean?

inercia:
(re-review requested from @grahamhayes @ereslibre due to `push -f`)

PR #111, opened 2017-06-12
Added salt files to manage the admin node

mssola:
Since kubic-project/velum#185, the admin node is managed by Salt. For
now we handle it in the same way as the CA node.

### DIFF
--- a/salt/admin/init.sls
+++ b/salt/admin/init.sls
@@ -1,0 +1,94 @@
+include:
+  - crypto
+
+# If we are in a docker container, `service.running` fails with error:
+#
+# "No service execution module loaded: check support for service management on Leap-42"
+#
+# Anyway, no need to check for the service on the docker container, since it is the PID 1,
+# so if it's not running, the container is gone.
+#
+# Additionally, virtual grains report `kvm` virtualization, so we just need to check if
+# /.dockerenv file exists to detect if we are running inside a docker container.
+#
+# Bug report: https://github.com/saltstack/salt/issues/22467
+#
+# Outside a container salt-minion service will listen for signing_policies file changes and
+# restart the service. Inside a container we mount this file, so when the salt minion starts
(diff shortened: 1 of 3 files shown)


mssola:
See kubic-project/caasp-container-manifests#57

mssola:
I've copy-pasted stuff from the `ca` directly. This is probably not the best approach ...

kiall:
I'm not sure I understand what this PR is supposed to be doing?

mssola:
> This is all specific to the ca node, the admin node doesn't require any of this.

> I think salt/admin/init.sls should just be blank, or not even exist.

> As long as we have an identifiable salt-minion (with id: admin, even with roles: admin too), probably created by the activate.sh script, and we can update our salt-states in the future to target it, we are fine.

> I don't think we want any of this for the admin node, since all this is only ca-related.

> The current contents of the files are unrelated to what the admin should do. We can add salt files for the admin node whenever we want, when we have something specific to do in it, as long as the salt-minion running in the administration dashboard node is connected to the salt-master (with the correct id/roles grain).

> I would drop this PR, at least for now, until the admin has specific logic to run.

Thanks for the review. I agree, I submitted this PR along with the one with kubic-project/caasp-container-manifests#57, because this was what I more or less used for kubic-project/velum#185. As discussed on IRC, this is not really the way to go, and instead we are moving towards having it on the host directly like the other salt minions.

Since, as you are saying, this brings nothing useful at the moment, we can happily drop this PR (I should've marked this PR as WIP...).

> I'm not sure I understand what this PR is supposed to be doing?

Hopefully kubic-project/terraform#72 is more clarifying 😅. As discussed for kubic-project/velum#185, we are having the admin node as a salt minion as well, since this way we'll be able to perform some operations like obtaining the `tx_*` grains, or rebooting the node in a transparent way. Since that PR, Velum already handles this minion id, and ignores it as it's done for the `ca`.

PR #112, opened 2017-06-13
Refresh grains at the start of orchestration

kiall:
Force a refresh of grains at the start of each orchestration run.

### DIFF
--- a/salt/orch/kubernetes.sls
+++ b/salt/orch/kubernetes.sls
@@ -20,12 +20,20 @@ update_pillar:
     - require:
       - salt: hostname_setup
 
+update_grains:
+  salt.function:
+    - tgt: '*'
+    - name: saltutil.refresh_grains
+    - require:
+      - salt: hostname_setup
+
 update_mine:
   salt.function:
     - tgt: '*'
     - name: mine.update
     - require:
       - salt: update_pillar
+      - salt: update_grains
 
 update_modules:
   salt.function:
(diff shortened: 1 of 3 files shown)


salt/orch/kubernetes.sls
```diff
@@ -19,12 +19,20 @@ update_pillar:
     - require:
       - salt: hostname_setup
 
+update_grains:
+  salt.function:
+    - tgt: '*'
+    - name: saltutil.refresh_grains
```
inercia: Do we already have a Salt version that supports `refresh_grains`? If that is the case, could you also update `orch/update-etc-hosts.sls` and add it?

kiall: We do, we're now on 2016.11.4 which supports it.

Is it needed for update-etc-hosts.sls? I'll add it, it's should be harmless + safer anyway

salt/orch/update-etc-hosts.sls
```diff
@@ -1,15 +1,20 @@
...
+  salt.function:
+    - tgt: '*'
+    - name: saltutil.refresh_pillar
+
 update_grains:
   salt.function:
-    - tgt: 'roles:kube-(master|minion)'
-    - tgt_type: grain_pcre
-    - name: saltutil.sync_all
+    - tgt: '*'
```
inercia: Please double check this `'*'` works fine, because I remember it dumped some errors on the `ca`.

salt/orch/kubernetes.sls
inercia: Thanks. This is what I had before @grahamhayes told me `refresh_grains` was not available in our Salt.

salt/orch/update-etc-hosts.sls
inercia: It seems it works fine.

PR #116, opened 2017-06-14
Ensure that reactor states only run on completed nodes

grahamhayes:
This ensures that we do not run reactor orchestrations on
nodes that have not completed bootstrapping.

This ensures that a node cannot have 2 states applied to
it at the same time.


kiall:
This fixes CI, still working out exactly why, but it does seem to.

PR #122, opened 2017-06-19
Do not fail if `salt.function` has no minions to target

ereslibre:
Currently, `update-etc-hosts` orchestration fails because `update_mine`
`salt.function` cannot target any minions at the beginning, and since
this is a prerequisite for other states, the Reactor orchestration
fails.

Only call to these `salt.function` if there are any minions to target.

### DIFF
--- a/salt/orch/update-etc-hosts.sls
+++ b/salt/orch/update-etc-hosts.sls
@@ -1,11 +1,14 @@
+{% if salt.saltutil.runner('mine.get', tgt='P@roles:kube-(master|minion) and G@bootstrap_complete:true', fun='caasp_fqdn', tgt_type='compound')|length > 0 %}
 update_pillar:
   salt.function:
-    - tgt: '*'
+    - tgt: 'P@roles:kube-(master|minion) and G@bootstrap_complete:true'
+    - tgt_type: compound
     - name: saltutil.refresh_pillar
 
 update_grains:
   salt.function:
-    - tgt: '*'
+    - tgt: 'P@roles:kube-(master|minion) and G@bootstrap_complete:true'
+    - tgt_type: compound
     - name: saltutil.refresh_grains
 
 update_mine:
@@ -36,3 +39,4 @@ kube_master_setup:
       - kubernetes-master
     - require:
       - salt: etc_hosts_setup
+{% endif %}


salt/orch/kubernetes.sls
```diff
@@ -127,6 +127,7 @@ reboot_setup:
     - require:
       - salt: kube_master_setup
 
+{% if salt['mine.get']('roles:kube-(master|minion)', 'caasp_fqdn', expr_form='grain_pcre').values()|length > 0 %}
```
grahamhayes: I am not sure this is needed - if there was no machines with these roles, the rest of the orch would have failed as well.

salt/orch/update-etc-hosts.sls
```diff
@@ -1,13 +1,18 @@
+{% if salt['mine.get']('roles:kube-(master|minion)', 'caasp_fqdn', expr_form='grain_pcre').values()|length > 0 %}
 update_pillar:
   salt.function:
-    - tgt: '*'
+    - tgt: 'roles:kube-(master|minion)'
```
grahamhayes: This should also target `G@bootstrap_complete:true` as the act of getting info for the pillar and grains can lock the node as well (just for a shorter period of time than running a state)

salt/orch/update-etc-hosts.sls
```diff
@@ -1,13 +1,18 @@
...
+    - tgt_type: grain_pcre
     - name: saltutil.refresh_pillar
 
 update_grains:
   salt.function:
-    - tgt: '*'
+    - tgt: 'roles:kube-(master|minion)'
```
grahamhayes: Same comment as the pillar refresh

salt/orch/kubernetes.sls
ereslibre: Wanted to play safe, but you are right, it wouldn't make sense as things would have failed far before we reach this point.

salt/orch/update-etc-hosts.sls
ereslibre: I am not sure with both of them... I know this can still raise the `function is running as PID ...` salt error... but if an ip address changes even before the orchestration is called we want to know... no?

ereslibre: Updated the PR with your comments, but I think we have the problem stated in the previous comment...

grahamhayes: It is a little more complex. 

We do not care if the IP has changed before the highstate has run. After the highstate has run, we care.

e.g for workers - if an IP changes before https://github.com/kubic-project/salt/blob/master/salt/orch/kubernetes.sls#L110 - we don't really care as the highstate should deal with it.

ereslibre: Since we are explicitly refreshing grains and updating the mine on the `kubernetes` orch I guess it's fine, yes. So disregard my comment.

ereslibre:
After having deployed a cluster, manually running `docker exec -it $(docker ps | grep salt-master | awk '{print $1}') salt-run state.orchestrate orch.update-etc-hosts` still returns:

```
freedom-desktop_master:

Summary for freedom-desktop_master
-----------
Succeeded: 0
Failed:   0
-----------
Total states run:    0
Total run time:  0.000 ms
```

It looks like the mine results is returning an empty array, while:

```
freedom-desktop suse/terraform@master docker exec -it $(docker ps | grep salt-master | awk '{print $1}') salt '*' grains.get bootstrap_complete          
1558c4aa94b14a5d9f93ed02b9f93c9f:
    True
ed18d07117cb482982edcddc514a2d70:
    True
ca:
33e1be588106421587d3c53693f7f651:
    True
```

Adding back the `WIP` flag. Will review this ASAP.

ereslibre:
As the [documentation states](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.mine.html#salt.modules.mine.get), the orchestration needs a special syntax, since it's run on the context of the `salt-master`.

ereslibre:
Now, when running: `docker exec -it $(docker ps | grep salt-master | awk '{print $1}') salt-run state.orchestrate orch.update-etc-hosts -l debug `

With an empty cluster (not bootstrapped yet):
```
freedom-desktop_master:

Summary for freedom-desktop_master
-----------
Succeeded: 0
Failed:   0
-----------
Total states run:    0
Total run time:  0.000 ms
```

With an existing cluster:
```
Summary for freedom-desktop_master
------------
Succeeded: 5 (changed=3)
Failed:    0
------------
Total states run:     5
Total run time:   3.492 s
```

ereslibre:
I can also confirm, running the e2e locally with: `VERBOSE=1 SALT_BRANCH=pull/122/head rspec spec/**/*`

```
Finished in 12 minutes 36 seconds (files took 0.48057 seconds to load)
1 example, 0 failures
```

ereslibre:
Tests passing, two +1, merging.

PR #125, opened 2017-06-19
Set etcd batch size to 3 nodes

kiall:
Currently, we never ask for more than 3 members. Setting this to 3 ensures
we don't let more than 3 members attempt etcd discovery before a cluster
has been fully formed. If we have less this 3, this will still succeed, as
the exact number of members we expect will also end up attempting discovery
at the same time.

### DIFF
--- a/salt/orch/kubernetes.sls
+++ b/salt/orch/kubernetes.sls
@@ -69,24 +69,20 @@ etcd_discovery_setup:
       - salt: ca_setup
       - salt: update_modules
 
-etcd_nodes_setup:
-  salt.state:
-    - tgt: 'roles:etcd'
-    - tgt_type: grain
-    - highstate: True
-    - batch: 1
-    - require:
-      - salt: etcd_discovery_setup
-
 etcd_proxy_setup:
   salt.state:
     - tgt: 'roles:kube-(master|minion)'
     - tgt_type: grain_pcre
     - sls:
       - etcd-proxy
-    - batch: 5
+    # Currently, we never ask for more than 3 members. Setting this to 3 ensures
+    # we don't let more than 3 members attempt etcd discovery before a cluster
+    # has been fully formed. If we have less this 3, this will still succeed, as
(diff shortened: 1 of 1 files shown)


salt/orch/kubernetes.sls
```diff
@@ -69,24 +69,20 @@ etcd_discovery_setup:
...
 etcd_proxy_setup:
   salt.state:
     - tgt: 'roles:kube-(master|minion)'
     - tgt_type: grain_pcre
     - sls:
       - etcd-proxy
-    - batch: 5
+    # Currently, we never ask for more than 3 members. Setting this to 3 ensures
+    # we don't let more than 3 members attempt etcd discovery before a cluster
+    # has been fully formed. If we have less this 3, this will still succeed, as
```
ereslibre: Typo I think `If we have less than 3`.

salt/orch/kubernetes.sls
```diff
@@ -69,24 +69,20 @@ etcd_discovery_setup:
...
+    # the exact number of members we expect will also end up attempting discovery
+    # at the same time.
+    - batch: 3
```
inercia: I think it would be good to move some of these magic numbers into pillar vars...

It should also be noted that this value would make impossible to have a etcd cluster with size={5,7}

kiall: Re pillar: Do we have access to pillar values here? I didn't think we did?

Re size={5,7}: Yea agree, However - I'm thinking that's better than the currently situation, given we don't actually expose the ability to choose the etcd cluster size currently.

ereslibre: > Re pillar: Do we have access to pillar values here? I didn't think we did?

I think following the same strategy as: https://github.com/kubic-project/salt/pull/122/files#diff-138362f38f0e3092ac3f7a99c15693d1R1 would give us access to the pillar (`pillar.get`).

PR #127, opened 2017-06-20
[WIP]  ETCD - Client interface exposed to the network with no authentication

PI-Victor:
Fixes Bug 1044590
ETCD Should listen for client connections only on loopback interface
I'm not sure if i'm missing anything else with this.

### DIFF
--- a/salt/etcd-proxy/etcd-proxy.conf.jinja
+++ b/salt/etcd-proxy/etcd-proxy.conf.jinja
@@ -4,7 +4,7 @@
 ETCD_DATA_DIR="/var/lib/etcd/"
 ETCD_NAME="{{ grains['id'] }}"
 
-ETCD_LISTEN_CLIENT_URLS="http://0.0.0.0:2379"
+ETCD_LISTEN_CLIENT_URLS="http://127.0.0.1:2379"
 ETCD_LISTEN_PEER_URLS="https://0.0.0.0:2380"
 ETCD_ADVERTISE_CLIENT_URLS="http://{{ grains['caasp_fqdn'] }}:2379"
 
--- a/salt/etcd/etcd.conf.jinja
+++ b/salt/etcd/etcd.conf.jinja
@@ -6,7 +6,7 @@
 ETCD_DATA_DIR="/var/lib/etcd/"
 ETCD_NAME="{{ grains['id'] }}"
 
-ETCD_LISTEN_CLIENT_URLS="https://0.0.0.0:2379"
+ETCD_LISTEN_CLIENT_URLS="https://127.0.0.1:2379"
 ETCD_LISTEN_PEER_URLS="https://0.0.0.0:2380"
 ETCD_ADVERTISE_CLIENT_URLS="https://{{ grains['caasp_fqdn'] }}:2379"
 


salt/etcd-proxy/etcd-proxy.conf.jinja
```diff
@@ -4,7 +4,7 @@
 ETCD_DATA_DIR="/var/lib/etcd/"
 ETCD_NAME="{{ grains['id'] }}"
 
-ETCD_LISTEN_CLIENT_URLS="http://0.0.0.0:2379"
+ETCD_LISTEN_CLIENT_URLS="http://127.0.0.1:2379"
```
mjura: If you are changing this option please also change `ETCD_ADVERTISE_CLIENT_URLS` because etcd cluster will advertise that it is still listening on all interfaces which is not true.

salt/etcd/etcd.conf.jinja
```diff
@@ -6,7 +6,7 @@
 ETCD_DATA_DIR="/var/lib/etcd/"
 ETCD_NAME="{{ grains['id'] }}"
 
-ETCD_LISTEN_CLIENT_URLS="https://0.0.0.0:2379"
+ETCD_LISTEN_CLIENT_URLS="https://127.0.0.1:2379"
```
mjura: If you are changing this option please also change `ETCD_ADVERTISE_CLIENT_URLS` because etcd cluster will advertise that it is still listening on all interfaces which is not true.

salt/etcd/etcd.conf.jinja
```diff
@@ -6,7 +6,7 @@
...
 ETCD_LISTEN_PEER_URLS="https://0.0.0.0:2380"
 ETCD_ADVERTISE_CLIENT_URLS="https://{{ grains['caasp_fqdn'] }}:2379"
```
mjura: This has to be changed too

salt/etcd-proxy/etcd-proxy.conf.jinja
```diff
@@ -4,7 +4,7 @@
...
 ETCD_LISTEN_PEER_URLS="https://0.0.0.0:2380"
 ETCD_ADVERTISE_CLIENT_URLS="http://{{ grains['caasp_fqdn'] }}:2379"
```
mjura: This has to be changed too

inercia: I think it doesn't: we must advertise the public IP.

salt/etcd/etcd.conf.jinja
inercia: Ditto

salt/etcd-proxy/etcd-proxy.conf.jinja
inercia: In fact, we don't advertise IPs: we advertise names that will be resolved with the help of `/etc/hosts` in all the machines...

mjura: @inercia : I would agree with you, but I think that `grains['caasp_fqdn']` does not point on `127.0.0.1`. So with this setup we will get situation when etcd cluster is advertising wrong client URL

inercia: I don't understand you @mjura. I don't see anything wrong with advertising `https://{{ grains['caasp_fqdn'] }}:2379` as the externally accessible address...

mjura: Ok, well I still have objections, but I understand what do you mean.

mjura:
I haven't tested this change but LGTM

ereslibre:
Testing this one right now.

salt/etcd/etcd.conf.jinja
grahamhayes: Changing this line means that the proxies can no longer talk to the cluster.

grahamhayes:
after a little bit of looking, it seems we never use `salt/etcd/etcd.conf.jinja` at all.

This means that all of our connections are over unencrypted HTTP.

kiall:
On top of what graham pointed out, since the proxy mode etcd's are using the member mode etcd's client rather than peering port for traffic, that means we can't set this to 127.0.0.1. And, likely, instead need to do full mutual TLS for the client port

grahamhayes:
#102 should fix this - if we can get it merged

mjura:
Currently I've rebased https://github.com/kubic-project/salt/pull/102 and I'm testing it

PI-Victor:
the PR wasn't about unencrypted data transport, it's about being able to connect to the etcd client interface without authentication. The suggested solution by David was to have the etcd server listen to the loopback interface for client connections.

i wasn't very clear on the purpose of the PR, however the details are in bugzilla.
will change the title to reflect that and try to find some way to use auth or something else.

PI-Victor:
p.s. updated the title to better reflect the purpose, so it's a WIP for now.
back to the drawing board, i guess.

barclaac:
Can we close this. We now know that listening on localhost just doesn't work.

PR #128, opened 2017-06-21
Do not run etcd discovery on every orchestration run, only the first time

ereslibre:
When adding new nodes, the `orch.kubernetes` orchestration was failing because
etcd is refusing to start since the etcd discovery mechanism was already used
when bootstrapping the cluster.

With this change we ensure that we use the discovery mechanism only when we
are boostrapping the cluster (no machines have the `bootstrap_complete` grain
set to true).

Required for the add nodes feature.

### DIFF
--- a/salt/etcd-discovery/init.sls
+++ b/salt/etcd-discovery/init.sls
@@ -9,7 +9,7 @@ etcd-discovery-setup:
     - name: curl
   cmd.run:
     - name: curl -L -X PUT {{ etcd_size_uri }} -d value={{ salt.k8s_etcd.get_cluster_size() }}
-    - unless: curl {{ etcd_size_uri }} | grep '"value":"{{ salt.k8s_etcd.get_cluster_size() }}"'
+    - onlyif: curl {{ etcd_size_uri }} | grep '"message":"Key not found"'
     - require:
       - pkg: curl
 


ereslibre:
I can confirm I could create a cluster from scratch (with 1 master and 1 worker). Then in separate steps I added two more workers (1 worker each time).

Output after adding one worker:
```
freedom-desktop ~ KUBECONFIG=kubeconfig kubectl get nodes
NAME                                                 STATUS    AGE       VERSION
2351b3dac3cd4a55b410ce855cea9fb9.infra.caasp.local   Ready     2m        v1.5.3
4ee5cfdb23ae431fb426d81f0ee54d55.infra.caasp.local   Ready     1m        v1.5.3
```

Output after adding another worker:
```
NAME                                                 STATUS    AGE       VERSION
2351b3dac3cd4a55b410ce855cea9fb9.infra.caasp.local   Ready     2m        v1.5.3
4ee5cfdb23ae431fb426d81f0ee54d55.infra.caasp.local   Ready     2m        v1.5.3
610ea29a16aa4e5a870e3eb8fa7498b8.infra.caasp.local   Ready     12s       v1.5.3
```

I used https://github.com/kubic-project/velum/pull/201 to add the workers through the UI.

salt/orch/kubernetes.sls
```diff
@@ -82,7 +84,12 @@ etcd_proxy_setup:
     # at the same time.
     - batch: 3
     - require:
+{% if salt.saltutil.runner('mine.get', tgt='P@roles:kube-(master|minion) and G@bootstrap_complete:true', fun='caasp_fqdn', tgt_type='compound')|length == 0 %}
```
inercia: As a minor comment, for the sake of simplicity, maybe you to move these `if` to the top of the file and set a variable like:

```jinja
{%- set bootstrap_complete = (salt.saltutil.runner('mine.get', tgt='P@roles:kube-(master|minion) and G@bootstrap_complete:true', fun='caasp_fqdn', tgt_type='compound')|length == 0) %}
```

(I don't know if Jinja supports this kind of expression evaluations)

and then you could just

```jinja
{% if bootstrap_complete %}
...
```

ereslibre:
@inercia Addressed your comment. Thanks, please review (force pushing dismisses old reviews)

inercia:
Sorry to bother you again @ereslibre but I think the whole change can be simplified to just surrounding the contents of `etcd-discovery/init.sls` with a `{% if not grains["bootstrap_complete"] %}`.

ereslibre:
> Sorry to bother you again @ereslibre but I think the whole change can be simplified to just surrounding the contents of etcd-discovery/init.sls with a {% if not grains["bootstrap_complete"] %}.

I'm not that sure about that, here the condition is more explicit. We want to only run this when the bootstrapping isn't done yet, and at orchestration level I think this condition makes more sense than inside the sls itself (and is more clear). Inside the sls file it's more hidden, when I think this logic belongs to the orchestration.

inercia:
> I'm not that sure about that, here the condition is more explicit. We want to only run this when the bootstrapping isn't done yet, and at orchestration level I think this condition makes more sense than inside the sls itself (and is more clear). Inside the sls file it's more hidden, when I think this logic belongs to the orchestration.

Yeah, but, besides the simplicity of doing it at the `sls` level, I think this is less error prone. What if `etcd-discovery` is `included`/`required` by some other state and it is run anyway?

In fact, why is this needed? We shouldn't need to use this `grain` at all, as there is a `unless` clause that checks if the etcd discovery has already been setup, right? So this should never be run twice, right?

ereslibre:
> In fact, why is this needed? We shouldn't need to use this grain at all, as there is a unless clause that checks if the etcd discovery has already been setup, right? So this should never be run twice, right?

It is, indeed. Because when you grow the cluster (add new nodes after it has been bootstrapped), the orchestration is run from Velum, and the unless here https://github.com/kubic-project/salt/blob/master/salt/etcd-discovery/init.sls#L12 cannot save us (the number of nodes did really change), so the discovery will be tried to be executed -- but, we shouldn't, since the discovery is meant only for the first boostrapping of etcd.

inercia:
> It is, indeed. Because when you grow the cluster (add new nodes after it has been bootstrapped), the orchestration is run from Velum, and the unless here https://github.com/kubic-project/salt/blob/master/salt/etcd-discovery/init.sls#L12 cannot save us (the number of nodes did really change), so the discovery will be tried to be executed -- but, we shouldn't, since the discovery is meant only for the first boostrapping of etcd.

So I would just change the condition for the `unless` and skip setting the etcd cluster size if there is any previous value.

ereslibre:
> So I would just change the condition for the unless and skip setting the etcd cluster size if there is any previous value.

I like this, a lot. On it, thank you.

ereslibre:
I can confirm I could bootstrap a cluster with 1 master and 1 worker, and added 2 workers afterwards (one by one, adding two times)

salt/etcd-discovery/init.sls
```diff
@@ -9,7 +9,7 @@ etcd-discovery-setup:
     - name: curl
   cmd.run:
     - name: curl -L -X PUT {{ etcd_size_uri }} -d value={{ salt.k8s_etcd.get_cluster_size() }}
-    - unless: curl {{ etcd_size_uri }} | grep '"value":"{{ salt.k8s_etcd.get_cluster_size() }}"'
+    - onlyif: curl {{ etcd_size_uri }} | grep '"message":"Key not found"'
```
ereslibre: We could also use `"errorCode":100` here... but it looks a bit harder to remember on our `sls` file, but safer...

PR #119, opened 2017-06-16
WIP: Additional certificates for repositories for Docker (bsc#1039877)

inercia:
Follow the instruction found [here](https://docs.docker.com/registry/insecure/#using-self-signed-certificates) for adding customers' certificates to Docker so they can pull from other registries.

## How to test this

* download [the caasp-registry.sh script](https://github.com/inercia/caasp-tf/blob/master/provision/admin/caasp-registry.sh) and copy it to `/root` in the Admin Node
* `ssh admin-node` and `./caasp-registry.sh start`. That will generate the certificates in the `/root/certificates.local` and start the registry.
* Import the `alpine` image to the registry with `./caasp-registry.sh import alpine`
* edit the `/usr/share/salt/kubernetes/pillar/certificates.sls` in the Admin Node and add the `/root/certificates.local/ca.pem` certificate as:
    ```
    repositories:
        - dashboard:5000: |
             ---- BEGIN CERTIFICATE -----
             ...
    ```
* orchestrate
* get the Admin Node's `IP` and `ssh node1` and `echo "IP dashboard" > /etc/hosts`
* try to pull the `alpine` image from the registry with `docker pull dashboard:5000/alpine`. After that, you should see the `dashboard:5000/alpine` image when doing `docker images`.

## TODO:
- [x] more testing
- [x] document how to test it
- [ ] update the pillar format

PR #129, opened 2017-06-21
enable multiple docker registries (bsc#1039876)

MaximilianMeister:
https://bugzilla.suse.com/show_bug.cgi?id=1039876

## Depends on
https://github.com/kubic-project/salt/pull/313

### DIFF
--- a/salt/docker/init.sls
+++ b/salt/docker/init.sls
@@ -30,11 +30,14 @@ include:
 
 {% set docker_args = salt['pillar.get']('docker:args', '') %}
 {% set docker_logs = salt['pillar.get']('docker:log_level', '') %}
-{% set docker_reg  = salt['pillar.get']('docker:registry', '') %}
+{% set registries  = salt['pillar.get']('docker:registries', []) %}
 {% set docker_opts = docker_args + " --log-level=" + docker_logs %}
-{% if docker_reg|length > 0 %}
-  {% set docker_opts = docker_opts + " --insecure-registry=" + docker_reg + " --registry-mirror=http://" + docker_reg  %}
-{% endif %}
+{% for registry in registries.values() %}
+  {% set docker_opts = docker_opts + " --insecure-registry=" + registry.get("name") %}
+  {% if registry.get("mirror") %}
+    {% set docker_opts = docker_opts + " --registry-mirror=http://" + registry.get("name") %}
+  {% endif %}
+{% endfor %}
 
 docker:
   pkg.installed:


MaximilianMeister:
[Bug 1039876 - docker cannot pull images from insecure registry](https://trello.com/c/PyAkd0il/57-bug-1039876-docker-cannot-pull-images-from-insecure-registry)

salt/docker/init.sls
```diff
@@ -23,10 +23,10 @@ include:
 #######################
 
 {% set docker_opts = salt['pillar.get']('docker:args', '')%}
-{% set docker_reg  = salt['pillar.get']('docker:registry', '') %}
-{% if docker_reg|length > 0 %}
-  {% set docker_opts = docker_opts + " --insecure-registry={{ docker_reg }} --registry-mirror=http://{{ docker_reg }}" %}
-{% endif %}
+{% set docker_regs  = salt['pillar.get']('docker:registry', '').split(',') %}
+{% for reg in docker_regs if docker_regs|length > 0 %}
+  {% set docker_opts = docker_opts + " --insecure-registry={{ reg }} --registry-mirror=http://{{ reg }}" %}
```
grahamhayes: why are we adding these as mirrors as well as insecure?

MaximilianMeister: tried to look back in [history](https://github.com/kubic-project/salt/commit/4d8f25bdafe0e6ba044548fa44b2098f60c1bb5b) but the commit message wasn't descriptive enough. maybe @inercia can provide some details.

would it make sense to separate `--insecure-registry` and `--registry-mirror` in 2 pillars? seems there could be a use case where you may want one **or** the other

inercia: Registries need the `--insecure-registry` if there is no trusted certificate for them.

robdaemon:
@MaximilianMeister - this PR is several months old - is it still needed? It's time to clear out some cruft.

MaximilianMeister:
> this PR is several months old - is it still needed? It's time to clear out some cruft.

@robdaemon it's on hold until our pillars can accept more complex data structures, something we plan for 2.0 so we will come back to it soon

inercia:
Depends on kubic-project/salt#238, kubic-project/velum#315

robdaemon:
Why does this depend on the Pillar API changes? These are two entirely separate concerns and they should not be linked.

ereslibre:
> Why does this depend on the Pillar API changes? These are two entirely separate concerns and they should not be linked.

> Looking at the code change, it doesn't rely on complex Pillar data structures. It implies data structure by splitting at a ,.

This code change is not complete. Our current pillar override implementation has serious shortcomings. Let me elaborate:

MySQL Pillar key: `some:value:override`, value: `whatever`

This will be a scalar, with key `some:value:override` and value `whatever`. The way arrays are represented is with collisions, so:

MySQL Pillar key: `some:values:override`, value: `whatever1`
MySQL Pillar key: `some:values:override`, value: `whatever2`

This will be read by salt as `some;values:override` key and value [`whatever1`, `whatever2`].

Let's keep aside the fact of salt accessing mysql, what is not the best design. We have a very serious shortcoming: you cannot represent a unitary array (it's either a scalar or an array > 1 item).

This also affects us with complex structures overriding other complex structures from the original pillar. Things were more simple before, and we need a final solution to stop hacking things around with the pillar.

inercia:
> Why does this depend on the Pillar API changes? These are two entirely separate concerns and they should not be linked.
> Looking at the code change, it doesn't rely on complex Pillar data structures. It implies data structure by splitting at a ,.

We could solve the only this issue with commas, but we would like to use the same solution for this_and_ for [this PR](https://github.com/kubic-project/salt/pull/119).

flavio:
Let's have a properly defined representation of the registry related data into the database (hence have something like [that](https://github.com/kubic-project/velum/pull/315) available). I don't want to introduce any more hacks to get this bug closed.

We will also need to have a discussion about what kind of scenarios we are going to support (mirroring is definitely going to be handled).

inercia:
Unblocked by #31, but this PR must be adapted to the new PR structure.

MaximilianMeister:
part of https://github.com/kubic-project/salt/pull/313 now

PR #130, opened 2017-06-22
[WIP] Do not run kubectl uncordon if it is not really needed

inercia:


salt/kubernetes-minion/init.sls
```diff
@@ -73,6 +73,8 @@ kubelet:
             sleep 1 && ELAPSED=$(( $ELAPSED + 1 ))
         done
         echo changed="$(echo $output | grep 'already uncordoned' &> /dev/null && echo no || echo yes)"
+    - unless: |
+        kubectl uncordon {{ grains['caasp_fqdn'] }} | grep 'already uncordoned'
```
inercia: Yeah, I think I committed & PRed too fast...

PR #131, opened 2017-06-22
Turn off `auto_accept`

grahamhayes:
This works in conjunction with 

https://github.com/kubic-project/caasp-container-manifests/pull/62 and 
https://github.com/kubic-project/velum/pull/206

to allow explicit acceptance of nodes into salt.

### DIFF
--- a/config/master.d/master.conf
+++ b/config/master.d/master.conf
@@ -1,5 +1,5 @@
 user: root
-auto_accept: True
+auto_accept: False
 interface: 0.0.0.0
 event_return: mysql
 presence_events: True


config/master.d/master.conf
```diff
@@ -1,5 +1,4 @@
 user: root
-auto_accept: True
```
ereslibre: Would it make sense to explicitly set it to `False`?, just in case for whatever reason this default changes in the future

grahamhayes: Yes, it would. Adding now.

grahamhayes:
e2e tests updated - https://github.com/kubic-project/e2e-tests/pull/63

kiall:
Without a rebase, this won't pass e2e-tests and the tools won't let you test with a locally rebased version. So, I've pushed a rebase.

PR #132, opened 2017-06-23
Set MySQL as the job cache for the Salt master

mssola:
First of all, we can specify an external job cache. If we don't do that,
then the `keep_jobs` option only applies to the local cache. This means that
Salt will not clean up jobs, events and returns older than the specified
`keep_jobs` value (default: 24h) for the MySQL returner that we have already
configured.

Moreover, since we'd already be using MySQL as a job cache, we don't
have to use the local system (/var/cache/salt/master/jobs/) as a cache
(note that Salt would still be using this directory to avoid JID
collisions). The documentation also says that the local cache can be a
burden for large deployments.

See bsc#1044133

### DIFF
--- a/config/master.d/master.conf
+++ b/config/master.d/master.conf
@@ -5,6 +5,11 @@ event_return: mysql
 presence_events: True
 worker_threads: 10
 timeout: 20
+
+master_job_cache: mysql
+job_cache: False
+keep_jobs: 24
+
 file_roots:
   base:
     - /usr/share/salt/kubernetes/salt


mssola:
To test this, I recommend setting `keep_jobs` to 1 as well (unless you want to keep your test running for 24 hours 😅).

ereslibre:
Is this also cleaning the `salt_events` table?

mssola:
> Is this also cleaning the salt_events table?

From the [source code (`_purge_jobs` function)](https://github.com/saltstack/salt/blob/develop/salt/returners/mysql.py#L466), it's cleaning: `jids`, `salt_events` and `salt_returns`.

ereslibre:
I have one question: it seems from the source code (https://github.com/saltstack/salt/blob/d1f80bab3993059060ca51263e9d048e3a618051/salt/returners/mysql.py#L560-L581) that every `loop_interval` from the master will be called to `clean_old_jobs`, that in turn will look at `keep_jobs`, so in theory:

- `keep_jobs`: 24
- `archive_jobs`: False

Should call to `_purge_jobs` every `loop_interval`, which will perform a cleanup: https://github.com/saltstack/salt/blob/d1f80bab3993059060ca51263e9d048e3a618051/salt/returners/mysql.py#L472-L498.

mssola:
> I don't understand why we need to set the master_job_cache to mysql.

If you look at the [code of the master](https://github.com/saltstack/salt/blob/d1f80bab3993059060ca51263e9d048e3a618051/salt/master.py#L235), you'll see where the `clean_old_jobs` is called from the master. As you can see, it calls instead: https://github.com/saltstack/salt/blob/d1f80bab3993059060ca51263e9d048e3a618051/salt/daemons/masterapi.py#L189. It simply calls the `clean_old_jobs` of the current `master_job_cache`. If we do not assign this to `mysql`, then the local cache is assumed, thus never cleaning the MySQL DB and accumulating old events as the bug complains.

I'd agree that this could be better documented...

mjura:
I see some big disadvantage of this solution, for medium and big installation all salt-minions will be connecting to MySQL db and report job status. It will be bottleneck for our solution, I would propose to use Redis as complimentary external job cache, which is memory database.

jordimassaguerpla:
As for redis, we don't have an RPM for this. This could be discussed for future releases but is not going to happen for 1.0.

kiall:
@mjura are you sure every minion connects? this is in the master config, and the config var even mentions "master" in it?

Re redis: agree with Jordi, deciding to add another DB at this point is too risky, and - even if it was available - I doubt it would be necessary.

mssola:
> @mjura are you sure every minion connects? this is in the master config, and the config var even mentions "master" in it?

That's right. This only makes the caching from the master side into the MySQL DB. Also, note that we are already storing all the events from minions anyways into the DB (which is fine), so this "bottleneck" was there.

In regards to Redis, we could give it a thought after the 1.0 release, but not before. So I agree with both @jordimassaguerpla and @kiall.

PI-Victor:
[Bug 1044133 - Velum's DB keeps growing](https://trello.com/c/CrsWLg4K/136-bug-1044133-velums-db-keeps-growing)

mjura:
Rebase is needed

inercia:
> I see some big disadvantage of this solution, for medium and big installation all salt-minions will be connecting to MySQL db and report job status. It will be bottleneck for our solution, I would propose to use Redis as complimentary external job cache, which is memory database.

I don't think we will reach that point at any time soon. Anyway, if we reach that point where MySQL could seem to be a bottleneck, we could evaluate different alternatives. Personally, I would prefer not to add another database to our setup...

flavio:
A couple of questions:

  * are we fine keeping jobs for 24 hours or do we want to increase the default value to something more (1 week?)
  * wouldn't it be better to be explicit about how much time everything is kept. I would be in favor of having `keep_jobs: 24` written down into our configuration to avoid to search for the default value use by salt.

mssola:
> are we fine keeping jobs for 24 hours or do we want to increase the default value to something more (1 week?)

and

> If anything - we should be reducing the time we keep them - the amount of events generated by a large cluster is huge, and causes MySQL to be a bottlekneck

In my opinion, Salt's default is quite good. It might be too low for small clusters, but it's actually to big for large deployments. I guess that we should tune this after getting feedback from customers (e.g. fetching what's the average cluster size).

> wouldn't it be better to be explicit about how much time everything is kept. I would be in favor of having keep_jobs: 24 written down into our configuration to avoid to search for the default value use by salt.

👍 

I'll rebase this PR now, so it can be properly reviewed.

mssola:
Rebased. Please, take a look again 😇 /ping @ereslibre @flavio @kiall @inercia @grahamhayes

mssola:
e2e tests are failing because of an older build (and we cannot trigger a re-build...). So, once I get more approvals I'll push -f it again to force a rebuild, and if it passes I can merge it myself.

mjura:
@mssola : Please force push this change again to retrigger CI tests

mssola:
Forced push.

ereslibre:
@flavio 
> are we fine keeping jobs for 24 hours or do we want to increase the default value to something more (1 week?)
I think a week is a more reasonable setting TBH. We need to make sure that we collect all the information of a problem on a customer within this range, so I'd feel more comfortable with a week at least.

Since @mssola's explanation in https://github.com/kubic-project/salt/pull/132#issuecomment-310974323, I think despite the window this change is good to go.

flavio:
The e2e tests are failing because of unrelated reasons, that's why I merged it.

PR #133, opened 2017-06-23
Ensure CA fields are static (bsc#1045766)

kiall:
As the DHCP domain name can change, we should avoid using
it in our CA cert in order to prevent it being unnecessarily
regenerated.

Fixes bsc#1045766


kiall:
Thanks guys!

PR #134, opened 2017-06-25
Add support for initializing certificates (user certificates)

ereslibre:
We add a reactor that will run an orchestration that will be executed
during the first boot of the services.

This new orchestration will make sure the CA generates both a private
and a public CA certificates, and depending on the service one or the
other will be used. E.g. salt-api uses the private CA, while velum will
use the public CA.

In the future, this salt-states can be modified, so the public CA is
a CA provided by the customer.

Fixes: bsc#1043589

Depends on:

- https://github.com/kubic-project/velum/pull/210
- https://github.com/kubic-project/caasp-devenv/pull/19
- https://github.com/kubic-project/caasp-container-manifests/pull/64
- https://github.com/kubic-project/e2e-tests/pull/65


salt/user-certs/init.sls
```diff
@@ -0,0 +1,44 @@
...
+    - user: root
+    - group: root
+    - mode: 644
+
+/etc/pki/{{ cert }}.crt:
+  x509.certificate_managed:
+    - ca_server: {{ salt['mine.get']('roles:ca', 'x509.get_pem_entries', expr_form='grain').keys()[0] }}
+    - signing_policy: minion
+    - public_key: /etc/pki/{{ cert }}.key
+    - CN: {{ grains['fqdn'] }}
```
flavio: Isn't the risk of picking a bad value here? However I don't think there's much we can do...

salt/user-certs/init.sls
```diff
@@ -0,0 +1,44 @@
...
+    - C: {{ pillar['certificate_information']['subject_properties']['C']|yaml_dquote }}
+    - Email: {{ pillar['certificate_information']['subject_properties']['Email']|yaml_dquote }}
+    - GN: {{ pillar['certificate_information']['subject_properties']['GN']|yaml_dquote }}
+    - L: {{ pillar['certificate_information']['subject_properties']['L']|yaml_dquote }}
+    - O: {{ pillar['certificate_information']['subject_properties']['O']|yaml_dquote }}
+    - OU: {{ pillar['certificate_information']['subject_properties']['OU']|yaml_dquote }}
+    - SN: {{ pillar['certificate_information']['subject_properties']['SN']|yaml_dquote }}
+    - ST: {{ pillar['certificate_information']['subject_properties']['ST']|yaml_dquote }}
```
flavio: About all the pillars used above (the `certificate_information`). I guess we will use the default value from [here](https://github.com/kubic-project/salt/blob/master/pillar/certificates.sls). Some of these values are empty, is this going to be a problem?

salt/user-certs/init.sls
ereslibre: Yeah... we have the same problem with any other solution AFAIK. Another possibility is to also add as `DNS names` well known names that we document, so the customer can point internally to one of those names (but also keeping the `fqdn` that salt detected).

salt/user-certs/init.sls
ereslibre: The current empty defaults shouldn't be a problem.

salt/infra-certs/init.sls
```diff
@@ -0,0 +1,58 @@
...
+    - keyUsage: nonRepudiation, digitalSignature, keyEncipherment
+    {% if ip_addresses|length > 0 %}
+    - subjectAltName: "{{ ", ".join(ip_addresses) }}"
+    {% endif %}
+    - days_valid: {{ pillar['certificate_information']['days_valid']['certificate'] }}
+    - days_remaining: {{ pillar['certificate_information']['days_remaining']['certificate'] }}
+    - backup: True
+    - require:
+      - sls:  crypto
+      - file: /etc/pki
```
inercia: Maybe not the right time (taking into account or tight schedule) but I was wondering if it could de possible to refactor these `x509.certificate_managed` into some common place (maybe with a `jinja2` `import`, a Salt `extend`...)

ereslibre: Yeah, I think we could consider a macro in this case... but I prefer to play safe at the moment

ereslibre:
Closing this one for now, will revisit it later.

PR #137, opened 2017-06-28
bsc#1045381 Ensure updates do not conflict with etc-hosts

grahamhayes:
This ensure that the etc-hosts orchestration does not
run during an upgrade, as this can cause conflicts on
the nodes, which cause salt to fail to complete an
`orch.update` run.

### DIFF
--- a/salt/orch/update-etc-hosts.sls
+++ b/salt/orch/update-etc-hosts.sls
@@ -1,19 +1,19 @@
-{% if salt.saltutil.runner('mine.get', tgt='P@roles:kube-(master|minion) and G@bootstrap_complete:true', fun='caasp_fqdn', tgt_type='compound')|length > 0 %}
+{% if salt.saltutil.runner('mine.get', tgt='P@roles:kube-(master|minion) and G@bootstrap_complete:true and not G@update_in_progress:true', fun='caasp_fqdn', tgt_type='compound')|length > 0 %}
 update_pillar:
   salt.function:
-    - tgt: 'P@roles:kube-(master|minion) and G@bootstrap_complete:true'
+    - tgt: 'P@roles:kube-(master|minion) and G@bootstrap_complete:true and not G@update_in_progress:true'
     - tgt_type: compound
     - name: saltutil.refresh_pillar
 
 update_grains:
   salt.function:
-    - tgt: 'P@roles:kube-(master|minion) and G@bootstrap_complete:true'
+    - tgt: 'P@roles:kube-(master|minion) and G@bootstrap_complete:true and not G@update_in_progress:true'
     - tgt_type: compound
     - name: saltutil.refresh_grains
 
 update_mine:
   salt.function:
-    - tgt: 'P@roles:kube-(master|minion) and G@bootstrap_complete:true'
+    - tgt: 'P@roles:kube-(master|minion) and G@bootstrap_complete:true and not G@update_in_progress:true'
     - tgt_type: compound
     - name: mine.update
     - require:
@@ -22,7 +22,7 @@ update_mine:
 
 etc_hosts_setup:
   salt.state:
-    - tgt: 'P@roles:kube-(master|minion) and G@bootstrap_complete:true'
+    - tgt: 'P@roles:kube-(master|minion) and G@bootstrap_complete:true and not G@update_in_progress:true'
     - tgt_type: compound
     - queue: True
     - sls:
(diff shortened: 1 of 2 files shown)


grahamhayes:
This was tested by doing the following:

- Apply the `tx_update_reboot_needed: true` grain by adding it to `/etc/salt/grains` on workers
- Run `docker exec -it $(docker ps | grep salt-master | awk '{print $1}') salt '*' saltutil.refresh_grains` to make sure the dashboard sees the update (you could wait for the 5 minute timer to trigger as well)
- Click "Update all Nodes" on Velum
- SSH to master, and run `service salt-minion restart` repeatedly

As the reactor fires on `minion/*/start` this should cause the orchestration to run while an update is in progress

salt/orch/update.sls
```diff
@@ -75,11 +83,29 @@ update_modules:
...
+      - update_in_progress
+      - false
+
 {% endfor %}
 
 {%- set workers = salt.saltutil.runner('mine.get', tgt='G@roles:kube-minion and G@tx_update_reboot_needed:true', fun='network.ip_addrs', tgt_type='compound') %}
 {%- for worker_id, ip in workers.items() %}
 
+# Ensure the node is marked as upgrading
+{{ worker_id }}-set_update_grain:
```
inercia: Many `_` and `-` mixed in names

grahamhayes:
@inercia `/etc/hosts` will be updated by the highstate after the reboot.

salt/orch/update.sls
grahamhayes: :man_facepalming:  - will fix now

inercia:
> @inercia /etc/hosts will be updated by the highstate after the reboot.

Not sure that is enough. Machine `X` has IP `10.10.0.9`, and it is updated:
* it downloads packages, `transactional-update update reboot`
* reboots
* gets new IP `10.10.0.39`
* applies highstate, so that updates the `/etc/hosts` at `X` with a line like `10.10.0.39 X.infra.local X`

But all the other machines in the cluster still have an entry like `10.10.0.9 X.infra.local X` in their `/etc/hosts`. Am I wrong?

grahamhayes:
Yes - the other hosts will have have the etc-hosts orchestration run on them - they would not have had the `update_in_progress` grain applied, so they will have the new hosts entry

salt/orch/update.sls
```diff
@@ -26,9 +26,17 @@ update_modules:
...
-
 {%- for master_id in masters.keys() %}
 
+# Ensure the node is marked as upgrading
+{{ master_id }}-set-update-grain:
+  salt.function:
+    - tgt: {{ master_id }}
+    - name: grains.setval
+    - arg:
+      - update_in_progress
```
mssola: Nitpicking: Following the same `-` vs `_` policy, maybe this should also be in `-` instead of `_` ?

inercia:
Ok, so the `update_in_progress` grain only blocks the `update-etc-hosts` orchestration in the machine that has it...

salt/orch/update-etc-hosts.sls
```diff
@@ -1,19 +1,19 @@
-{% if salt.saltutil.runner('mine.get', tgt='P@roles:kube-(master|minion) and G@bootstrap_complete:true', fun='caasp_fqdn', tgt_type='compound')|length > 0 %}
+{% if salt.saltutil.runner('mine.get', tgt='P@roles:kube-(master|minion) and G@bootstrap_complete:true and not G@update_in_progress:true', fun='caasp_fqdn', tgt_type='compound')|length > 0 %}
```
inercia: A suggestion:

```jinja
{%- set updates_all_target = "P@roles:kube-(master|minion) and G@bootstrap_complete:true and not G@update_in_progress:true" %}
```

And use it everywhere like `- tgt: {{ updates_all_target }}`

kiall: Makes sense, though, given this is a P1, we should do this after we've got the P1 closed out!

salt/orch/update.sls
kiall: Looking at our grains - `bootstrap_complete`, `update_in_progress`, `tx_update_reboot_needed`, `tx_update_failed` - we are consistently using underscores?

mssola: Makes sense 😄

PI-Victor:
[Bug 1045381 - The worker upgrade orchestration is interrupted by the update-etc-hosts orchestration](https://trello.com/c/oBJlBLPx/142-bug-1045381-the-worker-upgrade-orchestration-is-interrupted-by-the-update-etc-hosts-orchestration)

PR #138, opened 2017-06-28
Apply select states to Admin node

kiall:
Render the /etc/hosts file on the admin node, so nodes are reacable via
their internal FQDNs everywhere. Additionally, include the admin node
in the /etc/hosts files. (bsc#1045186)

Once the system bootstraps, we now disable rebootmgr on the
admin node. This allows the velum initiated updates to takeover
and prevent any unexpected surprises. (bsc#1046602)

Adds the proxy setup to the admin node (bsc#1043538)


kiall:
Since this one is difficult, at best, to test as it's all admin node - chatted with Flavio and we can merge as is.

PR #139, opened 2017-06-28
Increase salt-master timeout

kiall:
When dealing with a large number of minions, timeouts are visible when
using the default value of 5 seconds. Increasing the CPU/RAM resources
allocated to the master helps, but given it it's short bursts of heavy usage
(bootstrap and upgrade), this shouldn't be necessary.

We increase the timeout from 5 to 20 seconds, allowing tasks to take longer
yet still succeed.

### DIFF
--- a/config/master.d/master.conf
+++ b/config/master.d/master.conf
@@ -4,6 +4,7 @@ interface: 0.0.0.0
 event_return: mysql
 presence_events: True
 worker_threads: 10
+timeout: 20
 file_roots:
   base:
     - /usr/share/salt/kubernetes/salt


config/master.d/master.conf
```diff
@@ -4,6 +4,7 @@ interface: 0.0.0.0
 event_return: mysql
 presence_events: True
 worker_threads: 10
+timeout: 20
```
grahamhayes: this is not a job timeout - it is a time out for the salt master to connect to minions, and minions connecting to the salt master. see conversation on #126

kiall: https://docs.saltstack.com/en/latest/topics/troubleshooting/master.html#commands-time-out-or-do-not-return-output suggests and https://docs.saltstack.com/en/latest/topics/troubleshooting/master.html#salt-master-doesn-t-return-anything-while-running-jobs suggest this is exactly the timeout parameter I'm looking for?

PR #140, opened 2017-06-29
Handle curl proxy settings

flavio:
YaST is also configuring proxy settings inside of `/root/.curlrc`, this is needed because zypper is using libcurl. So if you run zypper from a cronjob or `su`, the `/etc/sysconfig/proxy` variables are not parsed
and set in the environment. Which means, zypper will not use the proxy and fail. With `/root/.curlrc`, libcurl will use the proxies configured there.

Thanks to @thkukuk for the [explanation](https://github.com/kubic-project/velum/pull/214#issuecomment-311910779).


kiall:
This won't set the proxy on the Admin node... However I'll add that to https://github.com/kubic-project/salt/pull/138

inercia:
Are you sure this is really necessary? I expected I would need to modify `curlrc` when I was working on #107 but I was surprised when I discovered I didn't need to...

PR #144, opened 2017-06-30
Test CI

kiall:


MaximilianMeister:
@kiall retriggered the test and it went green, not sure what was actually broken, or what has changed over the weekend

ereslibre:
@MaximilianMeister yeah... nothing changed over the weekend AFAIK, I have been observing that puma gets restarted several times after it starts, what leads to *sometimes* fail when capybara tries to make a request to it right when it's being restarted. Not sure why it gets restarted so often (can be checked on the e2e-tests logs). So we sometimes get the Capybara::Poltergeist error because puma isn't up... I'm currently trying to find out why this happens (why it gets restarted).

PR #146, opened 2017-07-04
Remove unneeded signing policies

ereslibre:
These signing policies were used when the CA wasn't containerized,
when we containerized it, they were moved to `caasp-container-manifests`,
and the CA container is mounting it from there.

If we uncontainerize the CA in the future we can move it back if
needed, but let's keep this clean so it's not misleading.


flavio:
Correct me if I'm wrong, but the etcd cleanups are not related with this PR. If that's the case they should go in with another PR

ereslibre:
@flavio Yes, you are right, I kept it as a separate commit for this reason. Cherry picking it to another PR.

ereslibre:
Splitted in https://github.com/kubic-project/salt/pull/149

PR #149, opened 2017-07-05
Cleanup etcd config files

ereslibre:
The `etcd.conf` file is no longer used, so remove the related logic around it.


ereslibre:
@inercia You are right. I will keep the `etcd` folder for hygiene reasons, but will remove `etcd` from the orchestration.

inercia:
> @inercia You are right. I will keep the etcd folder for hygiene reasons, but will remove etcd from the orchestration.
> Update: the etcd related code was already removed from the orchestration and top file, I'd keep this as it is.

I would suggest to remove it altogether. It is not going to be used anywhere, and it will be confusing to keep there some dead code that doesn't even work .

ereslibre:
> I would suggest to remove it altogether. It is not going to be used anywhere, and it will be confusing to keep there some dead code that doesn't even work .

`etcd-proxy` mainly includes `etcd`, so it's not technically dead what's already there. Do you suggest to move all the `etcd` code to `etcd-proxy`? I think the current separation is good to have, but I also get your point. However, the point is that with this PR there's literally no dead code.

inercia:
> etcd-proxy mainly includes etcd, so it's not technically dead what's already there. Do you suggest to move all the etcd code to etcd-proxy?

Yeah, I'm suggesting right that: to move everything from `etcd` to `etcd-proxy`. Having `etcd` as well as an `etcd-proxy` _subclass_ made sense because we were _instantiating_ both classes, but once we do not instantiate one of them, it makes sense to just merge everything.

PI-Victor:
what's the status of this? @ereslibre when you have some time can you rebase?

PR #151, opened 2017-07-06
When generating the certificate use the pillar path

ereslibre:
Since we added the minion certificate location to the pillar, also
take the public key location from the pillar, or the certificate
generation will fail if the pillar value changes.


flavio:
Sorry for being late, tested and it works. Thanks for having merged it.

PR #155, opened 2017-07-10
Grow flannel CIDR to accommodate 1024 workers

kiall:
Flannel was setup such that 150 workers could obtain a subnet before there
were not none left. By growing this range, and the size of the individual
allocations, we allow for up to 1024 workers with 510 pods on each.

bsc#1047847

### DIFF
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -1,21 +1,28 @@
 # the CIDR for cluster IPs (internal IPs for Pods)
-cluster_cidr:     '172.20.0.0/16'
+cluster_cidr:     '172.16.0.0/13'
+
+# The min and max subnet, and length for node allocations.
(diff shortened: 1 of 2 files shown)


kiall:
CI still failing, I've given it two runs. Logging into the CI box, I see:

	alchiba:~/:[0]# ip a show docker0
	8: docker0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN group default 
	    link/ether 02:42:94:5a:c9:90 brd ff:ff:ff:ff:ff:ff
	    inet 172.17.0.1/16 scope global docker0
	       valid_lft forever preferred_lft forever
	alchiba:~/:[0]# 

we should probably try to avoid dockers default value - moving back to WIP

PR #157, opened 2017-07-10
Revert "Set MySQL as the job cache for the Salt master"

kiall:
This reverts commit de22c660a99bc1425295c86be7d7dc3e79089845 / https://github.com/kubic-project/salt/pull/132

For some reason, enabling this means devenv and CI is broken:

                  ----------
                            ID: /etc/pki/minion.crt
                      Function: x509.certificate_managed
                        Result: False
                       Comment: An exception occurred in this state: Traceback (most recent call last):
                                  File "/usr/lib/python2.7/site-packages/salt/state.py", line 1746, in call
                                    **cdata['kwargs'])
                                  File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1704, in wrapper
                                    return f(*args, **kwargs)
                                  File "/usr/lib/python2.7/site-packages/salt/states/x509.py", line 494, in certificate_managed
                                    new = __salt__['x509.create_certificate'](testrun=True, **kwargs)
                                  File "/usr/lib/python2.7/site-packages/salt/modules/x509.py", line 1374, in create_certificate
                                    arg=str(kwargs))[ca_server]
                                KeyError: 'ca'
                       Started: 19:35:32.171471
                      Duration: 5323.054 ms

I don't have the slighest clue why these are related.

### DIFF
--- a/config/master.d/master.conf
+++ b/config/master.d/master.conf
@@ -5,11 +5,6 @@ event_return: mysql
 presence_events: True
 worker_threads: 10
 timeout: 20
-
-master_job_cache: mysql
-job_cache: False
-keep_jobs: 24
-
 file_roots:
   base:
     - /usr/share/salt/kubernetes/salt


kiall:
Possibly related?

    (1364, "Field 'alter_time' doesn't have a default value")[ERROR   ] Could not store events - returner 'mysql.event_return' raised exception: (1364, "Field 'alter_time' doesn't have a default value")

kiall:
@mssola ping - FYI, something is wrong with the original PR.

ereslibre:
@kiall it's related, very probably. That error prevents the salt-master from storing events in the mysql database. `alter_time` requires a `CURRENT_TIMESTAMP()` default function, and rails 4 only supports default column values, not functions (rails 5 supports them though). 

In the `devenv` it's solved here: https://github.com/kubic-project/caasp-devenv/blob/master/patch_schema.

This script copies the patched schema to `/var/lib/velum/schema.rb`, and this schema is used of present from `bin/init` in the `velum` project: https://github.com/kubic-project/velum/blob/master/bin/init#L41

mssola:
@ereslibre @kiall so, shouldn't we add this patch on `alter_time` instead of reverting this PR ?

ereslibre:
@mssola this patch should be applied in the development registry image:

```
~ docker run --rm docker-testing-registry.suse.de/devel/sles12/velum-development:latest cat /srv/velum/db/schema.rb | grep CURRENT_TIMESTAMP
    t.column   "alter_time",   "DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP"
    t.column   "alter_time", "DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP"
```

And since the `start` script calls the `patch_schema` script, the `sles12/velum:development` image should contain it.

```
~ docker run --rm sles12/velum:development cat /var/lib/velum/schema.rb | grep CURRENT_TIMESTAMP
    t.column   "alter_time",   "DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP"
    t.column   "alter_time", "DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP"
```

Not sure in what context this is failing for you so I cannot really tell about what the problem might be.

ereslibre:
I've just tried again and with the patch in this PR I am not getting the error you report @kiall. I suspect this hasn't to do with this PR somehow, but need to understand the context better.

ereslibre:
The error about the `ca` key missing is usually an error on the `ca` container itself, some kind of trouble, usually the `ca` not being registered against the salt-master, or some other connectivity error with it... at least that has been my experience with that error.

kiall:
> @kiall so, shouldn't we add this patch on alter_time instead of reverting this PR ?

When devenv and CI is broken, it's best to back out the suspect change, we can always add the code bak when people are no longer blocked :)

> I've just tried again (cleaning all my docker images) and with the patch in this PR (removing the revert) I am not getting the error you report @kiall. I suspect this hasn't to do with this PR somehow, but need to understand the context better.

Every CI run (both concouse and Jenkins) since the original PR landed had failed, and they all turned green again as soon as the revert merged! I have to admit, I don't quite understand why the DB schema error would result in missing mine data, but that seems to be exactly what happened.

Let me push a PR with the change again, I'm willing to bet it fails.

PR #165, opened 2017-07-13
Add some extra names to the API server certificate (bsc#1033671)

inercia:
Added:

* kubernetes
* kubernetes.default
* kubernetes.default.svc

### DIFF
--- a/salt/cert/init.sls
+++ b/salt/cert/init.sls
@@ -16,7 +16,10 @@ include:
   {% endfor %}
 
   # add some extra names the API server could have
-  {% set extra_names = extra_names + ["DNS: api",
+  {% set extra_names = extra_names + ["DNS: kubernetes",
+                                      "DNS: kubernetes.default",
+                                      "DNS: kubernetes.default.svc",
+                                      "DNS: api",
                                       "DNS: api." + pillar['internal_infra_domain']] %}
   {% for extra_name in pillar['api']['server']['extra_names'] %}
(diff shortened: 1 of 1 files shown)


salt/cert/init.sls
```diff
@@ -16,7 +16,12 @@ include:
   {% endfor %}
 
   # add some extra names the API server could have
-  {% set extra_names = extra_names + ["DNS: api",
+  {% set extra_names = extra_names + ["DNS: kubernetes",
+                                      "DNS: kubernetes.default",
+                                      "DNS: kubernetes.default.svc",
+                                      "DNS: kubernetes.default.svc.cluster",
+                                      "DNS: kubernetes.default.svc.cluster.local",
```
kiall: "cluster.local" -> is a pillar value, we should pull it from there. Specically, it's the dns:domain pillar.

The issue will be splitting it properly, since it could be replaced with for example "cluster.mycompany.com" - which would need 3 entries instead of two.

inercia: I've just realized we are adding those names for `pillar["dns:domain"]` [here](https://github.com/kubic-project/salt/blob/b48feae530cbd51ef6ca4c00903893cada020eb7/salt/cert/init.sls#L38). I guess can remove these `cluster.local` entries...

kiall:
Backported to 1.0 via https://github.com/kubic-project/salt/pull/170

PR #167, opened 2017-07-14
Upgrade: Wait longer for minions to reboot

kiall:
Wait 1200 seconds (20 minutes) for minions to reboot, instead of the
default 300 seconds (5 minutes). We increase this to cover off cases
where slower to boot physical hardware is used.

20 minutes was chosen as, I've seen physical hardware take 10-12 minutes
in the past, and someone likely has something that is slower to reboot.

bsc#1048683

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -60,6 +60,7 @@ update_modules:
 {{ master_id }}-wait-for-start:
   salt.wait_for_event:
     - name: salt/minion/*/start
+    - timeout: 1200
     - id_list:
       - {{ master_id }}
     - require:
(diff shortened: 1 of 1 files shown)


salt/orch/update.sls
```diff
@@ -60,6 +60,7 @@ update_modules:
 {{ master_id }}-wait-for-start:
   salt.wait_for_event:
     - name: salt/minion/*/start
+    - timeout: 1200
```
inercia: What about moving it to a pillar? It would be easier to find if we need to change it in the future...

kiall: Sadly, last time I tried to access the pillar here, it wouldn't work - so I went with a long enough period that anything should be able to reboot, but not so long people will be waiting hours to see the failure.

kiall:
Backported in https://github.com/kubic-project/salt/pull/171

PR #169, opened 2017-07-17
Explicitly refresh the mine on all minions after the `ca` has published the `ca.crt`

ereslibre:
We will explicitly force all minions to refresh the mine after the `ca` minion has
published the `ca.crt` certificate on the mine, to avoid rendering problems with
later SLS being executed. It might happen that a minion was missing this information
on its mine, so the rendering of the SLS failed, effectively stopping the whole
orchestration process.

Fixes: bsc#1048548

### DIFF
--- a/salt/orch/kubernetes.sls
+++ b/salt/orch/kubernetes.sls
@@ -66,8 +66,14 @@ etcd_discovery_setup:
     - sls:
       - etcd-discovery
     - require:
-      - salt: ca_setup
       - salt: update_modules
+
+refresh_mine:
+  salt.function:
+    - tgt: '*'
+    - name: mine.update
+    - require:
+      - salt: ca_setup
 
 etcd_proxy_setup:
   salt.state:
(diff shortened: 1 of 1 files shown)


ereslibre:
This is a `WIP` because I'm trying to find the way to reproduce the exact problem found on the production environment. It's proving itself to be really hard to reproduce without modifying the orchestration source code.

salt/orch/kubernetes.sls
```diff
@@ -66,9 +66,15 @@ etcd_discovery_setup:
     - sls:
       - etcd-discovery
     - require:
-      - salt: ca_setup
```
inercia: I think the `ca_setup` is still a requirement here: we would need the CA in order to have certificates for `etcdctl` for setting up the discovery `etcd`, right?

ereslibre: No, that's why I removed it, if you check `etcd-discovery` this targets the etcd instance on the admin node with `http` for discovery purposes only: https://github.com/kubic-project/salt/blob/master/salt/etcd-discovery/init.sls#L3-L4.

I removed it since it's not a real requirement.

inercia: Ok, then it is fine.

ereslibre:
Backport PR: https://github.com/kubic-project/salt/pull/172

PR #171, opened 2017-07-17
Upgrade: Wait longer for minions to reboot

kiall:
Wait 1200 seconds (20 minutes) for minions to reboot, instead of the
default 300 seconds (5 minutes). We increase this to cover off cases
where slower to boot physical hardware is used.

20 minutes was chosen as, I've seen physical hardware take 10-12 minutes
in the past, and someone likely has something that is slower to reboot.

bsc#1048683

Backport of https://github.com/kubic-project/salt/pull/167

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -60,6 +60,7 @@ update_modules:
 {{ master_id }}-wait-for-start:
   salt.wait_for_event:
     - name: salt/minion/*/start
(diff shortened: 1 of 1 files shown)


kiall:
Concourse CI is no longer triggering for release-1.0 branches, with 3x +A and a pass from Jenkins, this should be OK to merge.

PR #173, opened 2017-07-18
Don't wait for minion responses when rebooting

kiall:
When we instruct a minion to reboot, we can't reliably
wait for the response from salt-minion letting us know
that the "systemctl reboot" command succeeded, as systemd
may choose to shutdown the salt-minion service before it
can sent out the "Yes, that worked" response.

Salt does not make any attempt to finish in progress
tasks when it receives a SIGTERM, leaving us with few
other viable choices for this.

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -51,6 +51,7 @@ update_modules:
   salt.function:
     - tgt: {{ master_id }}
     - name: cmd.run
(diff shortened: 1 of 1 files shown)


kiall:
Backport to 1.0 in https://github.com/kubic-project/salt/pull/175

PR #174, opened 2017-07-19
Add extra requisites to the update orchestration

kiall:
These additional requisites enforce a stricter ordering of tasks
during the upgrade. In some case, "-set-update-grain" would not
execute in the right place, potentially leading to a failed
upgrade.

bsc#1045381


kiall:
Backport in https://github.com/kubic-project/salt/pull/176

PR #177, opened 2017-07-19
WIP: Refresh CA mine first, then the rest

ereslibre:
After publishing the `ca.crt` into the mine, we need to make sure that
the `mine.update` happens first on the `ca`, so we know that it syncs
this information with the master first, then we can proceed to run
`mine.update` on all nodes, so they again sync with the master and
find the information the `ca` minion published on it.

Fixes: bsc#1049137
Fixes: bsc#1048548

### DIFF
--- a/salt/orch/kubernetes.sls
+++ b/salt/orch/kubernetes.sls
@@ -59,6 +59,21 @@ ca_setup:
       - salt: etc_hosts_setup
       - salt: update_mine
 
+refresh_ca_mine:
+  salt.function:
+    - tgt: 'roles:ca'
+    - tgt_type: grain
+    - name: mine.update
+    - require:
+      - salt: ca_setup
+
+refresh_all_mines:
+  salt.function:
+    - tgt: '*'
+    - name: mine.update
+    - require:
+      - salt: refresh_ca_mine
+
 etcd_discovery_setup:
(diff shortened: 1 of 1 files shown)


ereslibre:
Adding reviewers despite of being marked as `WIP`. It's flagged as `WIP` because we need to verify this fixes the issue completely in a fresh cluster under the same conditions the error was 100% reproducible.

We should be able to verify this fix tomorrow, and possibly get this into 1.0 if at all possible.

kravciak:
Bootstrap passed on XEN-PV with this fix

ereslibre:
This working is just a coincidence. Reading salt's source code, `mine.send` will asynchronously send an event with the new information, and wait a fixed 0.5 seconds (https://github.com/saltstack/salt/blob/42bb1a64caa65bbf0649c50800a0e62897ba654c/salt/modules/mine.py#L60-L66).

We are adding more operations between the `mine.send` action and the real usage in `etcd_proxy_setup`, so it's more likely for the information to be there, but is not a real fix for the original issue.

This means that sometimes this mine information is there and sometimes not, depending on different factors.

I'm currently checking if there's a way to do an active wait for the mine information to be there (check contents, not wait for a fixed amount of time).

PR #180, opened 2017-07-20
Do not publish the `ca.crt` from the `ca` SLS, use `mine_functions`

ereslibre:
We will be publishing this contents when the `ca` minion starts, so there's no need
to do this during the orchestration.

`mine.send` is not reliable enough since we cannot confirm that the contents are there
yet, and waiting a random amount of time is not appropriate as we are just hiding the
real problem. In the near future we can do an active wait for the content to be there
using `retry`, but for now we just publish the contents of the `ca.crt` using
`mine_functions`, so it is sent when the `ca` minion starts.

There's no need to refresh the mine, as this was just hiding the real problem when
we were publishing this contents during the orchestration phase.

Fixes: bsc#1049137
Fixes: bsc#1048548

Backport of https://github.com/kubic-project/salt/pull/179/


kiall:
Clean cherry-pick from a just approved master branch change, merging without the second +A

PR #181, opened 2017-07-20
Delay reboots during upgrade by 15 seconds

kiall:
Even with backgrounding the call, salt-minion sometimes still does not
have enough time to respond before systemd shuts down salt-minion on
some environments. By adding a 15 second delay, we give salt-minion
much more time than it should need in a healthy cluster to respond.

Additionally, switch from the deprecated syntax for supplying bg=True,
to the newer syntax which no longer logs a warning.

Followup up fix for bsc#1049200

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -57,9 +57,10 @@ update_modules:
   salt.function:
     - tgt: {{ master_id }}
     - name: cmd.run
-    - bg: True
     - arg:
-      - systemctl reboot
+      - sleep 15; systemctl reboot
+    - kwarg:
+        bg: True
     - require:
       - salt: {{ master_id }}-clean-shutdown
 
@@ -133,9 +134,10 @@ update_modules:
   salt.function:
     - tgt: {{ worker_id }}
     - name: cmd.run
-    - bg: True
     - arg:
-      - systemctl reboot
+      - sleep 15; systemctl reboot
+    - kwarg:
+        bg: True
     - require:
       - salt: {{ worker_id }}-clean-shutdown
 


kiall:
FYI - I'm validating this in one of our QA environments by doing a series of upgrades over+over.

salt/orch/update.sls
```diff
@@ -57,9 +57,10 @@ update_modules:
   salt.function:
     - tgt: {{ master_id }}
     - name: cmd.run
-    - bg: True
     - arg:
-      - systemctl reboot
+      - sleep 15; systemctl reboot
```
ereslibre: I think we can consider using https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.system.html#salt.modules.system.reboot (it can wait minutes for the reboot to happen too).

kiall: If I remember correctly from @grahamhayes - this didn't actually work from orchestration, but it's been a while.

ereslibre: Might be, the current code we are using does not use much magic: https://github.com/saltstack/salt/blob/2016.11/salt/modules/system.py#L95-L96

Maybe the problem was related to an immediate reboot as we have been experiencing running `systemctl reboot` ourselves.

grahamhayes: The issue was the shutdown killing the minion before it could return - I wasn't aware it could take a time parameter.

The `system.reboot` with `at_time` param may be better, but if this works, we can come back to it.

inercia: Can't we just do `shutdown reboot +1`?

kiall: @inercia on a 50 node cluster, 1 minute will be nearly an hour delay in the upgrade - hence trying to get a smaller number than 1 minute.

ereslibre: I think it's good enough, it's also possible to build the reboot time (now + 15 seconds) and provide it as an absolute value (`shutdown -r absolute`) instead of (`shutdown -r +mins`), but then you have funny clock skew problems and alike...

Keeping the LGTM, no changes required from my side.

inercia: I don't think we should be worried about too long upgrades: the cluster should work fine. But anyway, the `sleep` should do the job...

kiall:
Okay - Successfully completed 10 upgrades of the QA env which was reliably exhibiting this problem.

kiall:
Backport @ https://github.com/kubic-project/salt/pull/182

PR #183, opened 2017-07-26
Make the kubernetes version configurable

inercia:
Make the kubernetes version configurable
Some minor improvements in dependencies between states.
Use the [recommended admission controllers](https://kubernetes.io/docs/admin/admission-controllers/#is-there-a-recommended-set-of-plug-ins-to-use) depending on the k8s version.

### DIFF
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -39,6 +39,12 @@ kube_log_level:   '2'
 # install the addons (ie, DNS)
 addons:           'false'
 
+versions:
+  # kubernetes version to use
+  # it can be "" (no specifi version is foced, just installed), "latest",
+  # a specific version or (for Salt>=2017.7.0) a version with wildcards (ie, "1.5*")
+  kubernetes:     ''
+
 ssl:
   enabled:        true
   ca_dir:         '/etc/pki/trust/anchors'
--- a/salt/kubernetes-master/apiserver.jinja
+++ b/salt/kubernetes-master/apiserver.jinja
@@ -4,6 +4,8 @@
 # The following values are used to configure the kube-apiserver
 #
 
++{% set kubernetes_version = salt['pillar.get']('versions:kubernetes', '') %}
+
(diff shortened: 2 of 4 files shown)


flavio:
JFYI, the updates repositories have no history of the older releases of kubernetes. There are always going to be two versions of kubernetes available:

  * The one shipped with the GA product (inside of the GA repo)
  * The latest one available from the updates channel

We can set the pillar to hold the version of kubernetes we consider stable.

inercia:
> JFYI, the updates repositories have no history of the older releases of kubernetes. There are always going to be two versions of kubernetes available:
>  The one shipped with the GA product (inside of the GA repo)
> The latest one available from the updates channel
> We can set the pillar to hold the version of kubernetes we consider stable.

@flavio Yeah, my intention was to force a match between the Salt scripts version and kubernetes version. This cannot install a specific version of k8s in CaaSP. However, I'm currently using this for development (for working with k8s 1.7)...

robdaemon:
Overall, I'm a 👎 on this entirely. As @flavio said, we don't have a history of versions, and you can't force the product to install a specific version. I don't like the idea of adding the conditionals - they're not actually going to be used, and just make the code more complicated where it does not need to be.

kiall:
I agree with @robdaemon here, we shouldn't make our code unnecessarily complex to support a scenario we're not actually going to support.

Having code like this is fine as a transitional measure to ease the upgrade to 1.7, or if we've actively made the decision to support multiple versions of kubernetes concurrently. But, we haven't, and I don't believe we have the resources to do so.

If we have code like this, it should be time boxed just to handle the transition + should be removed right after - and - we should certainly not make it look like we support 1.4 through 1.7.

kiall:
1.7 is now in and working (well - it bootstraps, I don't think anyone has checked does it really work), i.e. transitional code from 1.5 -> 1.7 shouldn't be needed anymore.

PR #186, opened 2017-07-28
[WIP] Use CNI as the kubernetes network fabric

inercia:
**Note**: this requires _Kubernetes v1.7_ (CI will not pass) and some packages that are currently not installed in CaaSP 1.0 images.

PR #187, opened 2017-07-28
Run kubelet and kube-proxy on the master node

robdaemon:
A standard Kubernetes installation runs a kubelet and kube-proxy on
every node, and then you decide where to run apiserver,
controller-manager and scheduler.

This change is required to support RBAC, DaemonSets and many other
changes.

Requires an updated kubernetes-client package that contains: https://build.opensuse.org/request/show/494998

Depends-On: https://github.com/kubic-project/testinfra/pull/1


robdaemon:
This is a precursor to splitting all the Kubernetes daemons into unique salt formulas, and I need it to continue my RBAC work (specifically, I need kube-proxy on the master, but we should put kubelet there too)

kiall:
And, we have a change that means updating the infrastructure tests - in this case, the extra kubelet means we have more nodes in `kubectl get nodes` than expected.

Other than this, it looks to have worked just fine. We should figure out how to handle cross repo PR deps sooner rather than later :)

salt/kubernetes-common/init.sls
```diff
@@ -0,0 +1,9 @@
+/etc/kubernetes/config:
```
inercia: This will conflict with [my kubernetes-common](https://github.com/kubic-project/salt/pull/186/files#diff-d1fc4951f00f2000a4e57d72366b959eR1).

salt/kubernetes-common/init.sls
```diff
@@ -0,0 +1,9 @@
...
+  file.managed:
+    - source:     salt://kubernetes-common/config.jinja
```
inercia: `/etc/kubernetes/config` is installed by the `kubernetes-common` package, so we must explicitly install it here _before_ trying to write our own `config`. Otherwise the _rpm_ will overwrite it...

salt/kubernetes-minion/init.sls
```diff
@@ -65,11 +76,13 @@ kubelet:
...
   # Salt doesn't seem to have a retry mechanism in the version were using, so I'm
   # doing a horrible hack right now.
+  # RAR: Increasing the timeout to 5 minutes, since this now occurs during the initial
+  # bootstrap - it takes more than 60 seconds before kube-apiserver is running.
   cmd.run:
     - name: |
         ELAPSED=0
         until output=$(kubectl uncordon {{ grains['caasp_fqdn'] }}) ; do
-            [ $ELAPSED -gt 60 ] && exit 1
+            [ $ELAPSED -gt 300 ] && exit 1
```
inercia: What about configuring it from the pillar? Something like `reboot_timout`?

robdaemon:
@kiall added explicit dependency on kubernetes-common rpm. This is ready to go.

PR #189, opened 2017-08-02
update k8s version

monstermunchkin:
Signed-off-by: Thomas Hipp <thipp@suse.de>


monstermunchkin:
> Aside: Should our salt package versioning really be tied to the K8S release we're using? I would have thought we should tie this to the CaaSP version, the same as velum.

I just thought the versions should match and/or it is somehow important they matched. If this is not the case, we can ignore this PR.

kiall:
It might be important, and they certainly match currently, so it's safe bet to assume this is a valid PR.

I just question if we should change how we're versioning our salt states as some point in the future :)

PR #190, opened 2017-08-04
WIP: Wait for Kube-API before installing Kube-DNS

kiall:
Wait for Kube-API to start and give a success to the /healthz endpoint before bringing up Kube-DNS, as otherwise we can run into an issue where the API extensions are not loaded.

See https://github.com/kubernetes/kubernetes/issues/47739 and it's resolving PR https://github.com/kubernetes/kubernetes/issues/49072

### DIFF
--- a/salt/kubernetes-master/init.sls
+++ b/salt/kubernetes-master/init.sls
@@ -70,7 +70,7 @@ kube-apiserver:
   cmd.run:
     - name: |
         ELAPSED=0
-        until curl --insecure --silent --fail -o /dev/null https://127.0.0.1:{{ pillar['api']['ssl_port'] }}/healthz ; do
+        until curl --insecure --silent --fail -o /dev/null http://127.0.0.1:8080/healthz ; do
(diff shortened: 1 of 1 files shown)


kiall:
I'm not sure why this is failing, I've not yet been able to test it

salt/kubernetes-master/init.sls
```diff
@@ -64,6 +64,20 @@ kube-apiserver:
       - file:     kube-apiserver
       - sls:      ca-cert
       - sls:      cert
+  # We need to wait for Kube API server to actually start, see k8s issue #47739
+  # TODO: Salt doesn't seem to have a retry mechanism in the version were using,
+  # so I'm doing a horrible hack right now.
```
robdaemon: I use this hack all over the place when I interact with kubectl. I don't like it, but... not a lot I can do. :(

robdaemon:
Okay, tests seem to pass now. Going to merge this in now to stabilize the build.

PR #191, opened 2017-08-04
Separate salt formulas per daemon

robdaemon:
Separating out each daemon into its own formula for composability and certs (needed for RBAC)


robdaemon:
Okay, the basics work - everything is a separate formula now. Starting on breaking apart the certs...

Also, the error message in Jenkins doesn't tell me why it thinks the build's broken. It installs here just fine.

robdaemon:
Okay this is all of it. Separate certificates for each kubernetes daemon with the right `O` entries to unblock RBAC. Seems to pass all the tests.

PR #192, opened 2017-08-09
Role-based access control

robdaemon:
Enabled full RBAC and node authorization, working on getting LDAP fully integrated next


robdaemon:
Related: kubic-project/caasp-container-manifests#89

robdaemon:
Will be adding a Velum change soon too, that makes velum use LDAP instead of the database for auth.

robdaemon:
Moving to sles-based containers for dex will come as a separate PR.

robdaemon:
😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 

devenv uses etcd 3.2.4

CaaSP nodes use etcd 3.1.0

😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 😱 

I feel like that mismatch may have just contributed to the commit I just added that forces kubernetes to use etcd v2 protocol

salt/dex/dex.yaml
```diff
@@ -0,0 +1,113 @@
...
+
+    oauth2:
+      skipApprovalScreen: true
+
+    staticClients:
+    - id: caasp-cli
+      redirectURIs:
+      - 'http://127.0.0.1'
+      name: "CaaSP CLI"
+      secret: "swac7qakes7AvucH8bRucucH"
```
robdaemon: This is a tough one. The CLI needs the exact same secret to perform the OAuth2 handshake. So, if we generate it, everyone who uses the CLI will need it in a local file.

robdaemon: As of right now, I'm embedding that secret in the CLI at compile-time.

robdaemon:
@kiall @grahamhayes this is ready to go.

salt/dex/init.sls
```diff
@@ -0,0 +1,124 @@
...
+  {% for extra_name in pillar['api']['server']['extra_names'] %}
+    {% do extra_names.append("DNS: " + extra_name) %}
+  {% endfor %}
+
+  # add the fqdn provided by the user
+  # this will be the name used by the kubeconfig generated file
+  {% if salt['pillar.get']('api:server:external_fqdn') %}
+    {% do extra_names.append("DNS: " + pillar['api']['server']['external_fqdn']) %}
+  {% endif %}
+{% endif %}
```
ereslibre: We probably want to move this whole block to a jinja macro... Similar code (if not the same) exists in https://github.com/kubic-project/salt/blob/1e41512047a193d30ad06f78e5fe7a19faeb4824/salt/cert/init.sls#L4-L32 and https://github.com/kubic-project/salt/blob/930ea6e8ee03555c27b018f0af93bb65a03e423d/salt/kube-apiserver/init.sls#L8-L40

robdaemon: Fair point - they're similar but I think may have slightly different names they generate.

robdaemon:
@kiall @grahamhayes @ereslibre rebased again, ready to merge

salt/dex/init.sls
```diff
@@ -0,0 +1,124 @@
...
+    - days_remaining: {{ pillar['certificate_information']['days_remaining']['certificate'] }}
+    - backup: True
+    - user: root
+    - group: root
+    - mode: 644
+    - require:
+      - sls:  crypto
+      - x509: /etc/pki/dex.key
+
+/root/dex.yaml:
```
kiall: Nit: Is there anywhere else we can put this, and the roles.yaml, other than `/root`? If not.. ignore.

Leaving stuff around in /root has always annoyed me ;)

salt/dex/roles.yaml
```diff
@@ -0,0 +1,47 @@
...
+  apiGroup: rbac.authorization.k8s.io
+- kind: Group
+  name: system:unauthenticated
+  apiGroup: rbac.authorization.k8s.io
+roleRef:
+  kind: Role
+  name: find-dex
+  apiGroup: rbac.authorization.k8s.io
+---
+# Map the LDAP Administrators role to the Kubernetes system:masters group
```
kiall: This feels like it shouldn't be part of the base install, as not all LDAP's will have an "Administrators" group which is intended to map to the system:masters role.

Maybe making this a pillar value is enough?

salt/dex/dex.yaml
```diff
@@ -0,0 +1,136 @@
...
+        inCluster: true
+    web:
+      https: 0.0.0.0:5556
+      tlsCert: /etc/dex/tls/dex.crt
+      tlsKey: /etc/dex/tls/dex.key
+    connectors:
+    - type: ldap
+      id: ldap
+      name: LDAP
+      config:
```
kiall: I think we'll need to make most of this block configurable to allow for integrating with a pre-existing LDAP instance.. However, I think that should come as a follow up PR - this is large enough already.

salt/dex/dex.yaml
```diff
@@ -0,0 +1,136 @@
...
+
+          nameAttr: cn       
+
+    oauth2:
+      skipApprovalScreen: true
+
+    staticClients:
+    - id: caasp-cli
+      redirectURIs:
+      - 'http://127.0.0.1'
```
kiall: Shouldn't this be set to the special out of browser URI? `urn:ietf:wg:oauth:2.0:oob`

salt/dex/init.sls
```diff
@@ -0,0 +1,124 @@
...
+  {% endfor %}
+  {% for extra_ip in pillar['api']['server']['extra_ips'] %}
+    {% do ip_addresses.append("IP: " + extra_ip) %}
+  {% endfor %}
+  # add some extra names the API server could have
+  {% set extra_names = extra_names + ["DNS: dex",
+                                      "DNS: dex.kube-system",
+                                      "DNS: dex.kube-system.svc",
+                                      "DNS: dex.kube-system.svc." + pillar['internal_infra_domain'],
+                                      "DNS: api." + pillar['internal_infra_domain']] %}
```
kiall: Does DEX need a cert to cover the `api.` name?

salt/dex/init.sls
```diff
@@ -0,0 +1,124 @@
...
+    - user: root
+    - group: root
+    - mode: 644
+    - require:
+      - file: /root/dex.yaml
+
+dex_secrets:
+  cmd.run:
+    - name: |
+        until kubectl get secret dex-tls --namespace=kube-system ; do
```
kiall: Q - Why are we running these in a loop like this?

Any updates to these files wont be applied, as the `until` will return true, and the the `kubectl create` will never happen.

Ideally, these would all be changed to `kubectl apply` - if `apply` is used from the beginning, additionally metadata is included that allows easier updates without doing a whole delete/recreate cycle on unchanged resources.

salt/etc-hosts/hosts.jinja
```diff
@@ -20,7 +20,7 @@
 	{%- else %}
 		{%- set ip = addrlist|first %}
 	{%- endif %}
-{{ ip }} {{ master_id }}.{{ pillar['internal_infra_domain'] }}
+{{ ip }} {{ master_id }}.{{ pillar['internal_infra_domain'] }} {{ pillar['api']['server']['external_fqdn'] }}
```
kiall: I believe this needs to be moved to the 127.0.0.1 entry above? With multiple masters, we'll end up duplicating this name over multiple IPs

salt/kube-apiserver/apiserver.jinja
```diff
@@ -36,8 +36,21 @@ KUBE_API_ARGS="--advertise-address={{ grains['ip4_interfaces']['eth0'][0] }} \
...
                {{ salt['pillar.get']('components:apiserver:args', '') }} \
                --client-ca-file={{ pillar['ssl']['ca_file'] }} \
-               --storage-backend={{ pillar['api']['etcd_version'] }}"
+               --storage-backend={{ pillar['api']['etcd_version'] }} \
+               --service-account-key-file={{ pillar['ssl']['key_file'] }} \
+               --service-account-lookup=true \
+               --runtime-config=admissionregistration.k8s.io/v1alpha1 \
+               --authorization-mode=Node,RBAC \
+               --oidc-issuer-url=https://{{ pillar['api']['server']['external_fqdn'] }}:{{ pillar['dex']['node_port'] }} \
+               --oidc-client-id=caasp-cli \
```
kiall: Q - Is reusing the `caasp-cli` client ID here required?

salt/top.sls
```diff
@@ -21,12 +21,13 @@ base:
...
     - kubelet
     - kube-proxy
   'roles:kube-master':
     - match: grain
     - kube-apiserver
     - kube-controller-manager
     - kube-scheduler
-    - kubectl-client-cert
     - addons
+    - dex
```
kiall: I think I would put dex before the addons, as addons might rely on RBAC stuff being up-to-date

salt/orch/kubernetes.sls
```diff
@@ -131,6 +131,16 @@ reboot_setup:
     - require:
       - salt: kube_master_setup
 
+dex_setup:
```
kiall: This block is unnecessary, dex is in the highstate via `salt/top.sls`, which is ran against the masters 3 states up.

salt/dex/roles.yaml
kiall: (This probably falls into the same "next PR" as the pre-existing LDAP comment below...)

salt/dex/dex.yaml
```diff
@@ -0,0 +1,136 @@
...
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  labels:
+    app: dex
+  name: dex
+  namespace: kube-system
+spec:
+  replicas: 3
```
inercia: It would be good if we could make this configurable, [as we do with the DNS replicas](https://github.com/kubic-project/salt/blob/master/salt/addons/addons/kubedns.yaml.jinja#L10).

salt/top.sls
robdaemon: No, this isn't important. Creation of the addons all use the kubectl config, which is cert-based auth, not username/password.

salt/dex/dex.yaml
robdaemon: I'm not terribly comfortable with it being less than 3 replicas because of the outage it would cause for users if this were set to 1 and were to go down. Configurable is fine, as long as the minimum is 3.

inercia: I agree that three is a good value. But I was thinking more on the case were three replicas are too many (ie, our devenv). I would set `replicas: {{ some pillar }}` in this file, and set the default pillar value to 3. Then the pillar could be modified manually (I use [my own script](https://github.com/inercia/caasp-tf/blob/master/resources/common/caaspctl) for this kind of things, with `caaspctl pillar set ...`).

PR #193, opened 2017-08-14
Add optional addon for helm's server "tiller"

rbwsam:
Setting `{ "addons": { "tiller": "true" } }` will install the tiller deployment and service.

```
# On salt master:
salt "*" state.apply pillar='{ "addons": { "tiller": "true" } }'
```


salt/addons/deploy_tiller.sh
```diff
@@ -0,0 +1,40 @@
...
+
+NS_ARGS="--namespace=kube-system"
+SRV_ARGS="--server=http://127.0.0.1:8080"
+
+echo "Deploying tiller on Kubernetes"
+
+TILLER=`eval "$KUBECTL_PATH $NS_ARGS $SRV_ARGS get deployment | grep tiller | cat"`
+if [ ! "$TILLER" ]; then
+	# use kubectl to create tiller deployment
+	$KUBECTL_PATH $NS_ARGS $SRV_ARGS create -f "$DIR/tiller.yaml"
```
SlickNik: Another option to consider is to combine the kubernetes deployment + service for tiller into a single app manifest (yaml). This way the add-on can be deployed by running a single "kubectl apply -f".

salt/addons/deploy_tiller.sh
```diff
@@ -0,0 +1,40 @@
...
+
+	echo "Tiller successfully deployed."
+else
+	echo "Tiller already deployed. Skipping."
+fi
+
+TILLER_SVC=`eval "$KUBECTL_PATH $NS_ARGS $SRV_ARGS get svc | grep tiller | cat"`
+if [ ! "$TILLER_SVC" ]; then
+	# use kubectl to create tiller service
+	$KUBECTL_PATH $NS_ARGS $SRV_ARGS create -f "$DIR/tiller-svc.yaml"
```
SlickNik: Prefer to use "kubectl apply -f"

salt/addons/create_namespace.sh
```diff
@@ -0,0 +1,27 @@
...
+abort() { echo "FATAL: $@" ; exit 1 ; }
+
+NS_ARGS="--namespace=kube-system"
+SRV_ARGS="--server=http://127.0.0.1:8080"
+
+echo "Creating kube-system namespace..."
+# use kubectl to create kube-system namespace
+NAMESPACE=`eval "$KUBECTL_PATH $SRV_ARGS get namespaces | grep kube-system | cat"`
+if [ ! "$NAMESPACE" ]; then
+	$KUBECTL_PATH $SRV_ARGS create -f "$DIR/namespace.yaml"
```
inercia: And could we just move this command to a `cmd.run` in the `sls` file? It doesn't make much sense to invoke a script with `cmd.script` just for running a command...

salt/addons/deploy_tiller.sh
inercia: Same here: I would just run `kubectl apply -f` from a `cmd.run` in the `sls` file...

salt/addons/deploy_tiller.sh
inercia: Ditto

salt/addons/init.sls
```diff
@@ -1,67 +1,39 @@
...
 
-/etc/kubernetes/addons/kubedns-svc.yaml:
-  file.managed:
-    - source:      salt://addons/addons/kubedns-svc.yaml.jinja
-    - template:    jinja
-
-deploy_dns.sh:
-  cmd.script:
-    - source:      salt://addons/deploy_dns.sh
+apply_dns:
```
inercia: Looking at the Salt code, I think this should be named `apply-dns` according to the implicit code convention. The same would apply to all the other `apply_*`.

salt/addons/init.sls
```diff
@@ -14,40 +10,44 @@ include:
...
-    - template:   jinja
-
-/etc/kubernetes/addons/kubedns-cm.yaml:
-  file.managed:
-    - source:      salt://addons/addons/kubedns-cm.yaml.jinja
-    - template:    jinja
+apply-namespace:
+  cmd.run:
+    - name: |
+        kubectl apply -f /etc/kubernetes/addons/namespace.yaml
```
inercia: We knew this would happen, but I don't think running `kubectl apply` twice could be considered safe. The only safe solution would be to run this only once in the cluster, ideally it in the Salt master with the k8s Salt module...

robdaemon: The Salt master isn't configured with an appropriate kubeconfig file at this time. If we want to go that route, we need to open a separate story / Trello card for that, as there are several places that would benefit from that.

salt/addons/init.sls
robdaemon: We have `_`'s all over the place. There is no established convention on this.

salt/addons/init.sls
kiall: Actually, this makes a lot of sense. We (briefly) chatted about this on IRC earlier and I had suggested either the admin node (which is actually much harder than it should be), or salts `subset: 1` matcher, which we don't have in our version of salt.

The salt master container could easily be setup to handle these "once per cluster" kind of tasks, is easily handled during upgrades (unlike the admin node), and doesn't add more responsibility onto the admin node - because we want to, at some point, stop wasting a hardware node.

But, yes, I agree with Rob that doing this it's it's own topic, something that should be looked at in the context of all the "once per cluster" things we do (etcd discovery setup, flannel setup, probably more) and designed, rather than just added as part of a random PR to cover just 1 use case.

salt/addons/init.sls
inercia: I think we started with `-` (see the most important ids) and then there have been many `_` introduced in the last times. I agree that Salt standard library is more fond of underscores, so we should agree on a convention for ids. Otherwise someone could start using spaces...

salt/addons/init.sls
SlickNik: Reading https://github.com/kubernetes/kubernetes/issues/44165, it seems that this is an issue iff the kubernetes manifests that are being run by the two "kubectl apply" calls are different. In that case, it's possible that the 'later one' isn't actually applied. In this case however, the two manifests are both the same so I think we should be okay.

Regardless, I agree with Kiall that it would be useful to have something that runs "once per cluster". I also agree that we should look at implementing something like that separately since it's out of the scope of this Pull Request.

salt/addons/init.sls
SlickNik: Is it worth converting all of the existing states from '-' to '_' (or vice versa) just for convention sake? If we want to standardize on a convention going forward, I think we should put this in the README for the repo and/or the project, and then we can enforce it once folks are on the same page. But until then, please let's continue to make progress on PRs (especially for states like deploy_dns / apply_dns which already exist today).

salt/addons/init.sls
kiall: It's definitely an issue, multi master deploys were failing due to kube-dns deployment (after updating them to use apply instead of create), given CI is now multi master, I suspect landing this as is will cause most other changes to hit the issue and fail CI :(

kiall: (I probably wasn't clear, running `kubectl apply .. || kubectl apply ..` is what I suggest as a fix for this PR, as it's perfectly safe to do so in the context of our salt states. I was not suggesting the whole "once per cluster" stuff, which should be added to the backlog and done separately as time allows in the future)

SlickNik: Ah, I see. Thanks for the clarification! @rbwsam can we make the change that kiall suggests above so that we continue to work in the multi-master scenario? (i.e. kubectl apply ..  || kubectl apply)

at1012:
I checked out this patch and brought up a caasp dev with addons: tiller set to true, and tiller deployement came up fine.

SlickNik:
Looks good to me as well. Thanks Sam!

SlickNik:
@inercia Can you take a look and approve when you get a chance? Looking to get this in soon, and it looks like Sam has addressed all the comments that you had. Thanks! :)

salt/kube-apiserver/init.sls
```diff
@@ -112,9 +112,3 @@ kube-apiserver:
       - file:     kube-apiserver
       - sls:      ca-cert
       - {{ pillar['ssl']['kube_apiserver_crt'] }}
-  # wait for the API server (see k8s issue #47739) so it can be used
-  http.wait_for_successful_query:
```
kiall: Oh, why did we remove this? It's better to keep this with kube-apiserver, so other things which need it (e.g. the CNI flannel change) can reuse it.

salt/addons/addons/tiller.yaml.jinja
```diff
@@ -0,0 +1,76 @@
...
+  loadBalancer: {}
+
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: tiller
+  namespace: kube-system
+  labels:
+    kubernetes.io/cluster-service: "true"
```
inercia: Wouldn't it make more sense to declare `ServiceAccount`, `Service` and then the `Deployment`? Maybe it is not really important, but if the YAML file is evaluated in strict order, I think you would first need the `ServiceAccount` and then the `Service` and last the `Deployment`...

salt/kube-apiserver/init.sls
inercia: ... and you could remove [this loop](https://github.com/kubic-project/salt/pull/193/files#diff-3d900be66a547ca3549316625a8e2dd5R24).

salt/addons/addons/tiller.yaml.jinja
SlickNik: The order here doesn't really matter since the labels are really what is used by the constructs to identify each other.

salt/kube-apiserver/init.sls
SlickNik: @kiall @inercia The http.wait_for_successful_query state doesn't work correctly ever since we added haproxy in front of the kube-api servers. We were running into an issues where http.wait_for_successful_query would pass (i.e. the kube-api server would report as ready when it was hit directly on the unauthenticated localhost port on 8080) but going through the local haproxy would error out (presumably because no backends were available yet -- not having passed the healthy interval check). 

To fix this, we updated the http.wait_for_successful_query call with that curl loop that actually makes an authenticated call through haproxy, which ensures that kube-api is actually up before we make issue any kubectl commands. 

It might be a good idea to add the curl check to this state though -- so that we don't break if another api-dependent state is introduced between this and the place where the check exists today.

inercia: Why don't we run `wait_for_successful_query` on the _haproxy_ port? I'm sure haproxy will return an error if the backend (the API server) is not ready yet. I guess that could be reliable enough...

salt/addons/addons/tiller.yaml.jinja
inercia: Ok, I see.

kiall:
Comments make sense, as a nit.. I think the loop to check health should say in the kube-apiserver role, but we can deal with that later.

salt/kube-apiserver/init.sls
kiall: `wait_for_successful_query` doesn't accept params for the client certs, so, as long as the authenticated port is used, it will fail.

Since there is a delay (on a per-backend basis) for haproxy to healthcheck nodes, the only reliable check is to check the authenticated port.

inercia: `wait_for_successful_query` seems to support certificates: [this function](https://github.com/saltstack/salt/blob/d1f80bab3993059060ca51263e9d048e3a618051/salt/modules/http.py#L41) will eventually call [salt.utils.http.query](https://github.com/saltstack/salt/blob/develop/salt/utils/http.py#L98), a function that can accept a [CA](https://github.com/saltstack/salt/blob/develop/salt/utils/http.py#L130) and a [certificate](https://github.com/saltstack/salt/blob/develop/salt/utils/http.py#L132) (looking at [these lines](https://github.com/saltstack/salt/blob/develop/salt/utils/http.py#L442), `cert` can be a tuple with `(cert, key)`).

I would give it a try, but if this is too much hassle now I would leave it for a subsequent PR...

robdaemon:
Okay, we're 35 comments deep on a fairly small change. @rbwsam - if you rebase, I'll merge.

SlickNik:
@rbwsam is working on the rebase -- there have been quite a few changes, so he's working on doing some more testing.

robdaemon:
I've asked @rbwsam to open a new PR that doesn't have any merge commits and to use rebase from now on instead of merge. We'll get the advantage that it's automatically squashed there too. I'll review in the morning (unless @kiall beats me to it.)

PR #196, opened 2017-08-17
Add kube-dns service account

at1012:


salt/addons/init.sls
```diff
@@ -15,6 +15,11 @@ include:
     - template:    jinja
 
 {% if salt['pillar.get']('addons:dns', 'false').lower() == 'true' %}
+/etc/kubernetes/addons/kubedns-sa.yaml:
+  file.managed:
+    - source:     salt://addons/addons/kubedns-sa.yaml.jinja
+    - template:   jinja
```
inercia: Nothing is replaced in this file, so no Jinja templating is needed,

salt/addons/deploy_addons.sh
```diff
@@ -26,9 +26,19 @@ fi
 
 echo "Deploying DNS on Kubernetes"
 
+KUBEDNS_SA=`eval "$KUBECTL_PATH $NS_ARGS $SRV_ARGS get sa | grep kube-dns | cat"`
```
inercia: I think we should try to move all this stuff from the `deploy_addons.sh` script to some state file, and just invoke `kubectl` with `cmd.run`.

salt/addons/init.sls
robdaemon: It's harmless.

salt/addons/deploy_addons.sh
robdaemon: That's outside of the scope of this change. A replacement of this script is in progress.

PR #197, opened 2017-08-17
Properly wait for a HTTP endpoint

inercia:


salt/addons/init.sls
```diff
@@ -30,19 +30,12 @@ include:
     - source:      salt://addons/addons/kubedns-svc.yaml.jinja
     - template:    jinja
 
-kube_apiserver_ready:
+kube-apiserver-ready:
   # We need to wait for Kube API server to actually start, see k8s issue #47739
   # TODO: Salt doesn't seem to have a retry mechanism in the version were using,
   # so I'm doing a horrible hack right now.
```
ereslibre: This comment can also go away.

salt/addons/init.sls
```diff
@@ -30,21 +30,11 @@ include:
...
-        until curl --insecure --silent --fail -o /dev/null http://127.0.0.1:8080/healthz ; do
-            [ $ELAPSED -gt 300 ] && exit 1
-            sleep 1 && ELAPSED=$(( $ELAPSED + 1 ))
-        done
-        echo changed="no"
-    - stateful: True
-
 deploy_addons.sh:
+  # We need to wait for Kube API server to actually start, see k8s issue #47739
+  http.wait_for_successful_query:
```
ereslibre: Do we know for sure `http.wait_for_successful_query` will be executed before `cmd.script`? This is a dictionary and I'm not sure if salt will honor order here.

grahamhayes: It will be - with no explicit "requires" salt will work top to bottom in a file.

ereslibre: @grahamhayes Yes, top level definitions like in `deploy_addons.sh` will be executed before `/etc/kubernetes/addons/kubedns-svc.yaml`, just wanted to make sure this also works like this when several actions are defined in one state like in this one.

If that's the case, +1 from me.

PR #198, opened 2017-08-17
Do not generate an empty --proxy line in curlrc

inercia:
It is ok to generate empty values in `/etc/sysconfig/proxy`, but `curl` does not like having an empty `--proxy` line and generates some warnings...

### DIFF
--- a/salt/proxy/init.sls
+++ b/salt/proxy/init.sls
@@ -1,18 +1,25 @@
 {% if salt['pillar.get']('proxy:systemwide', '').lower() == 'true' %}
 
+{% set proxy_http  = salt['pillar.get']('proxy:http', '') %}
+{% set proxy_https = salt['pillar.get']('proxy:https', '') %}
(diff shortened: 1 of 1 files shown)


salt/proxy/init.sls
```diff
@@ -1,18 +1,25 @@
...
 
+{% set proxy_http  = salt['pillar.get']('proxy:http', '') %}
+{% set proxy_https = salt['pillar.get']('proxy:https', '') %}
+{% set no_proxy    = salt['pillar.get']('proxy:no_proxy', '') %}
+
 /etc/sysconfig/proxy:
   file.managed:
     - makedirs: True
     - contents: |
         PROXY_ENABLED="yes"
```
inercia: Not sure we should have this here if `proxy_http|length == 0`...

kiall: I think it's harmless, as we've checkd for the proxy:systemwide pillar outside, and that means we have some sort of proxy to be enabled... If thats enabled, and no proxy has been supplied, it's a end user configuration error anyway IMO

PR #199, opened 2017-08-17
Disable Salt's Job Cache

kiall:
Salt's job cache is buggy, causing random failures to lookup mine data, which
in turn causes our deployments to fail.

Fixes bsc#1054256

### DIFF
--- a/config/master.d/master.conf
+++ b/config/master.d/master.conf
@@ -1,3 +1,7 @@
+# Note to future editors of this file, salt's job cache
+# is buggy, usually showing itself at scale rather than
+# in small deployments. Test at scale before choosing to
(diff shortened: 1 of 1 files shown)


ereslibre:
Should this change also include the Velum logic for cleaning up old jobs? (include as depend on other Velum PR)

kiall:
This change can't, as that logic needs to go to the velum repo :)

However, I'd like to merge this anyway, prior to getting the cleanup implemented, so we can continue to validate larger clusters without hacking this in :)

ereslibre:
Yup, I meant "include" as a reference to other PR, but fine by me.

LGTM.

PR #200, opened 2017-08-17
Wait for the API server after starting the service

inercia:
Wait for the API server port after starting the service. So we can just `require: kube-apiserver` for invoking  `kubectl` safely...

### DIFF
--- a/salt/addons/init.sls
+++ b/salt/addons/init.sls
@@ -31,14 +31,10 @@ include:
     - template:    jinja
 
 deploy_addons.sh:
-  # We need to wait for Kube API server to actually start, see k8s issue #47739
-  http.wait_for_successful_query:
-    - name:        'http://127.0.0.1:8080/healthz'
-    - status:      200
   cmd.script:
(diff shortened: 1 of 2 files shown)


salt/addons/init.sls
```diff
@@ -35,10 +35,11 @@ deploy_addons.sh:
   http.wait_for_successful_query:
```
inercia: Maybe we should move this `wait_for_successful_query` to `kube-apiserver/init.sls:kube-apiserver`, so anyone needing the API server could just `require: service: kube-apiserver`.

kiall: That's not a bad idea... these were all originally in kubernetes-master, and the check was moved as part of the split. Might be best to put it with kube-apiserver.

However, waiting on the service wouldn't be enough... You can wait on the whole state though with `require: [kube-apiserver]' - which will wait on the service, cmd.run, etc.

inercia: That's right

PR #202, opened 2017-08-18
Use haproxy to load balance Kube API requests

kiall:
Now that we can have multiple masters, we need a way for the various
services and end-users to be load balanced over the set of kube-api
servers.

We install haproxy on each node, inside a docker container, configured
to load balance requests over all the cluster masters. This haproxy
is configured to listen on 0.0.0.0 on the masters, and 127.0.0.1 on
the workers.

This is to allow the minions to simply "talk" to 127.0.0.0, and be
routed to an active kube-api server.

### DIFF
--- a/salt/haproxy/haproxy.cfg.jinja
+++ b/salt/haproxy/haproxy.cfg.jinja
@@ -1,0 +1,30 @@
+{%- if "kube-master" in salt['grains.get']('roles', []) -%}
+{%- set bind_ip = "0.0.0.0" -%}
+{%- else -%}
+{%- set bind_ip = "127.0.0.1" -%}
+{%- endif -%}
+global
+        log /dev/log    local0
+        log /dev/log    local1 notice
+        daemon
+
+defaults
+        log     global
+        mode    tcp
+        option  tcplog
+        option  dontlognull
+        timeout connect 5000
+        timeout client 50000
+        timeout server 50000
+
+# Listen on the standard Kube-API Public port, 6443 by default, and proxy to the masters on
+# the Kube-API internal port, 6444 by default.
+listen kubernetes-master
+        bind {{ bind_ip }}:{{ pillar['api']['ssl_port'] }}
+        mode tcp
+        default-server inter 10s fall 3
+        balance roundrobin
+
+{%- for minion_id, _ in salt['mine.get']('roles:kube-master', 'network.ip_addrs', 'grain').items() %}
+        server master-{{ minion_id }} {{ minion_id }}.{{ pillar['internal_infra_domain'] }}:{{ pillar['api']['int_ssl_port'] }} check
+{% endfor -%}
--- a/salt/haproxy/init.sls
+++ b/salt/haproxy/init.sls
@@ -1,0 +1,47 @@
+/etc/haproxy/haproxy.cfg:
+  file.managed:
+    - source: salt://haproxy/haproxy.cfg.jinja
+    - template: jinja
+    - user: root
+    - group: root
+    - mode: 644
+    - makedirs: True
+    - dir_mode: 755
+
+haproxy:
+  file.managed:
+    - name: /etc/kubernetes/manifests/haproxy.yaml
+    - source: salt://haproxy/haproxy.manifest
+    - template: jinja
+    - user: root
+    - group: root
+    - mode: 644
+    - makedirs: True
+    - dir_mode: 755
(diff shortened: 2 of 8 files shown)


kiall:
Most of this code was grabbed from an older change in the "ha" branch, and reworked as the diff was too large to just cherry-pick or rebase.

https://github.com/kubic-project/salt/commit/35da1bedf1116a6544693bdd5a24ce4a5eceb7ee

salt/haproxy/init.sls
```diff
@@ -0,0 +1,48 @@
...
+    - template: jinja
+    - user: root
+    - group: root
+    - mode: 644
+    - makedirs: True
+    - dir_mode: 755
+{% if "kube-master" in salt['grains.get']('roles', []) %}
+  iptables.append:
+{% else %}
+  iptables.delete:
```
robdaemon: Why delete this? It shouldn't matter what node gets the request, they can all forward to the API server.

kiall: While it's probably harmless to do so, we're not listening on 6443 on the workers, which would leave an open and unbound port - which I don't like the idea of :)

We could just listen on 6443 everywhere, but I think it's easier to ask people to aim the public FQDNs that compose the core service APIs to the reasonably fixed pool or masters, rather than the hopefully fluid pool of workers.

I don't have a strong opinion either way - other than -  we should only be exposing ports where we actually have a service listening on that machine

robdaemon: Ahhh. The initial implementation I did of this on the ha branch listened on 6443 on all nodes, and forwarded to the open API servers. The local kubelets / kube-proxy all talked to localhost:6443.

In the case of any software load balancer deployed inside a cluster, and with how kube-proxy works, you are going to be listening on nodes that do not host the service and performing a reverse proxy to the appropriate place, so....

kiall: Preface - I think you just spotted an issue, I'll need to double check in the morning.

Oh, let me rephrase slightly. We do listen on 6443 on all machines - however on masters, it's 0.0.0.0:6443, and on workers it's 127.0.0.1:6443.

The actual kube-api process listens on 0.0.0.0:6444 on the masters, and is not bound at all on the workers.

The issue I think I have is - I'm only adding 1 allow rule, when I need two (and eventually, we get network segregation, and the second allow rule is on the "internal" network)

kiall: Actually, I add the second rule in the kube-api role, nevermind :)

salt/haproxy/init.sls
```diff
@@ -0,0 +1,48 @@
+---
```
ereslibre: This line can be removed.

salt/haproxy/init.sls
```diff
@@ -0,0 +1,48 @@
...
+      - {{ pillar['api']['ssl_port'] }}
+    - proto:      tcp
+
+# Send a HUP to haproxy when the config changes
+# TODO: There should be a better way to handle this, but currently, there is not. See
+# kubernetes/kubernetes#24957
+haproxy_restart:
+  cmd.run:
+    - name: |-
+        haproxy_id=$(docker ps -a | grep -E "k8s_haproxy.*\.infra\.caasp\.local_kube-system_" | awk '{print $1}')
```
ereslibre: Minor: we only care about running containers so we should use `docker ps` instead of `docker ps -a`.

salt/haproxy/haproxy.cfg.jinja
```diff
@@ -0,0 +1,33 @@
...
+        mode tcp
+        default-server inter 10s fall 3
+        balance roundrobin
+
+{%- for minion_id, addrlist in salt['mine.get']('roles:kube-master', 'network.ip_addrs', 'grain').items() %}
+{% if addrlist is string -%}
+{% set addr = addrlist -%}
+{% else -%}
+{% set addr = addrlist|first -%}
+{% endif -%}
```
ereslibre: What purpose has this whole block? `addr` is not used after the declaration.

kiall: Ah, a leftover from the original commit to another branch a few months back. I'll remove it.

salt/haproxy/init.sls
kiall: Good catch.

grahamhayes:
Approving as the issue is with the devenv tooling (using an IP in place of a FQDN)

PR #203, opened 2017-08-18
Ensure k8s_etcd.get_cluster_size works for multi-master

kiall:
If we had enough masters to form a etcd cluster, we would end up returning
"None" from this method, preventing the cluster formation.


robdaemon:
Looks good, will need to run it to be sure.

PR #204, opened 2017-08-21
Use apply instead of create for addons

kiall:
kubectl apply is idempotent, while kubectl create is not. With multi-master
now enabled, if two masters execute this script at once, one of them is likely
to fail given the check+set race within this script - Switching to apply
removes this C+S race.


kiall:
So - I still got a failure with this on 1 my tests. I'll hold off merging for the moment.

salt/addons/deploy_addons.sh
```diff
@@ -14,11 +14,17 @@ abort() { echo "FATAL: $@" ; exit 1 ; }
...
+# + set race condition. If it fails, we try it again, if it fails again, we will
+# fail the entire execution of this script. This works around the "create or update"
+# check+set race.
+
 echo "Creating kube-system namespace..."
 # use kubectl to create kube-system namespace
 NAMESPACE=`eval "$KUBECTL_PATH $SRV_ARGS get namespaces | grep kube-system | cat"`
 if [ ! "$NAMESPACE" ]; then
-	$KUBECTL_PATH $SRV_ARGS create -f "$DIR/namespace.yaml"
+	$KUBECTL_PATH $SRV_ARGS apply -f "$DIR/namespace.yaml" || $KUBECTL_PATH $SRV_ARGS apply -f "$DIR/namespace.yaml"
```
inercia: Maybe you should refactor this to a common function:

```bash
kubectl_apply() {
    $KUBECTL_PATH $SRV_ARGS apply -f $1" || $KUBECTL_PATH $SRV_ARGS apply -f $1
}
```

and you could even loop until the `kubectl` succeeds.

kiall: Sorry, I hit merge and then this comment showed up as the page reloaded...

PR #205, opened 2017-08-22
Revert "Kube-API: Set storage-backend to etcd2"

mjura:
Etcd storage backend etcd2 is deprecated.
We would like to ship CaaSP v2.0 with default storage-backend etcd3, which doesn't need to migration in future.

### DIFF
--- a/salt/kube-apiserver/apiserver.jinja
+++ b/salt/kube-apiserver/apiserver.jinja
@@ -36,5 +36,4 @@ KUBE_API_ARGS="--advertise-address={{ grains['ip4_interfaces']['eth0'][0] }} \
                --tls-cert-file={{ pillar['ssl']['crt_file'] }} \
                --tls-private-key-file={{ pillar['ssl']['key_file'] }} \
                {{ salt['pillar.get']('components:apiserver:args', '') }} \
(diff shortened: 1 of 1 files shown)


mjura:
> Without explicitly saying we are using the v2 backend kube-apiserver throws errors, and does not fully work.

@grahamhayes: what errors? with what version of kubernetes packages? If we will enabled etcd2 as storage-backend in CaaSP2 then we will have big problem with migration later, because migration from etcd2 to etcd3 requires downtime for whole cluster.

grahamhayes:
@mjura we already have that problem, v1 used the v2 datastore, and we have to upgrade current users.
We have not written the tooling to move users from v2->v3 yet, so we cannot just enable v3 - current clusters would fail.

It was with the current version of k8s in the repos (1.7 I think). kube-apiserver was trying to connect to the grpc v3 endpoint, which was not configured (I think, I did not do too much debugging).

We should remove this line in the PR that adds the v2 -> v3 upgrade tooling, and ensure all components are actually talking to the v3 datastore

mjura:
Ok, please review following change
https://github.com/kubic-project/salt/pull/206

mjura:
Ok, please review following change
https://github.com/kubic-project/salt/pull/206

PR #209, opened 2017-09-05
WIP: Use reactors to notify Velum of important events

ereslibre:
**Work in progress**, please do not review yet.

- [x] Use SSL/TLS for callbacks and pillar
- [x] Use API auth for callbacks and pillar (basic auth?)
- [x] Implement pillar caching locally, so if Velum does not respond or is temporarily unavailable we still render the latest pillar we had.

Depends on:
* https://github.com/kubic-project/velum/pull/299
* https://github.com/kubic-project/caasp-container-manifests/pull/100
* https://github.com/kubic-project/container-images/pull/21

### DIFF
--- a/reactor/auth-notifier.sls
+++ b/reactor/auth-notifier.sls
@@ -1,0 +1,15 @@
+notify:
+  runner.http.query:
+    - url: https://localhost/internal-api/v1/auth
+    - method: POST
+    - header_list:
+      - 'Content-Type: application/json'
+    - ca_bundle: /etc/pki/ca.crt
+    - username: {{ salt['environ.get']('VELUM_INTERNAL_API_USERNAME') }}
+    - password: {{ salt['environ.get']('VELUM_INTERNAL_API_PASSWORD') }}
+    - data: |
+        {
+          "event_data": {
+            "minion_id": {{ data['id']|json }}
+          }
+        }
--- a/config/master.d/50-master.conf
+++ b/config/master.d/50-master.conf
@@ -5,7 +5,6 @@
 user: root
 auto_accept: False
 interface: 0.0.0.0
-event_return: mysql
 presence_events: True
 timeout: 20
 
--- a/config/master.d/50-reactor.conf
+++ b/config/master.d/50-reactor.conf
@@ -2,7 +2,18 @@ reactor:
   - 'salt/presence/change':
     - /usr/share/salt/kubernetes/reactor/presence.sls
     - /usr/share/salt/kubernetes/reactor/etc-hosts.sls
+    - /usr/share/salt/kubernetes/reactor/presence-notifier.sls
   - 'salt/beacon/*/default_network_interface_settings/*':
     - /usr/share/salt/kubernetes/reactor/etc-hosts.sls
   - 'salt/minion/*/start':
     - /usr/share/salt/kubernetes/reactor/etc-hosts.sls
+  - 'salt/run/*/new':
+    - /usr/share/salt/kubernetes/reactor/orchestration-trigger-notifier.sls
+  - 'salt/run/*/ret':
+    - /usr/share/salt/kubernetes/reactor/orchestration-result-notifier.sls
+  - 'salt/job/*/ret/*':
+    - /usr/share/salt/kubernetes/reactor/highstate-notifier.sls
+  - 'salt/auth':
+    - /usr/share/salt/kubernetes/reactor/auth-notifier.sls
+  - 'minion_start':
+    - /usr/share/salt/kubernetes/reactor/minion-start-notifier.sls
--- a/config/master.d/50-returner.conf
+++ b/config/master.d/50-returner.conf
@@ -1,8 +1,0 @@
-mysql:
-  # salt does not support specifying the UNIX socket location here - as a workaround,
-  # use the MYSQL_UNIX_PORT environment variable used by libmysqlclient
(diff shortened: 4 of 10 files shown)


inercia:
From my understanding of this PR, the reactor is basically used for filtering the events we are interested in. But those events are transformed in HTTPS calls that, at the end, will be stored in the database after going through Velum, all the the Ruby stack and so. Wouldn't it be easier just to [write our own returner](https://docs.saltstack.com/en/latest/ref/returners/#writing-a-returner) that filters the events we are interested in and saves them to the database?

ereslibre:
> From my understanding of this PR, the reactor is basically used for filtering the events we are interested in. But those events are transformed in HTTPS calls that, at the end, will be stored in the database after going through Velum, all the the Ruby stack and so. Wouldn't it be easier just to write our own returner that filters the events we are interested in and saves them to the database?

I would like to discuss this in a broader approach of the whole architectural point of view of the system:

* A returner writing to mysql doesn't separate concerns (salt will still need to have `mysql` access directly)
* A returner writing to mysql is very reduced on what it can do, only modify the database. It cannot trigger further actions or run logic that we might need
* The pillar will be fetched from Velum (we already need basic auth here at least) -> we would have auth for the internal-api of Velum for fetching the pillar and the mysql credentials
* Important changes on the mysql schema in Velum will require us to modify the returner to be in sync while this schema design is internal to Velum
* A returner that only saves what we want misses to save the rest of the events -> a reactor will notify us of what we are interested in, while saving all events

In general, the whole idea of two different components writing/reading with internal representations of the other in a third component (mysql in this case) is something I would like to avoid at all costs. Two different components should talk to each other with their designed API to do so instead, IMHO. This gives us a lot more power, concern separation, better maintanability in the long run.

inercia:
> * A returner writing to mysql doesn't separate concerns (salt will still need to have mysql access directly)

It would separate things quite well in my view. I don't quite understand the point about having mysql access.

> * A returner writing to mysql is very reduced on what it can do, only modify the database. It cannot trigger further actions or run logic that we might need.

This returner could just save things to the database, and actions can be done from the Velum side of things by querying the database.

> * The pillar will be fetched from Velum internal API (we already need basic auth here at least) -> we would have auth for the internal-api of Velum for fetching the pillar and the mysql credentials
Important changes on the mysql schema in Velum will require us to modify the returner to be in sync while this schema design is internal to Velum.

I don't think the schema changes so often, and changes are usually not so big. We are not talking about a database with 50 tables. Keeping in sync both things doesn't seem such a big deal to me...

> * A returner that only saves what we want misses to save the rest of the events -> a reactor will notify us of what we are interested in, while saving all events

You can use multiple returners if you want. We would use our own for the interesting events, but we could use another one for saving all the other events.

As you said, this is a broader architectural decision. I think this path will lead to much more complexity (ie, authentication will not be accepted to be just "basic auth", and we will need more certificates and more security stuff), while adding more load on Velum and the (slow) Ruby stack...

reactor/auth-notifier.sls
```diff
@@ -0,0 +1,8 @@
+notify:
+  runner.http.query:
+    - url: https://localhost/internal-api/v1/auth
+    - method: POST
+    - header_list:
+      - 'Content-Type: application/json'
+    - ca_bundle: /etc/pki/ca.crt
+    - data: '{ "event_data": { "minion_id": "{{ data['id'] }}" } }'
```
robdaemon: Building JSON this way is unsafe for many reasons - escaping, etc.

This is also the sort of behavior that leads to SQL injection if you're using string concatenation / interpolation to build SQL.

ereslibre: This information comes from salt (minion_id, which is controlled by us), also, this is JSON being sent to the API, which uses ActiveRecord ORM. Same threats as we have today, but with API sanitizing in between (today python reads this whole `data` structure and logs absolutely everything to mysql using raw SQL commands -- in an environment way unsafer than ActiveRecord [meant to properly sanitize queries because of always being open to the world]).

ereslibre: Example: https://github.com/saltstack/salt/blob/develop/salt/returners/mysql.py#L308-L314

ereslibre: Also, you'll see on other `-notifier.sls` files that we do `|json` filtering. In this case this wasn't really necessary as the `minion_id` should be safe. In any case we might want to add yet an extra safety check here so a specially crafted `minion_id` cannot do any harm (but as I said, we'd really need to understand a real use case in which this can be exploited, I have my doubts it would even reach the database).

ereslibre:
This PR is being splitted in different PR's. I will close this one when the whole functionality has been correctly splitted, so we can discuss smaller parts of it.

reactor/auth-notifier.sls
robdaemon: I'm looking at that string and there is no sanitization going on. You cannot build JSON this way. You need to use the `|json` stuff *everywhere* and build proper hashes / arrays that are then serialized to JSON properly.

ereslibre: Will update that when splitting the PR.

ereslibre:
[Reactor callbacks](https://trello.com/c/p7L7JVQo/451-reactor-callbacks)

MalloZup:
we can close here :japan:

MalloZup:
closing as discussed

PR #211, opened 2017-09-06
Ensure systemd is reloaded after units are changed

kiall:
Ensure systemd is reloaded as soon as a unit is changed, rather than relying
on a task later within the orchestration to execute.

Fixes bsc#1057641

### DIFF
--- a/salt/docker/init.sls
+++ b/salt/docker/init.sls
@@ -14,9 +14,10 @@ include:
         Environment="HTTP_PROXY={{ salt['pillar.get']('proxy:http', '') }}"
         Environment="HTTPS_PROXY={{ salt['pillar.get']('proxy:https', '') }}"
         Environment="NO_PROXY={{ salt['pillar.get']('proxy:no_proxy', '') }}"
-  cmd.run:
-    - name: systemctl daemon-reload
-    - stateful: True
+  module.run:
+    - name: service.systemctl_reload
+    - onchanges:
+      - file: /etc/systemd/system/docker.service.d/proxy.conf
 
 #######################
 # docker daemon
--- a/salt/etcd/init.sls
+++ b/salt/etcd/init.sls
@@ -59,6 +59,10 @@ etcd:
   file.managed:
     - source: salt://etcd/etcd.conf
     - makedirs: True
+  module.run:
+    - name: service.systemctl_reload
+    - onchanges:
+      - file: /etc/systemd/system/etcd.service.d/etcd.conf
 
 # note: this will be used to run etcdctl client command
 /etc/sysconfig/etcdctl:
--- a/salt/transactional-update/init.sls
+++ b/salt/transactional-update/init.sls
@@ -7,6 +7,10 @@
     - user: root
     - group: root
     - mode: 644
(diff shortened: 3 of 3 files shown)


salt/docker/init.sls
```diff
@@ -14,9 +14,10 @@ include:
         Environment="HTTP_PROXY={{ salt['pillar.get']('proxy:http', '') }}"
         Environment="HTTPS_PROXY={{ salt['pillar.get']('proxy:https', '') }}"
         Environment="NO_PROXY={{ salt['pillar.get']('proxy:no_proxy', '') }}"
-  cmd.run:
-    - name: systemctl daemon-reload
-    - stateful: True
+  module.run:
+    - name: service.systemctl_reload
```
inercia: Does `sytemctl reload` restart the `docker` daemon on a changes on this file? Looking at the man page I'm not sure: _"Reload the systemd manager configuration. This will rerun all generators (see systemd.generator(7)), reload all unit files, and recreate the entire dependency tree. While the daemon is being reloaded, all sockets systemd listens on behalf of user configuration will stay accessible."_

salt/etcd-proxy/init.sls
```diff
@@ -5,6 +5,10 @@ include:
   file.managed:
     - source: salt://etcd-proxy/etcd.conf
     - makedirs: True
+  module.run:
+    - name: service.systemctl_reload
+    - onchanges:
+      - file: /etc/systemd/system/etcd.service.d/etcd.conf
```
inercia: I think we should move this code to `/salt/etcd/init.sls`. This `init.sls` inherits from there and just overwrites some details (like the config file path).

salt/docker/init.sls
kiall: No, it won't restart the daemon - all we're doing in this PR is ensuring systemd is reloaded, if we're failing to reload/restart docker on changes to systemd unit files, that's a separate bug we should file and fix!

salt/etcd-proxy/init.sls
kiall: Agree, however, that's unrelated to ensuring systemd is reloaded when we change the unit files... There was another (abandoned) PR to do that.. I'll revive it.

kiall:
> Btw, Is this related to bsc#1057362 ?

Kinda ... a comment in that bug was the trigger to go fix this, however, this is not a fix for that bug.

salt/etcd-proxy/init.sls
kiall: PR to combine the roles: https://github.com/kubic-project/salt/pull/215

kiall:
Updated to include a bug number in the commit message @ereslibre @inercia could you review / +A again? Thanks!

salt/docker/init.sls
inercia: By the way, have you actually check this `service.systemctl_reload` works fine? We did the `cmd.run: systemctl daemon-reload` because it was not working fine in the past, but maybe it has been fixed...

kiall:
Yes, I checked these worked as expected.

PR #212, opened 2017-09-06
Include both v2 and v3 flags in etcdctl vars

kiall:


mjura:
We should also consider how to setup variable `ETCDCTL_API=<version>, 2 or 3`

kiall:
> We should also consider how to setup variable ETCDCTL_API=<version>, 2 or 3

Yea, I actually started that, and then realised it might cause you pain in the datastore migration PR, so left it for now

PR #201, opened 2017-08-17
Use CNI for the network fabric

inercia:
TODO:
- [x] Create a docker image based on SLE with flannel
- [x] Add the docker image to the CaaSP image
- [x] Remove `flanneld` from the CI tests (kubic-project/automation#88)
- [x] Update the manifests and use our docker image
- [x] Ensure the upgrade from non-CNI -> CNI is handled gracefully, ideally without downtime, and certainly without addressing changes.

Dependencies

* [Velum migration](kubic-project/velum#312)

PR #213, opened 2017-09-07
DO NOT MERGE Test what happens without this entry

kiall:


kiall:
@robdaemon Q - What is the host entry used for? I'm wondering if it's necessary?

kiall:
So - I think this passed because the external DNS names used in caasp-kvm are real, and actually resolve. If the name wasn't real, it likely would have failed - I think.

However, putting on masters in /etc/hosts will likely not work well for multi-master HA .. Maybe we need to treat Dex like we do kube-api, and use haproxy + the 127.0.0.1 hosts entry?

inercia:
I've [removed the same entryi n the CNI PR](https://github.com/kubic-project/salt/pull/201/files#diff-09c149f34d38bfe9b8406e474e155f54) and confirmane tests are passing. So I guess nobody really tries to resolve that name internally. Otherwise, I think things would be really broken, as the _external fqdn_ does not necessarily make sense in the internal network (it was introduced for being used in the `kubeconfig` provided to users and in the certificates).

robdaemon:
This entry is required so that Kubernetes API server can reach the OIDC issuer at all times. It cannot be removed. That's why I added it.

robdaemon:
This entry is required so that Kubernetes API server can reach the OIDC issuer at all times. It cannot be removed. That's why I added it.

kiall:
@robdaemon so - as is - I believe that if the first master fails, dex will be unreachable. We'll need to at the least move this over to use the 127.0.0.1 + haproxy pattern, and possibly use the internal names rather than external endpoint - we should avoid "hijacking" that name if at all possible.

robdaemon:
You can't use the internal name as the issuer must match what the client uses. 127.0.0.1 is going to cause certificate issues. We may be able to do something with haproxy though.

PR #220, opened 2017-09-08
Set a default external fqdn in the pillar

inercia:
This is set only from Velum, so we should not fail when we orchestrate things with Salt

### DIFF
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -25,6 +25,9 @@ api:
   # etcd storage backend version for cluster
   etcd_version:   'etcd3'
 
+  server:
+    # should be set from the UI
+    external_fqdn: ''
 
 # DNS service IP and some other stuff (must be inside the 'services_cidr')
 dns:
@@ -72,7 +75,7 @@ ssl:
 
   kube_controller_mgr_key: '/etc/pki/kube-controller-mgr.key'
   kube_controller_mgr_crt: '/etc/pki/kube-controller-mgr.crt'
-  
+
   kubelet_key: '/etc/pki/kubelet.key'
   kubelet_crt: '/etc/pki/kubelet.crt'
 


salt/etc-hosts/hosts.jinja
```diff
@@ -20,7 +20,7 @@
 	{%- else %}
 		{%- set ip = addrlist|first %}
 	{%- endif %}
-{{ ip }} {{ master_id }} {{ master_id }}.{{ pillar['internal_infra_domain'] }} {{ pillar['api']['server']['external_fqdn'] }}
+{{ ip }} {{ master_id }} {{ master_id }}.{{ pillar['internal_infra_domain'] }} {{ salt['pillar.get']('api:server:external_fqdn', '') }}
```
kiall: The external FQDN is a required parameter, I can't see a circumstance where it wouldn't be set? I'm guessing we have a bug somewhere else that causes this to be unset? This will just mask that bug.

ereslibre: Now the external fqdn field is the last step of the setup, what means that we can:

1. Set the dashboard value
2. Accept minions -> those trigger the `etc-hosts` orchestration
3. Orchestration fails because the setup was not yet finished (`api:server:external` fqdn was not yet set because we didn't finish the setup).

So I'd say this is not really a "bug" but a possibility with the current workflow, and I think it's fine.

ereslibre: But this should only run if roles have been set, so I'm also confused now by this change...

inercia: > The external FQDN is a required parameter, I can't see a circumstance where it wouldn't be set? I'm guessing we have a bug somewhere else that causes this to be unset? This will just mask that bug.

Why is the external FQDN _always_ required? I've never set it and it's always worked fine for me.  It will be required if the user accesses the API server at a external name (so that the kubeconfig/certificates contain that name), but I've always used the internal names fine (because I use the API server from an internal network).

kiall: How did you manage that? It's a required field, whatever is allowing you to proceed without that is where the bug is:

image

inercia: @kiall running `salt state.orchestrate orch.kubernetes`. You are forcing the use of the UI, when it is not necessary.

inercia: Ok, let's do something easier: we can set a default value in `pillar/params.sls` so it can be overwritten from Velum but a `salt state.orchestrate` will not fail either...

kiall: Yea, we should have a default in there as a reference - as we should for every pillar value we read. That said, you're still deploying CaaSP "wrong" if you don't set the pillar values :)

robdaemon: The fact that we added an additional step at the end is a bad UX while leaving the initial screen with the proxy. Yes, this is unrelated to this change, but I needed to say it somewhere. :)

mjura:
@inercia: please rebase

inercia:
Rebased... please review again @ereslibre @mjura @kiall

ereslibre:
@robdaemon Ouch, since we didn't yet branch release-2.0 I was waiting to merge this afterwards, since it's not strictly fixing any bugzilla registered bug and we were already in code freeze.

PR #222, opened 2017-09-08
Removing hardcoded LDAP admin password

robdaemon:
LDAP password is now generated when admin node starts, and LDAP parameters are passed via the pillar during bootstrap.

### DIFF
--- a/salt/dex/dex.yaml
+++ b/salt/dex/dex.yaml
@@ -44,25 +44,24 @@ data:
       config:
         host: "{{ pillar['dashboard'] }}:389"
         startTLS: true
-        bindDN: cn=admin,{{ pillar['ldap_internal_infra_domain'] }}
-        bindPW: admin
+        bindDN: "{{ pillar['ldap']['bind_dn'] }}"
+        bindPW: "{{ pillar['ldap']['bind_pw'] }}"
         rootCA: {{ pillar['ssl']['ca_file'] }}
         userSearch:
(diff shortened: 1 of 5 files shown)


robdaemon:
Depends-on: kubic-project/velum#301
Depends-on: kubic-project/salt#222
Depends-on: kubic-project/caasp-container-manifests#105

salt/dex/dex.yaml
```diff
@@ -44,25 +44,24 @@ data:
...
           filter: "(objectClass=groupOfUniqueNames)"
 
           userAttr: DN
           groupAttr: uniqueMember
 
-          nameAttr: cn       
-
+          nameAttr: cn
     oauth2:
       skipApprovalScreen: true
```
ereslibre: This comes from the previous PR but we have the convention to name files that will be processed by jinja with a jinja extension. Also I am not sure we need all that quoting that was added (same for other places in this PR where quoting was added).

I don't want this to be a blocker, can be tackled afterwards.

kiall: The quoting is, I believe, safe + correct, as people could - for example have a bind password that starts with a { - which would then break the YAML.

kiall:
These work fine, however there is a issue with the ldap container entrypoint and escaping slashes, which seems to hit every few runs.

kiall:
LDAP container fixed, working just fine now!

PR #223, opened 2017-09-08
Switch dex to sles12 image

robdaemon:


robdaemon:
This builds on kubic-project/salt#222

salt/dex/init.sls
```diff
@@ -93,32 +93,29 @@ dex_secrets:
...
       - {{ pillar['paths']['kubeconfig'] }}
 
 dex_instance:
   cmd.run:
     - name: |
-        until kubectl get sa dex --namespace=kube-system && kubectl get clusterrolebinding system:dex && kubectl get configmap dex --namespace=kube-system && kubectl get deployment dex --namespace=kube-system && kubectl get service dex --namespace=kube-system ; do
-            kubectl create -f /root/dex.yaml
-            sleep 5
-        done
+        kubectl apply -f /root/dex.yaml || kubectl apply -f /root/dex.yaml
```
inercia: Once the `kube-apiserver` is up and running no `kubectl apply` should fail, right? If that is the case why are we doing it twice?

kiall: No, this is now why it's being done. `kubectl apply` is ONLY idempotent if a resource already exists. There is a check+set race as the "create or update" logic is handled client side, so 3 masters running this all at the same time see "Dex doesn't exist", and all issue a create - 2 of the 3 will fail.

Running it twice was meant to ensure that if it gets a "Already exists" response to a create API call, that it will try once more, this time with the client logic deciding to send an update.

(However, that's still not fully idempotent - theres a bug open for that)

inercia: >  3 masters running this all at the same time see "Dex doesn't exist", and all issue a create - 2 of the 3 will fail.

Do you mean we will get 2 failures for the second `kubectl apply`? Then this whole state will fail, right?

kiall: No, lets say it's running on 3 nodes, there will be at most 5 attempts.

master 1 - attempt create (succeed!)
master 2 - attempt create (Fail!) - attempt update (succeed with no changes!)
master 3 - attempt create (Fail!) - attempt update (succeed with no changes!)


(However, there is STILL a race condition here, it's just rare, and there is a bug filed for it)

inercia: Ok, thanks for the clarification.

PR #224, opened 2017-09-11
Set etcd3 as default backend storage

ereslibre:
Depends on: https://github.com/kubic-project/velum/pull/303


mjura:
it is also related to https://github.com/kubic-project/salt/pull/210

PR #227, opened 2017-09-13
[bsc# 1057552] Fix the race condition that occurs with Kube-DNS creation in multi master environment

at1012:
...kube-dns creation in multi master set up

### DIFF
--- a/salt/addons/init.sls
+++ b/salt/addons/init.sls
@@ -38,6 +38,8 @@ apply-dns:
     - require:
       - kube-apiserver
       - file:      /etc/kubernetes/addons/kubedns.yaml
+    - check_cmd:
+      - kubectl get deploy kube-dns -n kube-system | grep kube-dns
 
 create-dns-clusterrolebinding:
   cmd.run:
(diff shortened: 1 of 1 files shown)


salt/addons/init.sls
```diff
@@ -38,6 +38,8 @@ apply-dns:
     - require:
       - kube-apiserver
       - file:      /etc/kubernetes/addons/kubedns.yaml
+    - check_cmd:
+      - kubectl get deploy kube-dns -n kube-system | grep kube-dns | cat
```
robdaemon: This will cause issues during an upgrade - if the kube-dns deployment is changed (say, to use a new version of the container), this will end up not applying the new kubedns.yaml file.

at1012: @robdaemon My understanding is that the kubectl apply will run irrespective of check_cmd's output. But, what this salt state will return is the return code from check_cmd rather than from kubectl apply. Let me know if you think otherwise.

https://docs.saltstack.com/en/latest/ref/states/requisites.html#check-cmd

SlickNik:
Hi Aish:

I know that the bug references only kube-dns, but I think the tiller deployment can also suffer from exactly the same race condition. Can you update that as well, please?

Thanks!

at1012:
Sure, will do.

kiall:
(I probably wasn't clear, in fact, I defiantly wasn't - we may want to ignore the "updates might not apply" for now, for the sake of getting reliable bootstraps :))

at1012:
Thanks for pointing it out @kiall. I have now removed the ` | cat `.

kiall:
Is this still WIP? Can we remove the WIP from the commit message? (and add the "bsc#0000000" bug number?)

robdaemon:
Kiall approved, merging

PR #228, opened 2017-09-15
Update paths to match SLES based Dex container

kiall:
The SLES based dex container does not put dex in /usr/local/bin

Part of bsc#1058833


jordimassaguerpla:
lgtm

PR #229, opened 2017-09-19
Dex: Wait for Dex to be fully up and running

kiall:
We shouldn't allow a bootstrap to complete without Dex being up and
running, so lets wait for the Dex API to start responding.

### DIFF
--- a/salt/orch/kubernetes.sls
+++ b/salt/orch/kubernetes.sls
@@ -160,13 +160,13 @@ reboot_setup:
     - require:
       - salt: kube_master_setup
 
-dex_setup:
+wait_for_dex_api:
   salt.state:
     - tgt: 'roles:kube-master'
     - tgt_type: grain
     - batch: 5
     - sls:
-      - dex
+      - dex.ensure-dex-running
     - require:
       - salt: reboot_setup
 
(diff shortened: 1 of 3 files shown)


kiall:
CI blocked by https://github.com/kubic-project/salt/pull/230

salt/orch/kubernetes.sls
```diff
@@ -160,13 +160,13 @@ reboot_setup:
     - require:
       - salt: kube_master_setup
 
-dex_setup:
```
kiall: Dex setup is already completed well before this stage, during the master node's highstate several steps up - hence replacing this with a "ensure its running".

ereslibre:
This would fix CI. Deploying a cluster to test it.

ereslibre:
We could use https://docs.saltstack.com/en/latest/ref/states/all/salt.states.http.html#salt.states.http.wait_for_successful_query, but let's refactor this later.

kiall:
> We could use https://docs.saltstack.com/en/latest/ref/states/all/salt.states.http.html#salt.states.http.wait_for_successful_query, but let's refactor this later.

Yea, we have a bunch of other waits like this using this pattern, if we want to change them, we should ideally do them all at once.

PR #230, opened 2017-09-19
Making the service account key the same on all nodes

robdaemon:
The kube-apiserver and kube-controller-manager must agree on what the
private key is for service account generation. In a multi-master
scenario, where an api server starts on one machine, and the
controller-manager on another machine becomes primary, pods cannot be
created because kube-controller-manager cannot communicate with the
apiserver.

So, now, we generate the service account key on the ca minion and store
it in the mine, so that it's generated once.

This will break if/when there is more than one machine with the ca role
in Salt.

Fixes bsc#1059398


kiall:
(merged - discussed with @ereslibre that the mount is unnecessary, we mount all of /etc/pki already)

ereslibre:
Right, forgot we were mounting `/etc/pki` as a whole. Everything is fine 👍

PR #231, opened 2017-09-20
Do not access infra machines through the proxy (bsc#1053739)

inercia:
Fixes https://bugzilla.suse.com/show_bug.cgi?id=1053739

### DIFF
--- a/salt/docker/init.sls
+++ b/salt/docker/init.sls
@@ -6,6 +6,11 @@ include:
 # proxy for the daemon
 #######################
 
+{% set no_proxy = ['.infra.caasp.local', '.cluster.local'] %}
+{% if salt['pillar.get']('proxy:no_proxy') %}
+  {% do no_proxy.append(pillar['proxy']['no_proxy']) %}
+{% endif %}
+
 /etc/systemd/system/docker.service.d/proxy.conf:
   file.managed:
     - makedirs: True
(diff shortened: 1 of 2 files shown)


salt/docker/init.sls
```diff
@@ -6,14 +6,16 @@ include:
...
+
 /etc/systemd/system/docker.service.d/proxy.conf:
   file.managed:
     - makedirs: True
     - contents: |
         [Service]
         Environment="HTTP_PROXY={{ salt['pillar.get']('proxy:http', '') }}"
         Environment="HTTPS_PROXY={{ salt['pillar.get']('proxy:https', '') }}"
-        Environment="NO_PROXY={{ salt['pillar.get']('proxy:no_proxy', '') }}"
+        Environment="NO_PROXY={{ extra_no_proxy }},{{ salt['pillar.get']('proxy:no_proxy', '') }}"
```
kiall: I wonder does the trailing comma cause issues for anything that reads and parses the no_proxy flag? Given it's hundreds of different pieces of code, I'm pretty sure something, somewhere, will somehow fail.

Can we wrap the , and pillar rendering into a if statement? (same in the other spots too)

inercia: This NO_PROXY is only used by `docker`, so I'm not sure it is really necessary at all...

kiall: We do the same thing in /etc/sysconfig/proxy and /root/.curlrc - any one of hundreds of tools will read these.

PR #233, opened 2017-09-20
Add `caasp_service.running_stable`

ereslibre:
This new state will allow us to make sure that a service is running
in a stable manner. Also, will do some waits in case systemd will
do retries on the background, what avoids instant failure from salt
being reported with a regular `service.running`.

Fixes: bsc#1059105

### DIFF
--- a/salt/_states/caasp_service.py
+++ b/salt/_states/caasp_service.py
@@ -1,0 +1,79 @@
+from __future__ import absolute_import
+import time
+
+import salt.exceptions
+import salt.utils
+
+def running_stable(name, enable=None, sig=None, init_delay=None, successful_retries_in_a_row=50,
+                      max_retries=300, delay_between_retries=0.1, **kwargs):
+    '''
+    Ensure that the service is running in a stable manner
+
+    name
+        The name of the init or rc script used to manage the service
+
+    enable
+        Set the service to be enabled at boot time, True sets the service to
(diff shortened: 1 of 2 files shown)


flavio:
This looks like a good idea to me, it would be even worth to be shared upstream with salt!

flavio:
While looking closer at the code I thought: why aren't we thinking the other way around? We could let the user specify a health check command and keep invoking this command until the health check passes or a maximum number of attempts are done.

The current code would slow down the deployment of the system quite a bit (see the delay between even the successful checks).

ereslibre:
> While looking closer at the code I thought: why aren't we thinking the other way around? We could let the user specify a health check command and keep invoking this command until the health check passes or a maximum number of attempts are done.

The current code would slow down the deployment of the system quite a bit (see the delay between even the successful checks).

I fundamentally agree that we can make it more generic, but I doubt in our case with etcd we will be able to avoid a bit of slowing down the deployment, since the idea is checking that etcd is running in a stable way.

I will confirm that this works and we can make it more generic and maybe even submit it to upstream as you suggested :)

ereslibre:
I can confirm I have deployed a cluster in the same environment that was failing. With this patch applied I could deploy a cluster even using DNS names for the administration dashboard.

I can confirm also that without this patch using DNS names on said environment failed 100% of the times.

PR #234, opened 2017-09-20
Prevent update-etc-hosts conflicting with bootstrap

kiall:
Fix another case where the etc hosts update orchestration would otherwise
conflict with the bootstrap / add node orchestration.

bsc#1059577


salt/orch/update-etc-hosts.sls
```diff
@@ -1,5 +1,5 @@
-{%- set updates_all_target = 'P@roles:(admin|kube-(master|minion)) and G@bootstrap_complete:true and not G@update_in_progress:true' %}
-{%- set updates_master_target = 'G@roles:kube-master and G@bootstrap_complete:true and not G@update_in_progress:true' %}
+{%- set updates_all_target = 'P@roles:(admin|kube-(master|minion)) and G@bootstrap_complete:true and not G@bootstrap_in_progress:true and not G@update_in_progress:true' %}
```
inercia: Imagine a node that has been added to the cluster but not provisioned yet. This could run the update in nodes that have not completed the bootstrap, right?

kiall: @inercia the code as written will not run the update on those nodes, they don't need it, and will get it as part of their initial bootstrap.

Given these nodes are likely to be added to the cluster any second now (e.g. Anyone doing auto deploy will accept and provision immediately), it's safest to avoid it running on those nodes.

PR #235, opened 2017-09-20
Add a URL off Velum as a valid OIDC redirect URI

robdaemon:
This will make it so that Dex will be happy to redirect you to velum

### DIFF
--- a/salt/dex/dex.yaml
+++ b/salt/dex/dex.yaml
@@ -70,6 +70,7 @@ data:
     - id: caasp-cli
       redirectURIs:
       - 'http://127.0.0.1'
+      - 'https://{{ pillar['dashboard'] }}/oidc/done'
       name: "CaaSP CLI"
       secret: "swac7qakes7AvucH8bRucucH"
 ---


salt/dex/dex.yaml
```diff
@@ -70,6 +70,7 @@ data:
     - id: caasp-cli
       redirectURIs:
       - 'http://127.0.0.1'
+      - 'https://{{ pillar['dashboard'] }}/oidc/done'
```
kiall: This won't (always) do what you expect - on clouds like OpenStack / AWS - The pillar['dashboard'] value should be (and must be on OpenStack) set to the instances internal IP.

However, the user accesses Velum through the instances external IP or DNS name, this will refuse to the the redirects, right?

kiall: I think we need to add a second user input field next to the Kubernetes FQDN option, asking for the Dashboard FQDN - then stash that in the pillar, and use it here.

Thoughts?

ereslibre: You are right.

inercia: ... and the default value for that field could be automatically filled from the `HTTP_HOST` (but some users could need to change it).

robdaemon: Didn't we already go down this path of HTTP_HOST being the wrong value?

robdaemon: @kiall - you should open a bugzilla entry for that. You're right it needs to be done, but it's outside of the scope of this. Also, we likely have this pattern somewhere else in the code that we need to look into.

kiall: I disagree that it's outside the scope - while we don't currently have that field, we have no current need for it. Merging this as is will end up creating a new bug and the functionality added here will be broken on OpenStack / AWS / etc.

robdaemon: This change may make that necessary for cloud providers, but I'm not working on cloud providers. I don't even have access to an OpenStack environment right this moment. Open something in Trello, and it can be handled after this is merged.

PR #236, opened 2017-09-21
Allow custom options to be passed to the Salt Master

kiall:
Rename the salt master configurations, so that custom options can
be loaded after the stock options, allowing an override.

bsc#1059724

Depends-On: https://github.com/kubic-project/caasp-container-manifests/pull/113


kiall:
Force merging due to cyclic dep

PR #238, opened 2017-09-22
Introduce Velum pillar

ereslibre:
* Use Velum pillar that serves json content
* Cache the result if it differs from what we got
* Serve the cached result if a connection problem happens

Depends on:

* https://github.com/kubic-project/velum/pull/315
* https://github.com/kubic-project/caasp-container-manifests/pull/116

Fixes: bsc#1069145


salt/_pillar/velum.py
```diff
@@ -0,0 +1,102 @@
...
+        current_sha256 = hashlib.sha256(json.dumps(data['dict']))
+        if pillar_sha256 is None or pillar_sha256 != current_sha256:
+            if pillar_cache:
+                pillar_cache.close()
+            pillar_cache = tempfile.TemporaryFile()
+            pillar_cache.write(json.dumps(data['dict']))
+            pillar_sha256 = current_sha256
+        return data['dict']
+    elif pillar_cache:
+        pillar_cache.seek(0)
```
inercia: Couldn't we just keep an in-memory copy of the pillar (a `pillar_cache_contents` global) instead of trying to load it from the file each time `dict is None`?

salt/_pillar/velum.py
```diff
@@ -0,0 +1,102 @@
...
+                                  password=password,
+                                  decode=True,
+                                  decode_type='json')
+
+    if 'dict' in data:
+        current_sha256 = hashlib.sha256(json.dumps(data['dict']))
+        if pillar_sha256 is None or pillar_sha256 != current_sha256:
+            if pillar_cache:
+                pillar_cache.close()
+            pillar_cache = tempfile.TemporaryFile()
```
inercia: Are we saving the contents of the pillar in clear? Aren't we going to have any sensitive data there?

ereslibre: `tempfile.TemporaryFile` should be safe, because it won't appear anywhere.

salt/_pillar/velum.py
ereslibre: Probably yes, will review this. We will lost the file anyways if the container or the host is restarted so it probably makes sense to directly cache it in memory...

config/master.d/50-master.conf
```diff
@@ -13,6 +13,12 @@ timeout: 20
...
   base:
     - /usr/share/salt/kubernetes/salt
+
 pillar_roots:
   base:
     - /usr/share/salt/kubernetes/pillar
+
+ext_pillar:
+  - velum:
+      url: https://localhost/internal-api/v1/pillar.json?version=1
```
MaximilianMeister: do we need the `?version=1` param when we have the `v1` namespace already?

ereslibre: I think it won't hurt. `v1` is more the version of the whole API in case we introduce not backwards-compatible changes in the future, but `version=1` for the pillar resource (ignored right now) can be really helpful between upgrades, if salt states expect the pillar structure in one way or the other. I'd keep it (despite we are ignoring it in Velum), but I think it can help in case it's needed.

robdaemon: No, choose to version one way, not two. Stick to the `/v1/` in the URL. The `/v1/` means your *entire* API is a v1 version, including the data model. @MaximilianMeister is right here.

If you want to change the model in any non-backwards compatible, you *must* rev the API version. We also have to expect that it's possible to have a cluster that needs both v1 and v2 of the API present at one time.

ereslibre: In my opinion they are different things. One is the version of the API (`/v1/`). The other is the version of one specific resource within the API (`pillar.json?version=n`), but since it doesn't make sense at the moment I will remove it.

robdaemon: This is a few years old now, but it generally lists some good standards for web APIs: https://github.com/WhiteHouse/api-standards

In an API design, you typically version the entire API, not the specific resources in the API. Take a look at the Kubernetes API for an example. They chose to version at the individual resource, which makes sense given the size of their API. My point here is that there aren't multiple levels of versioning in that API.

ereslibre: In a public API I'd agree with you without any doubt. However, this API is and was meant to be exclusively used by salt to implement a pillar override (and maybe if decided so, allow it to callback to it and update Velum from salt itself). Branching to new full API versions all endpoints because we decided to refactor the pillar between kubernetes-salt releases would force us to create a new set of controllers for all endpoints, when only the pillar resource contains relevant changes.

I see your point. I completely agree in a general purpose/public API about only having one way of versioning, but I see this API as something a bit different.

In any case I think my argument won't stand because 1) we didn't yet have this problem [why do this? only because we can], and 2) still having two versionings (general api and pillar resource specific) still doesn't sound like a good idea.

Will remove the `version` query parameter.

config/master.d/50-master.conf
```diff
@@ -13,6 +13,12 @@ timeout: 20
...
     - /usr/share/salt/kubernetes/salt
+
 pillar_roots:
   base:
     - /usr/share/salt/kubernetes/pillar
+
+ext_pillar:
+  - velum:
+      url: https://localhost/internal-api/v1/pillar.json
+      ca_bundle: /etc/pki/ca.crt
```
mjura: What editor do you use? I'm always confused with this "No newline at the enf of file"

config/master.d/50-returner.conf
```diff
@@ -5,71 +5,4 @@ mysql:
...
-    # Query for third level pillar values (two dots in `pillar` key)
-    - query: 'SELECT SUBSTRING_INDEX(SUBSTRING_INDEX(pillar, ":", 1), ":", -1) AS key1,
-                     SUBSTRING_INDEX(SUBSTRING_INDEX(pillar, ":", 2), ":", -1) AS key2,
-                     SUBSTRING_INDEX(SUBSTRING_INDEX(pillar, ":", 3), ":", -1) AS key3,
-                     value
-                     FROM pillars WHERE
-                       (LENGTH(pillar) - LENGTH(REPLACE(pillar, ":", "")) = 2) AND
-                       (minion_id IS NULL OR minion_id = %s)'
-      as_list: True
+  db: 'velum_production'
```
mjura: What editor do you use? I'm always confused with this "No newline at the enf of file"

config/master.d/50-master.conf
ereslibre: I use the TRUE editor, best known as emacs.

ereslibre:
PR has been rebased.

ereslibre:
[Serve pillars from Velum](https://trello.com/c/xiVesI4m/400-serve-pillars-from-velum)

salt/_pillar/velum.py
```diff
@@ -0,0 +1,97 @@
...
+# Import Salt libs
+import salt.ext.six as six
+
+import json
+import hashlib
+
+pillar_sha256 = None
+pillar_cache = None
+cached_username = None
+cached_password = None
```
flavio: Why do we need these global variables, is that because they persist across the whole life cycle of a salt-master instance?

salt/_pillar/velum.py
```diff
@@ -0,0 +1,97 @@
...
+
+    if username is None:
+        if cached_username is None:
+            with open('/var/lib/misc/infra-secrets/velum-internal-api-username', 'r') as f:
+                cached_username = f.read().strip()
+        username = cached_username
+
+    if password is None:
+        if cached_password is None:
+            with open('/var/lib/misc/infra-secrets/velum-internal-api-password', 'r') as f:
```
flavio: Just do double check, neither the `username` nor the `password` are specified into the `returner.conf` (see above). That means `/var/lib/misc/infra-secrets/velum-internal-api-password` and `/var/lib/misc/infra-secrets/velum-internal-api-username` are going to be used.

Their values are defined inside of PR https://github.com/kubic-project/caasp-container-manifests/pull/116/ but I cannot see them mounted into the salt container. Am I missing something?

ereslibre: Yes, `returner.conf` is not used anymore for the pillar access (note that we will be able to remove `returner.conf` completely when salt does not access the database for anything).

The `salt-master` container mounts `/var/lib/misc/infra-secrets` as a whole (https://github.com/kubic-project/caasp-container-manifests/blob/master/public.yaml#L186-L188), so this files should be accessible.

salt/_pillar/velum.py
flavio: Yes, their purpose is clear it me. I just wanted to double check about their "persistence".

flavio:
@inercia, yes the cache has been removed because it didn't work. @ereslibre is going to bring it back with a separate PR

kiall:
Needs a rebase over the other fixes within the salt repo to pass CI...

ereslibre:
Rebased.

ereslibre:
Rebased.

salt/_pillar/velum.py
```diff
@@ -0,0 +1,79 @@
+# -*- coding: utf-8 -*-
+'''
+A module that adds data to the Pillar structure retrieved by an http request made to Velum
+
+Configuring the VELUM ext_pillar
+====================================
+
+Set the following Salt config to setup Foreman as external pillar source:
```
flavio: "Foreman"? Is that some leftover from a copy & paste?

salt/_pillar/velum.py
```diff
@@ -0,0 +1,79 @@
+# -*- coding: utf-8 -*-
+'''
+A module that adds data to the Pillar structure retrieved by an http request made to Velum
```
flavio: Nitpicking: can you add a line explaining that results are also cached into salt's cache to ensure they are accessible even when velum is temporarily down?

salt/_pillar/velum.py
ereslibre: Ouch, copied the structure (laziness) and overlooked that! Will fix, thanks!

ereslibre:
@flavio PR updated.

PR #244, opened 2017-10-06
fix logic of kubectl invocation

MaximilianMeister:
as dex_instance is required there is no need to try and error for
two times, although i don't know if that was the intention

### DIFF
--- a/salt/dex/init.sls
+++ b/salt/dex/init.sls
@@ -101,6 +101,18 @@ dex_secrets:
 dex_instance:
   cmd.run:
     - name: |
+        # FIXME: workaround a limitation in our salt version to execute a command
+        # only on one master node.
+        # "kubectl apply" makes a client side decision on if it needs to issue a
+        # "CREATE" or "UPDATE" API call. When more than 1 master executes this
+        # at the same time, for the first time, the check to decide "CREATE or UPDATE"
+        # returns "CREATE" on more than 1 master, assuming the timing is right.
+        # Now, when both send the create API call, kubernetes will reject one of them
+        # with an "AlreadyExists" failure. This will cause salt to fail the bootstrap.
+        # However, if we try the command again on the machine that failed, it will
+        # this time see the resource created by the other master, and decide to issue
+        # a no-op UPDATE instead. This will pass, and all resources will be in a
+        # consistent state - at the expense of a an extra few API calls.
         kubectl apply -f /root/dex.yaml || kubectl apply -f /root/dex.yaml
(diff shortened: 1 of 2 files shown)


salt/dex/init.sls
```diff
@@ -112,7 +112,7 @@ dex_instance:
 kubernetes_roles:
   cmd.run:
     - name: |
-        kubectl apply -f /root/roles.yaml || kubectl apply -f /root/roles.yaml
```
mjura: This looks really confusing :) I understand that we would like remove it, but first probably we would like to resolve potential race condition as @kiall said.

MaximilianMeister:
@kiall although i still don't fully understand how this prevents a race condition, i wonder if there is not a better way to do that. i guess the right way is not easily possible without some sort of quorum based execution by a leader

i also found https://github.com/saltstack/salt/issues/14074 on the way so this seems like a salt limitation for now until we have a good HA mechanism for this usecases (executing sth only once on one master)

could you provide some comment that we could at least add here for clearing the confusion a bit?

kiall:
> @kiall although i still don't fully understand how this prevents a race condition, i wonder if there is not a better way to do that. i guess the right way is not easily possible without some sort of quorum based execution by a leader

Yea, what we really want is to run this (and all the other kubectl apply's) on 1 host, and only 1 host. Salt doesn't really give us a way to do that though (well, our version of salt.. the latest versions do).

> could you provide some comment that we could at least add here for clearing the confusion a bit?

This pattern is used in at least 3 different places in the salt codebase (dex, kube-dns, helm). The issue is that: `kubectl apply` makes a client side decision on if it needs to issue a "CREATE" or "UPDATE" API call. When more than 1 master executes this at the same time, for the first time, the check to decide "create or update" returns "create" on more than 1 master, assuming the timing is right. Now, when both send the create API call, kubernetes will reject one of them with an "AlreadyExists" failure. This will cause salt to fail the bootstrap. However, if we try the command again on the machine that failed, it will this time see the resource created by the other master, and decide to issue a no-op UPDATE instead. This will pass, and all resources will be in a consistent state - at the expense of a an extra few API calls.

MaximilianMeister:
@kiall thanks for the detailed explanation, makes total sense now. i was so free to add it exactly like you said as comments with a FIXME. it's a bit verbose and duplicated but i think there is no shorter way to explain this.

MaximilianMeister:
comment is now added to the commit message in https://github.com/kubic-project/salt/pull/256

closing...

PR #249, opened 2017-10-06
Add floating network to cloud-provider integration with OpenStack

flavio:
Backport of PR #242 242

We would like add new pillar value floating, which will be used to configure floating network
for cloud provider intergration with OpenStack. If this option is specified, it will create floating ip
for loadbalancer automatically.

(cherry picked from commit 497891de69f904f9de8cb87e7bb186926af990ce)

### DIFF
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -162,6 +162,8 @@ cloud:
     password:       ''
     # OpenStack subnet UUID for the CaasP private network
     subnet:         ''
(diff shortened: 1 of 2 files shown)


kiall:
<random note>Looks like my regex in Jenkins is matching this branch and running a second job - ignore the CI result for "continuous-integration/jenkins/branch"</random note>

PR #250, opened 2017-10-09
Use the default network interface instead of the hardcoded 'eth0' (bsc#1058079)

inercia:


pillar/beacons.sls
```diff
@@ -2,6 +2,7 @@ beacons:
   network_settings:
     interval: 5
     # monitorize any change in IP addresses in eth0
+    # TODO: make this interface configurable
```
flavio: Can you file a bug for that, so we don't forget about it?

monstermunchkin:
[Bug 1058079 — kubernetes-salt uses hardcoded eth0](https://trello.com/c/LKWXmcty/415-bug-1058079-kubernetes-salt-uses-hardcoded-eth0)

pillar/beacons.sls
inercia: Filed [here](https://bugzilla.suse.com/show_bug.cgi?id=1063709)

kiall:
Needs a rebase over the other fixes within the salt repo to pass CI...

kiall:
Now that the supportconfig dump has the right logs again, this seems to be failing due to:

> ? cmd_|-load_flannel_cfg_|-/usr/bin/etcdctl --ca-file /etc/pki/trust/anchors/SUSE_CaaSP_CA.crt
>   --key-file /etc/pki/minion.key --cert-file /etc/pki/minion.crt --endpoints https://478f26722ce74a099907be27d103b29e.infra.caasp.local:2379
>   --no-sync set /flannel/network/config < /root/flannel-config.json_|-run
> : comment: Command "/usr/bin/etcdctl --ca-file /etc/pki/trust/anchors/SUSE_CaaSP_CA.crt
>     --key-file /etc/pki/minion.key --cert-file /etc/pki/minion.crt --endpoints
>     https://478f26722ce74a099907be27d103b29e.infra.caasp.local:2379 --no-sync
>     set /flannel/network/config < /root/flannel-config.json" run
>   name: "/usr/bin/etcdctl --ca-file /etc/pki/trust/anchors/SUSE_CaaSP_CA.crt --key-file
>     /etc/pki/minion.key --cert-file /etc/pki/minion.crt --endpoints https://478f26722ce74a099907be27d103b29e.infra.caasp.local:2379
>     --no-sync set /flannel/network/config < /root/flannel-config.json"
>   start_time: '12:28:21.560184'
>   result: false
>   duration: 2665.791
>   __run_num__: 19
>   changes:
>     pid: 3605
>     retcode: 4
>     stderr: |-
>       Error:  client: etcd cluster is unavailable or misconfigured; error #0: client: endpoint https://478f26722ce74a099907be27d103b29e.infra.caasp.local:2379 exceeded header timeout
> 
>       error #0: client: endpoint https://478f26722ce74a099907be27d103b29e.infra.caasp.local:2379 exceeded header timeout
>     stdout: ''
>   __id__: load_flannel_cfg

mjura:
I'm not sure if using gateway_interface is good idea. For many deployments, there will be many different networks used and for core services like flanneld or etcd, users will prepare private networks without external networks. This is true for OpenStack and Kubernetes as well. We should allow users to configure this networks by themselves

flavio:
What if:

  * We make the network card configurable via a pillar value (I guess this is what @inercia proposed with [his comment](https://github.com/kubic-project/salt/pull/250/files#diff-b5edd177eab2d06f40b20ea3079872c0R5)))
  * Set the default value of the pillar to be `eth0`
  * With a different PR we extend velum to change the value of this pillar value

inercia:
> I'm not sure if using gateway_interface is good idea. For many deployments, there will be many different networks used and for core services like flanneld or etcd, users will prepare private networks without external networks. This is true for OpenStack and Kubernetes as well. We should allow users to configure this networks by themselves

Please refer to the discussion in bugzilla. We decided there that using the default interface would be a good enough solution.

PR #251, opened 2017-10-09
Allow Dex to redirect to the Dashboard's external FQDN

kiall:
Some scenarios where the admin node's private IP is not accessible
to the outside world require that we use a end user provided FQDN
- e.g. as is the case on OpenStack and possibly other cloud
environments. Allow redirections to this FQDN.

Part of bsc#1062291

Depends-On: https://github.com/kubic-project/velum/pull/322
Depends-On: https://github.com/kubic-project/automation/pull/164

### DIFF
--- a/salt/dex/dex.yaml
+++ b/salt/dex/dex.yaml
@@ -73,6 +73,7 @@ data:
       redirectURIs:
       - 'http://127.0.0.1'
       - 'https://{{ pillar['dashboard'] }}/oidc/done'
+      - 'https://{{ pillar['dashboard_external_fqdn'] }}/oidc/done'
       name: "CaaSP CLI"
       secret: "swac7qakes7AvucH8bRucucH"
 ---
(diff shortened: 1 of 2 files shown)


kiall:
So - I think this is incomplete, as the newly collected external FQDN will not be included within the Velum TLS certificate, and it will need to be.

salt/dex/dex.yaml
```diff
@@ -73,6 +73,7 @@ data:
       redirectURIs:
       - 'http://127.0.0.1'
       - 'https://{{ pillar['dashboard'] }}/oidc/done'
```
ereslibre: Do we want to keep this redirection? We'll use the new redirection host always, no?

kiall: As I understand it, we'll use whatever URL the user has in their browser - considering both could be used, I left it in.. I don't *think* it would cause any issues.

robdaemon: Kiall's right here - anything that could possibly be used as the redirect has to be here, or Dex will reject it.

PR #252, opened 2017-10-09
Update tiller deployment to use sles-based docker image

SlickNik:
Currently the tiller image being used for the tiller deployment is
from the upstream registry at gcr.io. We should be using the SLES
based docker image instead of the upstream one.

Fixes: bsc#1062380


kiall:
CI is broken right now, and won't test this anyway. Verified the new image name is correct:

> sles12/tiller             2.5.1               a47207fd2d81        3 days ago          261 MB

will merge without CI.

PR #255, opened 2017-10-11
[2.0] Allow Dex to redirect to the Dashboard's external FQDN

kiall:
Some scenarios where the admin node's private IP is not accessible
to the outside world require that we use a end user provided FQDN
- e.g. as is the case on OpenStack and possibly other cloud
environments. Allow redirections to this FQDN.

Part of bsc#1062291

Backport of https://github.com/kubic-project/salt/pull/251

### DIFF
--- a/pillar/vars.sls
+++ b/pillar/vars.sls
@@ -1,4 +1,7 @@
 # some global variables that will be set dynamically
 
 # a distinguishable name/IP for the dashboard
-dashboard:     'dashboard'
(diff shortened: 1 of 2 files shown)


monstermunchkin:
[Bug 1058079 — kubernetes-salt uses hardcoded eth0](https://trello.com/c/LKWXmcty/415-bug-1058079-kubernetes-salt-uses-hardcoded-eth0)

PR #256, opened 2017-10-11
Add a delay before retrying kubectl commands

kiall:
We currently retry straight away, which can sometimes simply be too fast and we'll
get a similar error again. Add a short 3 second delay before retrying any kubectl
commands.

Additionally, remove the check_cmd's from the K8S addons, as these don't actually
check anything which has not already been asserted through the primary command
itself, and again, these occasionally run too fast, before K8S has fully processed
the tasks from the primary command.

bsc#1062818

### DIFF
--- a/salt/addons/init.sls
+++ b/salt/addons/init.sls
@@ -16,7 +16,7 @@ include:
 apply-namespace:
   cmd.run:
     - name: |
-        kubectl apply -f /etc/kubernetes/addons/namespace.yaml || kubectl apply -f /etc/kubernetes/addons/namespace.yaml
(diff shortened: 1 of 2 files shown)


kiall:
This seems to affect slower environments (like CI) much more than fast environments (like most real environments), but should work out better for all anyway.

monstermunchkin:
[Bug 1062818 - On slower environments, kubectl commands from salt may fail unexpectedly](https://trello.com/c/XcCtzRRC/412-bug-1062818-on-slower-environments-kubectl-commands-from-salt-may-fail-unexpectedly)

kiall:
This seems to no longer be an issue now that kube-api is no longer restarting in a loop.

MaximilianMeister:
@kiall could it be that this resurfaced in https://bugzilla.suse.com/show_bug.cgi?id=1059105 ?

PR #262, opened 2017-10-17
Revert K8S to use etcd2 storage format

kiall:
With etcd3, the kubernetes api server will sit in a (slow) restart loop
when multimaster is enabled, logging a stacktrace and then restarting.
This will manifest as, most commonly, "Unable to connect to the server:
unexpected EOF" from kubectl. This will break bootstrap as we need to
talk to K8S API to deploy dex, kube-dns, and tiller.

bsc#1063235
bsc#1063285
bsc#1063543

### DIFF
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -23,7 +23,7 @@ api:
   # port for listening for SSL connections (kube-api)
   int_ssl_port:   '6444'
   # etcd storage backend version for cluster
-  etcd_version:   'etcd3'
+  etcd_version:   'etcd2'
 
   server:
     # should be set from the UI


MaximilianMeister:
> do we also need to drop code from #210 ?

and https://github.com/kubic-project/salt/blob/572773ed40865487574acb00c303cdf491238142/salt/etcd/etcdctl.conf.jinja#L9 ?

MaximilianMeister:
[Bug 1063235 — Cluster bootstrap fails](https://trello.com/c/6961TAv2/417-bug-1063235-cluster-bootstrap-fails)

kiall:
No, both of those are OK to leave in place.

We do probably want to drop the DB migration from Velum that does the pillar row creation during upgrades, if this lands. (Otherwise, we'll end up with a weird mix of possibilities)

MaximilianMeister:
> No, both of those are OK to leave in place.
> 
> We do probably want to drop the DB migration from Velum that does the pillar row creation during upgrades, if this lands. (Otherwise, we'll end up with a weird mix of possibilities)

ok, [this](https://github.com/kubic-project/salt/blob/c5998cc9bb4d050123e5ffcf59fdd050a33fdf02/salt/etcd/create-or-update-etcd-pillar) looks like the migration. wouldnt we need to drop it as part of this PR?
in velum etcd2 is the default. see [here](https://github.com/kubic-project/velum/blob/master/db/migrate/20170911123313_set_etcd2_as_default_backend.rb)

PR #264, opened 2017-10-17
WIP: Add RBAC Role/Binding for /healthz endpoint

kiall:
WIP as currently untested

This allows for anyone to read the healthz endpoint, including HAProxy or
other load balancers.


kiall:
This is unnecessary, it seems that once the restart loops were fixed, the healthz endpoint was correctly accessible.

PR #271, opened 2017-10-18
Manage the Velum TLS cert

kiall:
This ensures that the dashboard_external_fqdn is registered within
the velum TLS certificate.

bsc#1063998

### DIFF
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -67,6 +67,9 @@ ssl:
   crt_dir:        '/etc/pki'
   key_dir:        '/etc/pki'
 
+  velum_key: '/etc/pki/private/velum.key'
+  velum_crt: '/etc/pki/velum.crt'
+
   kube_apiserver_key: '/etc/pki/kube-apiserver.key'
   kube_apiserver_crt: '/etc/pki/kube-apiserver.crt'
 
--- a/salt/velum/init.sls
+++ b/salt/velum/init.sls
@@ -1,0 +1,59 @@
+include:
+  - ca-cert
+  - cert
+
+{% set ip_addresses = ["IP: " + pillar['dashboard']] -%}
+{% set extra_names = ["DNS: " + grains['caasp_fqdn'], "DNS: " + pillar['dashboard_external_fqdn']] -%}
+
+{{ pillar['ssl']['velum_key'] }}:
+  x509.private_key_managed:    
+    - bits: 4096
+    - user: root
+    - group: root
+    - mode: 444
+    - require:
+      - sls:  crypto
+      - file: /etc/pki
+
+{{ pillar['ssl']['velum_crt'] }}:
+  x509.certificate_managed:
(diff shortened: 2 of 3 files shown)


pillar/params.sls
```diff
@@ -67,6 +67,9 @@ ssl:
   crt_dir:        '/etc/pki'
   key_dir:        '/etc/pki'
 
+  velum_key: '/etc/pki/velum2.key'
```
kiall: Leaving it with a "bad" name for testing + comparing

monstermunchkin:
[Bug 1063998 — Velum TLS certificate needs to contain the dashboard_external_fqdn / dashboard values](https://trello.com/c/vD3E44Zx/418-bug-1063998-velum-tls-certificate-needs-to-contain-the-dashboardexternalfqdn-dashboard-values)

salt/velum/init.sls
```diff
@@ -0,0 +1,59 @@
...
+# TODO: There should be a better way to handle this, but currently, there is not. See
+# kubernetes/kubernetes#24957
+velum_restart:
+  cmd.run:
+    - name: |-
+        velum_id=$(docker ps | grep "velum-dashboard" | awk '{print $1}')
+        if [ -n "$velum_id" ]; then
+            docker kill -s USR2 $velum_id
+        fi
+    - onchanges:
```
inercia: Maybe we should `require: - docker`, as we are doing a `docker kill`...

pillar/params.sls
```diff
@@ -67,6 +67,9 @@ ssl:
   crt_dir:        '/etc/pki'
   key_dir:        '/etc/pki'
 
+  velum_key: '/etc/pki/private/velum2.key'
+  velum_crt: '/etc/pki/velum2.crt'
```
inercia: I guess these certificates are `bind`-mounted in the Velum container, so are you sure the container will get an updated version of the certificate?

salt/velum/init.sls
kiall: Adding a require is salt is only needed if your're adding a requisite on a task within the other state - we're not here, so it's probably best to keep it simple..

pillar/params.sls
kiall: Sadly, you're right - the new certs did not get loaded.. I'll (probably) need to switch to a `docker restart`

inercia: Try to mount in the container _the directory_ and not the files.

ereslibre: I think this is more a result of the `USR2` signal on puma not reloading the certificates rather than the changes done by salt are not i visible on the container. Switching to a hard restart of the container should work, but it will bring a brief downtime.

kiall: If we mount the directory, then velum has access to every certificate within it. This would be a security issue.. Besides, there is a second issue - Puma doesn't reload the certs from disk without a restart anyway.

inercia: Then just `docker restart`! :-D

PR #278, opened 2017-10-20
Add `caasp_cmd` state module featuring `run` with retry feature

ereslibre:
This state module will provide `run` state with `retry` option that
accepts `attempts` and `interval` arguments. This allow us to retry a command
if it failed, and retry to this maximum number of retries, sleeping
between retries.

### DIFF
--- a/salt/_states/caasp_cmd.py
+++ b/salt/_states/caasp_cmd.py
@@ -1,0 +1,202 @@
+from __future__ import absolute_import
+import time
+
+import salt.exceptions
+import salt.utils
+
+def run(name,
+        onlyif=None,
+        unless=None,
+        creates=None,
+        cwd=None,
+        runas=None,
+        shell=None,
+        env=None,
+        stateful=False,
+        umask=None,
+        output_loglevel='debug',
+        quiet=False,
+        timeout=None,
+        ignore_timeout=False,
+        use_vt=False,
+        retry={},
+        **kwargs):
+    '''
+    Run a command if certain circumstances are met.  Use ``cmd.wait`` if you
+    want to use the ``watch`` requisite.
+
+    name
+        The command to execute, remember that the command will execute with the
+        path and permissions of the salt-minion.
+
+    onlyif
+        A command to run as a check, run the named command only if the command
+        passed to the ``onlyif`` option returns true
+
+    unless
+        A command to run as a check, only run the named command if the command
+        passed to the ``unless`` option returns false
+
+    cwd
+        The current working directory to execute the command in, defaults to
+        /root
+
+    runas
+        The user name to run the command as
+
+    shell
+        The shell to use for execution, defaults to the shell grain
+
+    env
+        A list of environment variables to be set prior to execution.
+        Example:
+
+        .. code-block:: yaml
+
+            script-foo:
+              caasp_cmd.run:
+                - env:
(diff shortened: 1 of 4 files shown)


salt/_states/caasp_cmd.py
```diff
@@ -0,0 +1,202 @@
...
+                                    shell=shell,
+                                    env=env,
+                                    stateful=stateful,
+                                    umask=umask,
+                                    output_loglevel=output_loglevel,
+                                    quiet=quiet,
+                                    timeout=timeout,
+                                    ignore_timeout=ignore_timeout,
+                                    use_vt=use_vt,
+                                    **kwargs)
```
inercia: Instead of doing:

```python
def run(name,
        onlyif=None,
        unless=None,
        creates=None,
        cwd=None,
        runas=None,
        shell=None,
        env=None,
        stateful=False,
        umask=None,
        output_loglevel='debug',
        quiet=False,
        timeout=None,
        ignore_timeout=False,
        use_vt=False,
        retry={},
        **kwargs):

    # ...

    ret = __states__['cmd.run'](name=name,
                                    onlyif=onlyif,
                                    unless=unless,
                                    creates=creates,
                                    cwd=cwd,
                                    runas=runas,
                                    shell=shell,
                                    env=env,
                                    stateful=stateful,
                                    umask=umask,
                                    output_loglevel=output_loglevel,
                                    quiet=quiet,
                                    timeout=timeout,
                                    ignore_timeout=ignore_timeout,
                                    use_vt=use_vt,
                                    **kwargs)
```

couldn't we just do?

```python
def run(name, **kwargs):
    # ...
    ret = __states__['cmd.run'](name, **kwargs)
```

Do we need to force default values for those arguments?

flavio: I think that's needed because @ereslibre introduced the new `retry` dictionary option. However it would be possible to change the code to take a `kwargs`, remove the `retry` option and hand it over to `cmd.run`...

ereslibre: Yes, it will be cleaner. Will refactor it, thanks for your feedback @inercia and @flavio.

flavio:
BTW, I also guess this is going to go away once we have the [retrying states](https://docs.saltstack.com/en/latest/ref/states/requisites.html#retrying-states) feature of salt 2017.7.0 (either by updating to this version of salt of by backporting the patch into our salt)

ereslibre:
> BTW, I also guess this is going to go away once we have the retrying states feature of salt 2017.7.0 (either by updating to this version of salt of by backporting the patch into our salt)

Exactly, this whole state module is going to be away when we update, that's why we shared the same syntax 2017.7.0 will use. When we update we will remove `caasp_cmd` and just replace `caasp_cmd.run` with `cmd.run` as salt 2017.7.0 will support `retry` with `attempts` and `interval`.

inercia:
> Exactly, this whole state module is going to be away when we update, that's why we shared the same syntax 2017.7.0 will use. When we update we will remove caasp_cmd and just replace caasp_cmd.run with cmd.run as salt 2017.7.0 will support retry with attempts and interval.

We can use this until we have `salt 2017.7.0`. Merging...

flavio:
Why haven't we backported this PR to the 2.0 branch?

PR #283, opened 2017-10-24
Use `http.wait_for_successful_query` instead of looping with `curl`

inercia:


salt/dex/ensure-dex-running.sls
```diff
@@ -1,16 +1,13 @@
...
-    - name: |
-        {% set dex_api_server = pillar['api']['server']['external_fqdn'] -%}
-        {% set dex_api_port = pillar['dex']['node_port'] -%}
-        {% set dex_api_server_url = 'https://' + dex_api_server + ':' + dex_api_port -%}
+  http.wait_for_successful_query:
+    {% set dex_api_server = pillar['api']['server']['external_fqdn'] -%}
+    {% set dex_api_port = pillar['dex']['node_port'] -%}
+    - name:       {{ 'https://' + dex_api_server + ':' + dex_api_port }}/.well-known/openid-configuration
+    - wait_for:   300
+    #- cert:       [{{ pillar['ssl']['crt_file'] }}, {{ pillar['ssl']['key_file'] }}]
```
inercia: Quite sure we do not need these certs...

ereslibre: I'd be in favour of removing it instead of commenting it out.

inercia:
> I didn't test this patch, does ensure_dex_running still work as expected? (I don't see a reason why it wouldn't, but confirming just in case).

Yeah, it works fine.

PR #285, opened 2017-10-24
Remove some nasty loops

inercia:
New `retry['until']` argument for `caasp_cmd.run`
Use a `unless/onlyif` and `retry['until']` for skipping some executions and not using some nasty loops

### DIFF
--- a/salt/_states/caasp_cmd.py
+++ b/salt/_states/caasp_cmd.py
@@ -152,7 +152,9 @@ def run(name,
 
     retry
         This allows you to provide `attempts` and `interval`, what will retry the command as much
-        ``attempts`` times, separated by `interval` seconds.
+        ``attempts`` times, separated by `interval` seconds. If 'until' is provided,
+        verify we must not try again after running the command successfully
+        by runing the `until` condition.
 
     '''
     ### NOTE: The keyword arguments in **kwargs are passed directly to the
@@ -161,7 +163,8 @@ def run(name,
     ###       ``caasp_cmd.run`` state will result in a traceback.
 
     retry_ = {'attempts': 1,
-              'interval': 1}
+              'interval': 1,
+              'until': None}
     retry_.update(retry)
 
     ret = None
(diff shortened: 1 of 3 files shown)


inercia:
The `until` argument is not strictly necessary for removing the loops (we can avoid looping by using `unless/onlyif` in most cases). Maybe we can remove it once we have the `retry` in the standard Salt package...

salt/_states/caasp_cmd.py
```diff
@@ -152,7 +152,9 @@ def run(name,
 
     retry
         This allows you to provide `attempts` and `interval`, what will retry the command as much
-        ``attempts`` times, separated by `interval` seconds.
+        ``attempts`` times, separated by `interval` seconds. If 'until' is provided,
+        verify we must not try again after running the command successfully
+        by runing the `until` condition.
```
ereslibre: Minor: typo: `running`

salt/_states/caasp_cmd.py
```diff
@@ -185,10 +188,33 @@ def run(name,
...
+                                                    quiet=quiet,
+                                                    timeout=timeout,
+                                                    ignore_timeout=ignore_timeout,
+                                                    use_vt=use_vt,
+                                                    **kwargs)
+            if retry_until_ret['result']:
+                return ret_success
+
+            # append the 'until' command output, so we can have some debugging info...
+            ret['comment'] = ret['comment'] + "(until: " + retry_until_ret['comment'] + ")"
```
ereslibre: Is this meant for the last retry? Otherwise `ret` is being overriden on each loop pass.

salt/dex/init.sls
```diff
@@ -85,12 +85,16 @@ include:
...
-        until kubectl get secret dex-tls --namespace=kube-system ; do
-            kubectl create secret generic dex-tls --namespace=kube-system --from-file=/etc/pki/dex.crt --from-file=/etc/pki/dex.key
-            sleep 5
-        done
+        kubectl create secret generic dex-tls --namespace=kube-system --from-file=/etc/pki/dex.crt --from-file=/etc/pki/dex.key
+    - unless: |
+        kubectl get secret dex-tls --namespace=kube-system
+    - retry:
+        until: |
+          kubectl get secret dex-tls --namespace=kube-system
```
ereslibre: If `kubectl create` returns a 0 errcode it's supposed to have created this secret, so I think we can remove the `until` here. The `retry` will only retry if the errcode is not 0.

salt/_states/caasp_cmd.py
inercia: Yes, it is

salt/dex/init.sls
inercia: We are keeping the same logic we previously had: we cannot trust `kubectl create/apply` because of concurrency issues. So we must repeat the command (regardless of the output) until the `kubectl get` succeeds.

PR #286, opened 2017-10-24
Replace some other certificates by Jinja templates

inercia:
... like the `kubelet`, `dex` and `kube-proxy` ones (continues the work started in #280)

### DIFF
--- a/pillar/certificates.sls
+++ b/pillar/certificates.sls
@@ -14,3 +14,36 @@ certificate_information:
   days_remaining:
     ca_certificate: 90
     certificate: 90
+
+ssl:
+  ca_dir: '/etc/pki/trust/anchors'
+  ca_file: '/etc/pki/trust/anchors/SUSE_CaaSP_CA.crt'
+  crt_file: '/etc/pki/minion.crt'
+  key_file: '/etc/pki/minion.key'
+
+  crt_dir: '/etc/pki'
+  key_dir: '/etc/pki'
+
+  velum_key: '/etc/pki/private/velum.key'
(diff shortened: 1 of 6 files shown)


pillar/certificates.sls
```diff
@@ -14,3 +14,37 @@ certificate_information:
   days_remaining:
     ca_certificate: 90
     certificate: 90
+
+ssl:
```
inercia: Maybe we could remove this top `ssl` name by something more consistent: `certs_paths`, `certificates_paths`...

salt/dex/init.sls
```diff
@@ -88,14 +39,14 @@ dex_secrets:
   cmd.run:
     - name: |
         until kubectl get secret dex-tls --namespace=kube-system ; do
-            kubectl create secret generic dex-tls --namespace=kube-system --from-file=/etc/pki/dex.crt --from-file=/etc/pki/dex.key
+            kubectl create secret generic dex-tls --namespace=kube-system --from-file={{ pillar['ssl']['dex_crt'] }} --from-file={{ pillar['ssl']['dex_key'] }}
```
inercia: This will conflict with #285

inercia:
> Do you think we could change the pillar structure, so for example, instead of having velum_key and velum_crt we could have:

Yeah, my idea was to refactor the pillar later on, after these PRs. This could allow us to just pass, for example, `certs="velum"`, as an argument and the Jinja template could take the `(key,crt)` pair from the pillar.

ereslibre:
LGTM, will give it a try on monday.

PR #287, opened 2017-10-25
Store some key containers IDs as grains

inercia:
Store some key containers IDs as grains, and use those grains for doing things like 'docker restart'

### DIFF
--- a/salt/velum/init.sls
+++ b/salt/velum/init.sls
@@ -27,11 +27,8 @@ include:
 #       we have to sync this file again with Admin node.
 velum_restart:
   cmd.run:
-    - name: |-
-        velum_id=$(docker ps | grep "velum-dashboard" | awk '{print $1}')
-        if [ -n "$velum_id" ]; then
-            docker restart $velum_id
-        fi
+    - name: docker restart {{ salt['grains.get']('containers:velum', '') }}
+    - onlyif: test -n "{{ salt['grains.get']('containers:velum', '') }}"
     - onchanges:
       - x509: {{ pillar['ssl']['velum_key'] }}
       - x509: {{ pillar['ssl']['velum_crt'] }}
--- a/salt/_grains/containers.py
+++ b/salt/_grains/containers.py
@@ -1,0 +1,38 @@
+import logging
+import subprocess
+
+import salt.log
+
+log = logging.getLogger(__name__)
+
+def caasp_containers():
+    # NOTE: containers that are not running will have an entry in
+    #       the "containers" dictionary anyway, but it will be empty:
+    #       this way we can check if the value is empty (easy) instead of
+    #       a missing key in a grain dictionary (hard).
+    containers = {
+        'haproxy': '',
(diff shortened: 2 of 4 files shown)


kiall:
I'd be worried these get out of sync, the grains get refreshed periodically, and we want to always have the ID available when we need to restart etc. If they get out of sync, orchestration will fail, and a re-run will often NOT restart the container - as the "onchanges" etc won't trigger the states again on the second run.

inercia:
> I'd be worried these get out of sync, the grains get refreshed periodically, and we want to always have the ID available when we need to restart etc. If they get out of sync, orchestration will fail, and a re-run will often NOT restart the container - as the "onchanges" etc won't trigger the states again on the second run.

@kiall Containers IDs do not change on a `docker restart`, so I think the only case when a container could change the ID is after a crash, where the container would be restarted with a different ID. In that case the state could fail if the grain is not up-to-date. However, I think we are forcing a grain update as the first step in the orchestration, so the problem could arise only if the container crashes in between the orchestration beginning and the point when the container must be _restarted/killed/whatever_.  I think that is rather unlikely, but we could add a `retry` just to be sure...

> Can you provide some context regarding this @inercia?

@ereslibre I'm just trying to reduce the amount of `bash-in-Salt` we have, something that makes our code fragile in my view...

ereslibre:
> @ereslibre I'm just trying to reduce the amount of bash-in-Salt we have, something that makes our code fragile in my view...

I'm a totally +1 there, but I meant context regarding the caching/invalidation of this grains.

salt/velum/init.sls
```diff
@@ -24,11 +24,8 @@ include:
 #       by web servers instead of application servers (apache, nginx...).
 velum_restart:
   cmd.run:
-    - name: |-
-        velum_id=$(docker ps | grep "velum-dashboard" | awk '{print $1}')
-        if [ -n "$velum_id" ]; then
-            docker restart $velum_id
-        fi
+    - name: docker restart {{ grains['containers']['velum'] }}
+    - onlyif: test -n "{{ grains['containers']['velum'] }}"
```
inercia: We could include `docker-py` and use the `dockerng` module as:

```Saltstack
velum_restart:
  module.run:
    - name: dockerng.restart
    - args:
      - {{ grains['containers']['velum'] }}   # we could use a name too
```

flavio:
As said quickly during our standup, I'm agree we have to remove the usage of bash inside of our salt states. I think we have to use the docker module of salt however.

nanoscopic:
If we want to move away from bash in salt in this way; we cannot simply move to dockerng salt module as we are moving towards using crio instead of docker for cri.

There is no crio salt module.

My suggestion here would be to start our own salt module for podman; calling the podman API:
https://github.com/containers/libpod/blob/master/API.md

That would then be able to do checks/operations against both docker or crio setups.

Realistically this PR should just be closed at this point as these changes won't be integrated. Closing.

PR #289, opened 2017-10-25
Wait for etcd to be alive

inercia:
Wait for `etcd` to be alive before trying to set anything
Remove the 'ssl:enabled' pillar: we do not really support non-SSL deployments...

### DIFF
--- a/salt/etcd/init.sls
+++ b/salt/etcd/init.sls
@@ -40,16 +40,27 @@ etcd:
     - enable: True
     - require:
       - sls: ca-cert
-      - sls: cert
       - pkg: etcd
       - iptables: etcd
     - watch:
+      - {{ pillar['ssl']['crt_file'] }}
+      - {{ pillar['ssl']['key_file'] }}
+      - {{ pillar['ssl']['ca_file'] }}
+    - watch:
       - file: /etc/sysconfig/etcd
-
-etcd-running:
-  service.running:
-    - name: etcd
-    - enable: True
+  # wait until etcd is actually up and running
+  caasp_cmd.run:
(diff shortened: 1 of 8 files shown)


salt/etcd/init.sls
```diff
@@ -40,16 +40,27 @@ etcd:
...
-    - enable: True
+  # wait until etcd is actually up and running
+  cmd.run:
+    - name: |
+        etcdctl --key-file {{ pillar['ssl']['key_file'] }} \
+                --cert-file {{ pillar['ssl']['crt_file'] }} \
+                --ca-file {{ pillar['ssl']['ca_file'] }} \
+                --endpoints https://{{ grains['caasp_fqdn'] }}:2379 \
+                cluster-health | grep "cluster is healthy"
+    - retry:
```
kiall: I don't understand how this passed? Retry isn't in our version of salt? Huh?

inercia: Right, I should have used `caasp_cmd.run`. I think Salt doesn't complain about unrecognized arguments...

salt/flannel-setup/init.sls
```diff
@@ -21,9 +21,12 @@ load_flannel_cfg:
     - name: /usr/bin/etcdctl {{ etcd_opt }}
                              --no-sync
                              set {{ pillar['flannel']['etcd_key'] }}/config < /root/flannel-config.json
+    - retry:
+        attempts: 10
```
kiall: More retry's using the stock cmd.run :)

inercia: Fixed

PR #291, opened 2017-10-27
Fix random iptables crashes bsc 1064186

flavio:
This is an attempt to fix bsc#1064186 by executing the iptables statements more than once.

This PR introcuces also `caasp_retriable.retry` state that allows to retry any kind of salt state. Once merged we could get rid of the `caasp_cmd`.


flavio:
Updated, the orchestration is working fine locally

PR #296, opened 2017-11-02
Reorganize the addons and use some Jinja macros

inercia:
Reorganize the addons in a subdirectory per addon
Use some Jinja macros for running kubectl with retries, the kubectl path and the right dependencies


salt/addons/dns/init.sls
```diff
@@ -0,0 +1,26 @@
+{% if salt['pillar.get']('addons:dns', 'false').lower() == 'true' %}
+
+include:
+  - kube-apiserver
+  - kubectl-config
```
kiall: I wonder, can the macro do the include?

inercia: It could, but then there could be a problem if the macro is invoked twice... I would prefer to keep it as it is.

PR #297, opened 2017-11-03
Queue states for the update-etc-hosts orchestration

inercia:


salt/orch/update-etc-hosts.sls
```diff
@@ -27,17 +30,27 @@ etc_hosts_setup:
...
+{% endif %}
+
+{%- if salt.saltutil.runner('mine.get', tgt=updates_master_target, fun='caasp_fqdn', tgt_type='compound')|length > 0 %}
+
 kube_master_setup:
   salt.state:
     - tgt: {{ updates_master_target }}
     - tgt_type: compound
-    - queue: True
+    - highstate: True
```
kiall: highstate: True? This will apply all states, extending the time of orch this significantly.

salt/orch/update-etc-hosts.sls
```diff
@@ -27,17 +30,27 @@ etc_hosts_setup:
   salt.state:
     - tgt: {{ updates_all_target }}
     - tgt_type: compound
-    - queue: True
+    # see https://github.com/saltstack/salt/issues/31562
+    - kwarg:
+        queue: True
```
kiall: Fixing this so `queue: True` is actually picked up may lead to issues. If we're going to merge this, we need to figure out how to validate this won't increase the likelihood of conflcits with the orch/kubernetes.sls or orch/update.sls

At the very least, we should test this change with a upgrade of 5 or 10 nodes - the upgrade orch hits conflicts with this one much more than bootstrap.

salt/orch/update-etc-hosts.sls
inercia: I think this was the original intention of this `salt.state`. Otherwise, what is this `kube_master_setup` for?

inercia: In fact, I've just discovered an error because of this:

```
                "salt_|-kube_master_setup_|-kube_master_setup_|-state": {
                    "__id__": "kube_master_setup",
                    "__run_num__": 4,
                    "changes": {},
                    "comment": "No highstate or sls specified, no execution made",
                    "duration": 0.518,
                    "name": "kube_master_setup",
                    "result": false,
                    "start_time": "14:02:13.237097"
                },
```

so `kube_master_setup` is doing nothing.

PR #298, opened 2017-11-03
Minor certificates optimizations

inercia:
By embedding the private key in the `certificate_managed` Salt can skip some certificate regenerations.

We are trying to reduce the number of `x509.sign_remote_certificate` events in the master: this reduced the number from 45 to 32 in a cluster with 1 admin, 1 master and 1 node, but it is still too high...

### DIFF
--- a/salt/_macros/certs.jinja
+++ b/salt/_macros/certs.jinja
@@ -49,6 +49,7 @@
 
 {{ crt }}:
   x509.certificate_managed:
+    - overwrite: False
     - ca_server: {{ salt['mine.get']('roles:ca', 'ca.crt', expr_form='grain').keys()[0] }}
     - signing_policy: minion
   {%- if cn|length > 0 %}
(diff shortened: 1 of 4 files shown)


inercia:
> LGTM - As far as reducing the number of calls even more - how many unique certs are created for that cluster? We'll not be able to drop below that (e.g. we shouldn't start sharing cert/key files across purposes..)

The problem is that, even when we are generating certificates for each service only once, we are requesting a certificate signature to the CA many more times: every time we `require` or `include` a `crt`/`key`, Salt will check if we currently have the certificate and it will think _"yeah, the certificate is there, but let's [generate a new certificate](https://github.com/saltstack/salt/blob/develop/salt/states/x509.py#L493) with all those `CN`, `subjectAltName` and so in the state and I will [send it to the CA](https://github.com/saltstack/salt/blob/develop/salt/modules/x509.py#L1381) so I can compare both"_. So Salt is spamming the CA with `sign_remote_certificate`s for calculating the `changed` of certificates.

Possible solutions to this problem:

* to implement some kind of _memoization_ on the CA so that `sign_remote_certificate` could use cached values. This would implement some changes in the Salt code.
* to implement some early comparison in the minions... I'm trying something: I set a serial number for every orchestration and I'll use that serial for the certificates. If the certificate is on disk and it has the same serial as the serial for the current orchestration, `changes=False`.

inercia:
On hold until we find a better way to skip some certificates regeneration...

PR #299, opened 2017-11-06
[2.0] Use the default network interface instead of the hardcoded 'eth0' (bsc#1058079)

inercia:
Backport of https://github.com/kubic-project/salt/pull/250

### DIFF
--- a/pillar/beacons.sls
+++ b/pillar/beacons.sls
@@ -2,6 +2,7 @@ beacons:
   network_settings:
     interval: 5
     # monitorize any change in IP addresses in eth0
+    # TODO: make this interface configurable
     eth0:
       ipaddr:
 
--- a/salt/_grains/gateway.py
+++ b/salt/_grains/gateway.py
@@ -1,0 +1,16 @@
+import socket, struct
+
+def gateway():
+    """Get the default gateway (and device)."""
+
+    with open("/proc/net/route") as fh:
+        for line in fh:
+            fields = line.strip().split()
+            if fields[1] != '00000000' or not int(fields[3], 16) & 2:
+                continue
+
+            # Some code for logic that sets grains like
+            grains = {}
+            grains['gateway_iface'] = fields[0]
+            grains['gateway_ip'] = socket.inet_ntoa(struct.pack("<L", int(fields[2], 16)))
+            return grains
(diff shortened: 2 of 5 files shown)


kiall:
Humm - I'm not sure if we should merge this, or not... Given it's only a partial fix, and doesn't actually allow everything to work 100% correctly (the TODO in the code) makes me wonder if we should merge to 2.0...

inercia:
> Humm - I'm not sure if we should merge this, or not... Given it's only a partial fix, and doesn't actually allow everything to work 100% correctly (the TODO in the code) makes me wonder if we should merge to 2.0...

@flavio, what do you think?

pillar/beacons.sls
```diff
@@ -2,6 +2,7 @@ beacons:
   network_settings:
     interval: 5
     # monitorize any change in IP addresses in eth0
+    # TODO: make this interface configurable
```
ereslibre: Based on the documentation, couldn't we use `eth*` and other possible names for network cards? (https://docs.saltstack.com/en/latest/ref/beacons/all/salt.beacons.network_settings.html#salt.beacons.network_settings.beacon)

inercia: We could just use `*`, but then any IP address change in any interface would lead to a `update-etc-hosts` orchestration.

The only solution I see is to implement our own beacon, but that would require some work...

flavio:
It's too late to get this fix into 2.0. Let's get everything sorted out and make it part of the 1st maintenance update.

BTW, I think we have to resume the discussion on bugzilla and take into account the use case Rob Knight brought in.

ereslibre:
Oh this is the backport, it's OK as it was in `master` until we find a better way for the beacons (we might use wildcards with the most common names and only listen for the default interface one if that's possible).

Since this is a P2 I don't think we should merge this on 2.0 as we have discussed in the standup. We can include this as a 2.0 upgrade.

inercia:
> BTW, I think we have to resume the discussion on bugzilla and take into account the use case Rob Knight brought in.

I think Rob was describing a workaround more than a use case... The real solution to the eth0 problem will come in #301 (and some other PR for the reactor case).

I will close this PR in favour of #301.

PR #301, opened 2017-11-08
Remove all the remaining references of eth0 (bsc#1058079)

inercia:
Remove the previous custom grain by using some (previously unknown) `network.default_route`
Simplify things by using some Jinja macros.

NOTE: the case for the hardcoded `eth0` in the reactor will be treated in a separate PR (maybe not necessary if we add the requirement for a working DNS)


inercia:
@flavio I guess this should be backported to 2.0, right?

PR #302, opened 2017-11-09
only set service entries for localhost on kube-master

MaximilianMeister:
(bsc#1067219)

### DIFF
--- a/salt/etc-hosts/hosts.jinja
+++ b/salt/etc-hosts/hosts.jinja
@@ -1,5 +1,11 @@
 ### service names ###
+# set the apiserver for 127.0.0.1 on all hosts as haproxy is listening on all nodes
+# and forwarding connections to the real master
+{% if "kube-master" in salt['grains.get']('roles', []) %}
 127.0.0.1 api api.{{ pillar['internal_infra_domain'] }} {{ pillar['api']['server']['external_fqdn'] }}
+{% else %}
+127.0.0.1 api api.{{ pillar['internal_infra_domain'] }}
+{% endif %}
 
 ### admin nodes ###
 {%- set admins = salt['mine.get']('roles:admin', 'network.ip_addrs', 'grain') %}


MaximilianMeister:
should we also explicitly set a loopback for admin and workers?

salt/etc-hosts/hosts.jinja
```diff
@@ -1,5 +1,7 @@
 ### service names ###
+{% if "kube-master" in salt['grains.get']('roles', []) %}
```
inercia: As far as I know, the API server is configured in 127.0.0.1 in all the machines in the cluster because we have haproxy listening in localhost and it will forward connections to the real API server. So I think this `127.0.0.1 api` entry should be present in all the machines...

inercia: I think we should remove the last part of the entry, so it should just be `127.0.0.1 api api.{{ pillar['internal_infra_domain'] }}` (but my guts tell me there will be some reason for this related to certificates). Maybe @kiall knows about this...

inercia: I think we could only include the FQDN alias in the master by using something like:

```language:SaltStack
127.0.0.1 api api.{{ pillar['internal_infra_domain'] }}
{%- if "kube-master" in salt['grains.get']('roles', []) -%}
{{- pillar['api']['server']['external_fqdn'] -}} {#- this should be appended to the previous line -#}
{%- endif %}
```
So we would get something like `127.0.0.1 api api.infra.caasp.local c166.arch.suse.de` in the master but `127.0.0.1 api api.infra.caasp.local` in the rest of the cluster.

MaximilianMeister:
[Bug 1067219 - Master-Node FQDN ends up as Admin-Node loopback entry in /etc/hosts](https://trello.com/c/hANyUXIj/453-bug-1067219-master-node-fqdn-ends-up-as-admin-node-loopback-entry-in-etc-hosts)

kiall:
I believe the external FQDN entry is only needed by Velum (and caasp-cli, if used). Which means it must be on the Admin node.

inercia:
Shouldn't we avoid to set `127.0.0.1 api external_fqdn` in the `admin`? There is not haproxy running in the `admin` node, so it would resolve to a dead end...

ereslibre:
> Shouldn't we avoid to set 127.0.0.1 api external_fqdn in the admin? There is not haproxy running in the admin node, so it would resolve to a dead end...

I *think* this is fine because salt will actually deploy `haproxy` on the `admin` node too: https://github.com/kubic-project/salt/blob/f74c756f7cddac7307e0aab8aa549621c2d443d9/salt/top.sls#L12

inercia:
> I think this is fine because salt will actually deploy haproxy on the admin node too:

It seems you are right @ereslibre

kiall:
@inercia  I'm nearly certain this will break Velum+Dex integration unless the entry is left on the admin node - without this, Velum can't resolve the name it uses to access Dex - see my comment above.

ereslibre:
@kiall Just tested the patch. We should not have merged this without the two approvals, but I can confirm it works fine. I just bootstrapped a cluster with this patch. It succeeded. Also, I can download the kubeconfig file just fine.

kiall:
Looking closer, the interaction from velum uses the internal name - https://github.com/kubic-project/velum/blob/master/app/controllers/oidc_controller.rb#L122-L124 not the external one - I thought that too used the external name the same as kubernetes does.

PR #303, opened 2017-11-10
Disable container-feeder before rebooting

ereslibre:
This will allow us to control when container-feeder starts to load new
images from the filesystem. Due to some possible docker configuration changes
it might be restarted while container-feeder is working (if we keep it
enabled). Force to disable the service before rebooting.

Fixes: bsc#1066653

### DIFF
--- a/salt/container-feeder/stop.sls
+++ b/salt/container-feeder/stop.sls
@@ -1,0 +1,4 @@
+# Stop and disable container-feeder
+container-feeder:
+  service.dead:
+    - enable: False
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -45,6 +45,7 @@ update_modules:
   salt.state:
     - tgt: {{ master_id }}
     - sls:
+      - container-feeder.stop
       - kube-apiserver.stop
       - kube-controller-manager.stop
       - kube-scheduler.stop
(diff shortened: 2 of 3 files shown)


salt/container-feeder/stop.sls
```diff
@@ -0,0 +1,4 @@
+# Stop and disable container-feeder
+container-feeder:
```
inercia: I don't think we really need a new state: we can just use `docker/stop.sls`:

```
# docker/stop.sls
# Stop and disable the docker daemon
docker:
  service.dead:
    - enable: False

container-feeder:
  service.dead:
    - enable: False
```

salt/orch/update.sls
```diff
@@ -125,10 +126,11 @@ update_modules:
     - tgt: {{ worker_id }}
     - sls:
       - kubelet.stop
-      - kube-proxy.stop      
+      - kube-proxy.stop
       - docker.stop
       - flannel.stop
       - etcd.stop
+      - container-feeder.stop
```
inercia: I would stop `container-feeder` before `docker`, just in case...

ereslibre: This is on an already running machine, so container-feeder should not be running, but I will update the PR, makes sense.

inercia:
(minor: the commit should also include the `bsc#`)

ereslibre:
> (minor: the commit should also include the bsc#)

It does, from the very beginning :)

PR #308, opened 2017-11-10
WIP: Generate sa key in the update orchestration

flavio:
Reintroduced fix from PR https://github.com/kubic-project/salt/pull/304 that was accidentally reverted.


flavio:
wrong PR

PR #310, opened 2017-11-21
Include `Internal Dashboard FQDN/IP` value in the LDAP certificate

ereslibre:
Since Dex will connect to LDAP using this FQDN/IP, make sure that the
TLS handshake will succeed by regenerating the certificate early in
the orchestration, so it includes this FQDN/IP in the SAN extensions
of the LDAP certificate.

Fixes: bsc#1069175

### DIFF
--- a/pillar/certificates.sls
+++ b/pillar/certificates.sls
@@ -27,6 +27,9 @@ ssl:
   velum_key: '/etc/pki/private/velum.key'
   velum_crt: '/etc/pki/velum.crt'
 
+  ldap_key: '/etc/pki/private/ldap.key'
+  ldap_crt: '/etc/pki/ldap.crt'
+
   dex_key: '/etc/pki/dex.key'
   dex_crt: '/etc/pki/dex.crt'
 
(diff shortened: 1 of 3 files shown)


vitoravelino:
[Bug 1069175 - Auth dialog when downloading kubectl errors out](https://trello.com/c/FNRjsUvU/458-bug-1069175-auth-dialog-when-downloading-kubectl-errors-out)

ereslibre:
How I validated the change:

Internal Dashboard FQDN/IP: `10.17.1.0.nip.io`

Without this patch, LDAP will contain an invalid certificate when dex tries to connect to it using TLS and the internal dashboard reference: `10.17.1.0.nip.io:389`, as the certificate will contain all ip addresses of the machine along with all detected hostnames in the machine (`10.17.1.0.nip.io` is not among them).

With this patch, I can properly go to Dex, and login, downloading the kubeconfig file.

flavio:
I think this should go be backported to 2.0 and released with a maintenance update too.

PR #312, opened 2017-11-22
Sync _pillar modules only

ereslibre:
We want to sync the pillars on the master first.

### DIFF
--- a/salt/orch/kubernetes.sls
+++ b/salt/orch/kubernetes.sls
@@ -7,10 +7,9 @@ set_bootstrap_in_progress_flag:
       - bootstrap_in_progress
       - true
 
-sync_all:
-  module.run:
-    - name: saltutil.sync_all
-    - refresh: True
+sync_pillar:
+  salt.runner:
+    - name: saltutil.sync_pillar
 
 disable_rebootmgr:
   salt.state:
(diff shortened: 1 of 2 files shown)


inercia:
@ereslibre Can you provide some details on the difference between both? I mean, is this really solving any problem?

ereslibre:
> @ereslibre Can you provide some details on the difference between both? I mean, is this really solving any problem?

So the problem is that salt is not auto discovering the `velum.py` pillar module in the `_pillar` directory. For that, and for syncing everything now that we were at it I added the `sync_all` call, but that call looks to be pretty expensive (and probably because of the low number of worker processes in puma) the bootstrap orchestration was failing with 3 masters + 2 workers.

The `sync_pillar` call is clearly lighter than `sync_all` and will only sync the modules in the `_pillar` directory (is what we want after all, the salt-master to discover the pillar module), and I wanted to avoid messing with the extensions directory configuration on the salt-master.

ereslibre:
https://docs.saltstack.com/en/latest/ref/runners/all/salt.runners.saltutil.html#salt.runners.saltutil.sync_pillar

ereslibre:
The modified code was introduced in https://github.com/kubic-project/salt/pull/238 and was clearly overkill. We only want the `velum.py` ext_pillar to be discovered by the master at this point in the orchestration, not doing a `sync_all` that will sync all modules with all minions.

PR #313, opened 2017-11-23
Registries and mirrors (bsc#1039877)

inercia:
(This is basically #119 revamped, and prepared for insecure registries as implemented in #129)

A new pillar structure is introduced where registroes can be specified in the followig way:

```yaml
 registries:
     - url: my-reg1.com:5000
     - url: my-reg2.com:5000
       cert: |
           # some certificate         
     - url: some.registry.com:5000
       mirrors:
        - url: http://local.mirror.lan:5000
          cert: |
              # the certificate for the mirror
```

I've followed the instruction found [here](https://docs.docker.com/registry/insecure/#using-self-signed-certificates) for adding customers' certificates to Docker so they can pull from other registries.

## How to test this

* download [the caaspctl script](https://raw.githubusercontent.com/inercia/caasp-tf/master/resources/common/caaspctl) and copy it to `/root` in the Admin Node
* `ssh admin-node` and `./caaspctl registry certs` and `./caaspctl registry start`. That will generate the certificates in the `/root/certificates.local` and start the registry.
* Import the `alpine` image to the registry with `./caaspctl registry import alpine`
* edit the `/usr/share/salt/kubernetes/pillar/certificates.sls` in the Admin Node and add the `/root/certificates.local/ca.pem` certificate as:
    ```yaml
      registries:
          - url: dashboard:5000
            cert: |
                 ---- BEGIN CERTIFICATE -----
           ...
    ```
* orchestrate
* get the Admin Node's `IP` and `ssh node1` and `echo "IP dashboard" > /etc/hosts`
* try to pull the `alpine` image from the registry with `docker pull dashboard:5000/alpine`. After that, you should see the `dashboard:5000/alpine` image when doing `docker images`.

### DIFF
--- a/salt/docker/init.sls
+++ b/salt/docker/init.sls
@@ -3,6 +3,61 @@ include:
   - flannel
 
 ######################
+# additional ca.crt(s)
+#######################
+
+{% set registries = salt['pillar.get']('docker:registries', []) %}
+{% for registry in registries %}
+  {% set cert = registry.get("cert", "") %}
+  {% if cert|length > 0 -%}
+    {% set host_port = registry.get("name") %}
+
+/etc/docker/certs.d/{{ host_port }}/ca.crt:
+  file.managed:
+    - makedirs: True
+    - contents: |
+        {{ cert | indent(8) }}
+
+  # When using the standar port (443), Docker is not very smart:
+  # if the user introduces "my-registry:443" as a trusted registry,
+  # we must also create the "ca.crt" for "my-registry"
+  # as he/she could just access "docker pull my-registry/some/image",
+  # and Docker would fail to find "my-registry/ca.crt"
+    {% set host_port_lst = host_port.split(':') %}
+    {% if host_port_lst|length > 1 %}
+      {% set host = host_port_lst[0] %}
+      {% set port = host_port_lst[1] %}
+      {% if port == '443' %}
+/etc/docker/certs.d/{{ host }}/ca.crt:
+  file.symlink:
+    - target: /etc/docker/certs.d/{{ host_port }}/ca.crt
+    - force: True
+    - makedirs: True
+    - require:
+      - file: /etc/docker/certs.d/{{ host_port }}/ca.crt
+      {% endif %}
+    {% else %}
+  # the same happens if the user introduced a certificate for
+  # "my-registry": we must fix the "docker pull my-registry:443/some/image" case.
+/etc/docker/certs.d/{{ host_port }}:443/ca.crt:
+  file.symlink:
+    - target: /etc/docker/certs.d/{{ host_port }}/ca.crt
+    - force: True
+    - makedirs: True
+    - require:
+      - file: /etc/docker/certs.d/{{ host_port }}/ca.crt
+    {% endif %}
+  {% endif %}
+{% endfor %}
+
+# Notes:
+# - from https://docs.docker.com/registry/insecure/#using-self-signed-certificates
+#   we do not need to restart docker after adding/removing certificates
+# - after a certificate is removed from the pillar by the user, the certifcate
+#   will remain there. Maybe we should consider to wipe the certificates
+#   directory if we are the only ones managing them...
+
+######################
 # proxy for the daemon
 #######################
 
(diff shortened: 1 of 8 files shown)


MaximilianMeister:
@inercia as i am currently reworking the insecure registries would it make sense to have the `docker` pillar definition in one pillar file? and more docker component specific? sth like

```sls
docker:
  daemon:
    # this mirrors the structure in /etc/docker/daemon.json (plus additional registry mapping)
    iptables: 'false',
    registries: [],
    log_level: 'warn'
```

the args will be dropped anyways.

and then we can detach the registry config (names/certs) into a `pillar/registries.sls` (as the registry is also detached from docker itself)

```sls
{% set docker_config_registries = salt.caasp_pillar.get('docker:daemon:registries') %}
registries:
{% for registry in docker_config_registries.values() %}
  name: {{ registry.get('prefix') }},
  {% if registry.get('cert') %}
  cert: |
    {{ registry.get('cert') }}
  {% endif %}
{% endfor %}
```

it might also make sense to do this in one PR as it's so related. what do you think?

inercia:
@MaximilianMeister I like the idea of having a `docker:daemon:` that mimics the contents of `/etc/docker/daemon.json`. _registries_ are entities that are not specific to docker, so I think they shoud be a top-level configuration parameter too. I think you are suggesting to cross-reference values from both files, but I'm afraid you cannot fill pillar values with the contents of other pillars. So maybe we could have (this is basically your proposal):

```yaml
# pillar/docker.sls
docker:
  daemon:
    # this mirrors the structure in /etc/docker/daemon.json
    iptables: 'false'
    log_level: 'warn'

# pillar/registries.sls
registries:
   - name: my-reg1.com:5000
   - name: my-reg2.com:5000
     cert: |
         # some certificate         
   - name: my-mirror3.com:5000
     mirror: true
```

MaximilianMeister:
> I like the idea of having a docker:daemon: that mimics the contents of /etc/docker/daemon.json. registries are entities that are not specific to docker, so I think they shoud be a top-level configuration parameter too. I think you are suggesting to cross-reference values from both files, but I'm afraid you cannot fill pillar values with the contents of other pillars.

@inercia ok, thanks for the hint, it's a bit bad as the `daemon.json` needs to contain a `registry` structure like [this](https://github.com/vrothberg/moby/commit/0fc6eacd07ef749632517894a81cca97211ca6e2) which would contain the registry names (prefix)... can we reference that somehow? otherwise we need to save one value (registry name) in 2 pillars.

salt/docker/init.sls
```diff
@@ -29,13 +84,31 @@ include:
...
+{
+  "iptables": "{{ docker_daemon_config.get('iptables') }}",
+  "log_level": "{{ docker_daemon_config.get('log_level') }}",
+  "registries": [
+  {% for registry in registries -%}
+    {% set mirrors = registry.get('mirrors') -%}
+    {
+      "Prefix": "{{ registry.get('name') }}",
+      "Mirrors": [
+        {% for mirror in mirrors if mirrors -%}
```
inercia: This could be confusing, right?

salt/docker/init.sls
```diff
@@ -29,13 +84,31 @@ include:
...
+
+/etc/docker/daemon.json:
+  file.managed:
+    - contents: |
+{
+  "iptables": "{{ docker_daemon_config.get('iptables') }}",
+  "log_level": "{{ docker_daemon_config.get('log_level') }}",
+  "registries": [
+  {% for registry in registries -%}
+    {% set mirrors = registry.get('mirrors') -%}
```
inercia: I thing it should be just `"mirror"`

inercia: I would do something like `{% set is_mirror = registry.get('mirror') -%}`

salt/docker/init.sls
inercia: I thinkn it would look better with a variable like `is_mirror`...

salt/docker/init.sls
inercia: And would it need a default value? Something like `{% set is_mirror = registry.get('mirror', False) -%}`

salt/docker/init.sls
```diff
@@ -2,6 +2,61 @@ include:
   - repositories
   - flannel
 
+######################
+# additional ca.crt(s)
+#######################
+
+{% set registries = salt['pillar.get']('docker:registries', []) %}
```
inercia: This would need to be updated to `{% set registries = salt['pillar.get']('registries', []) %}`

inercia:
Related to kubic-project/velum#193

MaximilianMeister:
[salt: configure the docker engine to use on-premise registry mirrors](https://trello.com/c/skVwgvED/478-salt-configure-the-docker-engine-to-use-on-premise-registry-mirrors)

flavio:
We need to backport that to 2.0

PR #314, opened 2017-11-23
Preserve haproxy configurations for Velum

ereslibre:
- [x] Regenerate `velum.pem` when `velum.crt` or `velum.key` is updated.
- [x] Add ACL's for HAProxy

* Handle `haproxy` configuration.
* Generate `pem` certificates, that include the certificate and private key.
* Remove `velum` container restart.

Depends on: https://github.com/kubic-project/velum/pull/387
Depends on: https://github.com/kubic-project/caasp-container-manifests/pull/134


salt/haproxy/haproxy.manifest
```diff
@@ -24,7 +24,23 @@ spec:
...
     - name: haproxy-cfg
       hostPath:
         path: /etc/haproxy
+{%- if "admin" in salt['grains.get']('roles', []) -%}
+    - name: velum-bundle-certificate
+      hostPath:
+        path: /etc/pki/velum.pem
+    - name: ca-certificate
+      hostPath:
+        path: /etc/pki/ca.crt
```
inercia: Couldn't we just mount the `/etc/pki` directory? Mounting individual files is always problematic (the container will not see the change after a rename/rewrite/etc)

kiall: If we mount the whole /etc/pki folder, then the container has access to all certificates. But - you are right - mounting files is problematic - we should probably re-evaluate where and how we store these files... something like this maybe?

* /etc/caasp/pki/haproxy/ca.crt (symlink if possible?)
* /etc/caasp/pki/haproxy/haproxy.crt
* /etc/caasp/pki/haproxy/haproxy.key
* /etc/caasp/pki/haproxy/haproxy-bundle.pem

But - that's a different issue..

ereslibre: We could do the refactor you say @kiall, however the problem with bind mounts happens when the file is replaced in an automatic way. I don't know if this is the case at the moment with the certificates, something we can check. The downside is that even if it works, if salt on the future makes this replacement atomic we might be in a bad situation without even knowing it.

I think @kiall's suggestion is a good approach (if the symlink works -- even if it doesn't we could copy the CA cert with salt on each directory). If it's not much trouble, I'd propose to do that in a later refactor, as this changes are intrusive enough themselves...

inercia: Why don't we create a certificates dir per service? `/etc/pki/haproxy` and so? With `cert.key` and `cert.crt` inside?

inercia: (By the way, I think we should apply the same philosophy to all the files that are mounted in the manifests in the `containers-manifests` repo...)

ereslibre:
How to test:

- Checkout all PR's (this one, velum and caasp-container-manifests)
- After all machines have been created, ssh into the admin node and run:
  - `cp /usr/share/caasp-container-manifests/config/haproxy/haproxy.cfg /etc/haproxy/`
  - Wait for `haproxy` container to be up
- Bootstrap normally

Bootstrap 3+2:

screenshot from 2017-11-25 00-22-06

Bootstrap 3+5:

screenshot from 2017-11-25 00-50-51

kiall:
Force merging the 3 PRs in together.

PR #317, opened 2017-11-28
helm should detect salt-installed tiller service

SlickNik:
The helm client looks for a tiller deployment called 'tiller-deploy' to
establish if tiller is already installed in the cluster, or not. Update
our salt install of tiller to use a deployment with the same name so
that it will be recognized by the helm client as already being
installed.

Fixes: bsc#1066201

### DIFF
--- a/salt/_macros/kubectl.jinja
+++ b/salt/_macros/kubectl.jinja
@@ -39,6 +39,10 @@
     - check_cmd:
        - {{ kwargs['check_cmd'] }}
 {%- endif %}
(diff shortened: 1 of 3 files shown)


SlickNik:
Funny you should mention this -- I was about to merge this yesterday when the same thought crossed my mind (after my discussion with Hart on IRC). I'm working on some clean-up code to clean up the old deployment as well. Let me mark this as WIP until then. Thanks!

PR #320, opened 2017-11-28
Run some things in only one master instead of in all of them

inercia:
We are using `tgt:kube-master` as the target for some states in our orchestration, when we could/should run those states in only one of the masters. This PR uses a `super_master` (the first master in the list of masters) as the `tgt:<super_master>`.


inercia:
> Oh, neat. However, rather than random, can we make a consistent choice? Things like where the latest helm.yaml file etc is would be easier on operators if they knew to look in a specific place (I know the SE's have been looking for and needing things like the dex and helm manifest YAML files, so I'm guessing theres some sort of need for knowing where they are)

Then we could install `helm.yaml` at `masters|first`, and this orchestration could follow the same scheme...

Anyway, I thougth our plan was to use the [kubernetes module](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.kubernetes.html) instead of keeping individual yaml files and dealing with `kubectl` idiosincracies...

kiall:
> Then we could install helm.yaml at masters|first, and this orchestration could follow the same scheme...

That sounds good to me

> Anyway, I thougth our plan was to use the kubernetes module instead of keeping individual yaml files and dealing with kubectl idiosyncrasies...

Yea, we should remove those shell outs to kubectl.. But there will always be *something* somewhere that needs to run once, so having the pattern available is nice!

salt/top.sls
```diff
@@ -33,7 +33,3 @@ base:
     - kube-apiserver
     - kube-controller-manager
     - kube-scheduler
-    - addons
-    - addons.dns
-    - addons.tiller
-    - dex
```
kiall: We shoudn't be removing these from the highstate as the highstate should apply all states related to the node, the orchestration is purely about giving some ordering to allow for first time setup etc.

inercia:
Ok, after my latest `push -f`, the orchestration uses a `super_master` (currently the first master) as the master where we do all these things: setup flannel, etcd discoery, load the dex/kubedns/etc manifests, etc...

salt/top.sls
kiall: Oh. I see why you're doing this. Right - these don't actually "belong" to a master, they belong to the cluster as a whole. Humm. Fair enough :)

inercia: I think we had this discussion before. These things are not intrinsic to the Masters: they are cluster services. Making them something tighted to a master leads to this problem where we try to install them multiple times.

PR #321, opened 2017-11-28
Fix use of no HTTP proxies with the new Pillar API

kiall:
With the new Pillar API, velum always returns all pillar keys, so we can
no longer rely on the absence of a pillar key to disable proxies.


kiall:
See https://github.com/kubic-project/salt/pull/322/files

PR #324, opened 2017-11-29
Remove empty state from etc-hosts orch

kiall:
The final state in the etc-hosts orch was not actually calling
anything, and hasn't been for quite a while. Lets remove it, so
that the error it logs can be finally be gone!

### DIFF
--- a/salt/orch/update-etc-hosts.sls
+++ b/salt/orch/update-etc-hosts.sls
@@ -1,5 +1,4 @@
 {%- set updates_all_target = 'P@roles:(admin|kube-(master|minion)) and G@bootstrap_complete:true and not G@bootstrap_in_progress:true and not G@update_in_progress:true' %}
-{%- set updates_master_target = 'G@roles:kube-master and G@bootstrap_complete:true and not G@bootstrap_in_progress:true and not G@update_in_progress:true' %}
 
 {%- if salt.saltutil.runner('mine.get', tgt=updates_all_target, fun='caasp_fqdn', tgt_type='compound')|length > 0 %}
 update_pillar:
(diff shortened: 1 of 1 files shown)


ereslibre:
> I think that, at some moment, there was a highstate: True in that kube_master_setup but it was removed accidentally. The question is, would it still make sense to apply a highstate on the masters when a node appears/disappers? What if that node is another master?

No, this was what we had previously:

```patch
commit c4eef4dc30d3b83d4608edcf4773398e1f91d753
Author: Robert Roland <robert.roland@suse.com>
Date:   Fri Aug 4 14:20:42 2017 -0700

    eliminated the kubernetes-master formula
    
    the daemons are all separate now, so it's controlled by role membership
    in the top.sls file
    
    moved addons to a separate salt formula

diff --git a/salt/orch/update-etc-hosts.sls b/salt/orch/update-etc-hosts.sls
index ab47547..077c617 100644
--- a/salt/orch/update-etc-hosts.sls
+++ b/salt/orch/update-etc-hosts.sls
@@ -38,8 +38,6 @@ kube_master_setup:
     - tgt: {{ updates_master_target }}
     - tgt_type: compound
     - queue: True
-    - sls:
-      - kubernetes-master
     - require:
       - salt: etc_hosts_setup
 {% endif %}
```

I wonder why we were forcing this `sls` in the past to be run at the moment.

kiall:
> I think that, at some moment, there was a highstate: True in that kube_master_setup but it was removed accidentally. The question is, would it still make sense to apply a highstate on the masters when a node appears/disappers? What if that node is another master?

The history shows it used to call the kube-master state (rather than a whole highstate), though, I haven't been able to figure out why it might have been doing that.. My guess is so that it would regenerate any certs with the new IPs, but... We shouldn't be using detected IPs in the certs at all - they change and aren't needed - we use our own names, or user provided static names/IPs.

PR #327, opened 2017-12-11
Use a sanitized version of pillar.get

inercia:
if `pillar[x]` can return `None`, we can get a nasty `"None"` string when replacing `{{pillar[x]}}` in a Jinja template. And, in general, `None`s are rather annoying in Salt/Jinja...

So this PR implements our own version of `pillar.get` that sanitizes the output, never returning `None` and converting `"true"`/`"false"` to `True`/`False`.

### DIFF
--- a/salt/_macros/certs.jinja
+++ b/salt/_macros/certs.jinja
@@ -29,13 +29,13 @@
                        "kubernetes.default.svc",
                        "api",
                        "api." + pillar['internal_infra_domain'],
-                       salt['pillar.get']('api:server:external_fqdn', ''),
-                       salt['pillar.get']('api:cluster_ip', '')] +
-                       salt['pillar.get']('api:server:extra_names', []) +
-                       salt['pillar.get']('api:server:extra_ips', []) +
+                       salt.caasp_pillar.get('api:server:external_fqdn'),
+                       salt.caasp_pillar.get('api:cluster_ip')] +
+                       salt.caasp_pillar.get('api:server:extra_names', []) +
+                       salt.caasp_pillar.get('api:server:extra_ips', []) +
                        lst -%}
   {#- add some standard extra names from the DNS domain -#}
-  {%- if salt['pillar.get']('dns:domain') -%}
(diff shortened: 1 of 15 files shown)


salt/_macros/certs.jinja
```diff
@@ -29,13 +29,13 @@
                        "kubernetes.default.svc",
                        "api",
                        "api." + pillar['internal_infra_domain'],
-                       salt['pillar.get']('api:server:external_fqdn', ''),
-                       salt['pillar.get']('api:cluster_ip', '')] +
-                       salt['pillar.get']('api:server:extra_names', []) +
-                       salt['pillar.get']('api:server:extra_ips', []) +
+                       salt.caasp_pillar.get('api:server:external_fqdn'),
+                       salt.caasp_pillar.get('api:cluster_ip')] +
```
ereslibre: Typo? `]`

salt/_modules/caasp_pillar.py
```diff
@@ -0,0 +1,28 @@
...
+    * will return `True`/`False` instead of `"true"`/`"false"`
+
+    The rationale is that, if pillar[x] can return None,
+    we can get a nasty "None" when replacing {{pillar[x]}} in a
+    jinja template. In other words, `None`s are annoying when used
+    in Salt/Jinja...
+    '''
+    res = __salt__['pillar.get'](name, default)
+    if not res:
+        res = default
```
ereslibre: This will do a strange round trip if we have `False` (boolean type) as a pillar value for `name`, since `caasp_pillar.get` will return whatever the `default` value is instead of `False`.

inercia: Good catch. I will use another solution.

salt/_macros/certs.jinja
inercia: No, it is concatenating lists. For example, with default values, it will be `[... + salt.caasp_pillar.get('api:cluster_ip')] + [] + []`

ereslibre: Okay, the diff in github has a partial context. Thanks!

PR #328, opened 2017-12-13
Let flannel auto-calculate the Max and Min subnet

inercia:
`SubnetMin` and `SubnetMax` specify the first and last subnets that will be assigned to nodes in the cluster. Given the network CIDR, Flannel can calculate both automatically. Do not specify these parameters except when users really know wheta they are doing.

Anyway, I think both parameters will disappear once we move to CNI, as the kubernetes subnets manager is responsible for these things...

### DIFF
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -1,12 +1,25 @@
 # the CIDR for cluster IPs (internal IPs for Pods)
 cluster_cidr:     '172.16.0.0/13'
 
-# The min and max subnet, and length for node allocations.
-# The defaults allow for 510 pod IPs per host (/23) allowing
-# for a maximum of 1024 hosts.
-cluster_cidr_min: '172.16.0.0'
-cluster_cidr_max: '172.23.255.255'
+# The size of the subnet allocated to each host.
+# Defaults to 23 (i.e. /23) unless Network was configured to
+# be smaller than a /23 in which case it is one less than the
+# network.
+# The defaults allow for
+# * 2^(23-13)     = 10 bits for hosts = 1024 hosts
+# * 2^(32-23) - 2 = 9 bits - 2 for pods = 510 pod IPs per host,
+#   (with IPs from '172.16.0.1' to '172.16.1.254')
+# TODO: we currently have 19 bits for hosts|pods, so we should ask
+#       users in the UI about the "max number of hosts"
+#       showing all the powers of 2 (ie, 256, 512, 1024), and then
+#       calculate and show the max number of pods...
 cluster_cidr_len: 23
+
+# These are automatically calculated by Flannel or
+# k8s's Network Manager.
+# WARNING: do not uncomment these unless you really know what you are doing!
+# cluster_cidr_min: '172.16.0.0'
+# cluster_cidr_max: '172.23.254.0'
 
 # the CIDR for services (virtual IPs for services)
 services_cidr:    '172.24.0.0/16'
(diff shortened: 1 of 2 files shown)


ereslibre:
With this change we should also remove: https://github.com/kubic-project/velum/blob/master/app/models/pillar.rb#L21-L22, and all related code (like the Velum view that includes this two fields).

We should also run a migration to remove this values from the database, or might this be problematic if on specific installations there are somewhat exotic values?

inercia:
> With this change we should also remove: https://github.com/kubic-project/velum/blob/7a5f1fec9aed67afff35f535a774b9c7be31a24d/app/models/pillar.rb#L21-L22, and all related code (like the Velum view that includes this two fields).

Yes, we could remove those settings. I would also like to address [the changes suggested in the TODO](https://github.com/kubic-project/salt/pull/328/files#diff-1b366d73824e60734ecbabc4b7925343R10) and let users choose how to distribute the bits.

> We should also run a migration to remove this values from the database, or might this be problematic if on specific installations there are somewhat exotic values?

We could remove these values from the database if we want. Not a big deal to left them there though...

pillar/params.sls
```diff
@@ -1,13 +1,23 @@
...
-# for a maximum of 1024 hosts.
-cluster_cidr_min: '172.16.0.0'
-cluster_cidr_max: '172.23.255.255'
+# The size of the subnet allocated to each host.
+# Defaults to 24 (i.e. /24) unless Network was configured to be smaller than a /24 in which case it is one less than the network.
+# The defaults allow for
+# * 2^(23-13)     = 10 bits for hosts = 1024 hosts
+# * 2^(32-23) - 2 = 9 bits - 2 for pods = 510 pod IPs per host,
+#   (with IPs from '172.16.0.1' to '172.16.1.254')
+# TODO: we currebtly have 19 bits for hosts|pods, so we should ask
```
ereslibre: Minor: typo, `currently`.

pillar/params.sls
```diff
@@ -1,13 +1,23 @@
 # the CIDR for cluster IPs (internal IPs for Pods)
 cluster_cidr:     '172.16.0.0/13'
 
-# The min and max subnet, and length for node allocations.
-# The defaults allow for 510 pod IPs per host (/23) allowing
-# for a maximum of 1024 hosts.
-cluster_cidr_min: '172.16.0.0'
-cluster_cidr_max: '172.23.255.255'
+# The size of the subnet allocated to each host.
+# Defaults to 24 (i.e. /24) unless Network was configured to be smaller than a /24 in which case it is one less than the network.
```
ereslibre: Why it defaults to 24? `cluster_cidr_len` defaults to /23, no?

pillar/params.sls
```diff
@@ -1,13 +1,23 @@
...
+# TODO: we currebtly have 19 bits for hosts|pods, so we should ask
+#       users in the UI about the "max number of hosts"
+#       showing all the powers of 2 (ie, 256, 512, 1024), and then
+#       calculate and show the max number of pods...
 cluster_cidr_len: 23
 
+# These are automatically calculated by Flannel or
+# k8s's Network Manager
+# cluster_cidr_min: '172.16.0.0'
+# cluster_cidr_max: '172.23.254.0'
```
ereslibre: I suggest to set them to `None` by default instead of commented out. The effect will be the same but we are clearly stating that our default pillar values for them are to be autocalculated unless overriden.

pillar/params.sls
inercia: Typo

pillar/params.sls
inercia: I would prefer not to rely on how we deal with `None`s (and they would be the only `None`s in our pillar).

ereslibre: I'm a bit worried in general about commented code because you never know if it has a purpose or it was a leftover or what... that's why I think commented out code is in general not a good practice. With our latest way to access the pillar with `salt.caasp_pillar.get` a `None` should not do any harm here...

inercia: I think it is more confusing to leave this kind of behavior to some obscure mechanism of _how-we-deal-with-Nones_ than a comment. I will clearly state this is not a spurious comment...

ereslibre:
Rebased on top of master so it includes the correct `caasp-dex` version.

PR #331, opened 2017-12-15
Add beacon to notify network changes only on the default network interface

ereslibre:
Fixes: bsc#1063709


ereslibre:
Rebased so CI is green again here.

PR #335, opened 2017-12-21
Explicitly pass the kubeconfig file to kubectl

inercia:
Do not try to pass the kubeconfig in the `KUBECONFIG` env var: it can be problematic. Just use the `--kubeconfig` argument of `kubectl`

### DIFF
--- a/salt/_macros/kubectl.jinja
+++ b/salt/_macros/kubectl.jinja
@@ -13,12 +13,10 @@
 
 {% macro _kubectl_run(args) -%}
   caasp_cmd.run:
-    - name: kubectl {{ args }}
+    - name: kubectl --kubeconfig={{ pillar['paths']['kubeconfig'] }} {{ args }}
     - retry:
         attempts: 10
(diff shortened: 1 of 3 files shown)


ereslibre:
I'm still evaluating the root cause of this problem. It *looks* like the envvar really exists on the `until` and `onlyif` contexts....

ereslibre:
The problem in this case is the subshell `$()`.

ereslibre:
Just to provide more context. In the update orchestration nodes were left in `SchedulingDisabled` status (all of them). We are doing a `drain` before rebooting them, but then the update orchestration despite being successful had:

```
                        ID: kubelet
                  Function: caasp_cmd.run
                      Name: kubectl uncordon 9e8ed6f15c754766ac64e6ee620890fe.infra.caasp.local
              
                    Result: True
                   Comment: onlyif execution failed
                   Started: 15:37:44.208830
                  Duration: 495.543 ms
                   Changes:
```

PR #338, opened 2017-12-26
Only uncordon nodes that were cordoned because of our own processes

ereslibre:
Fix kubelet highstate to uncordon the node only if we did cordon it by
one of our processes (like an update).

Without this patch, adding new nodes or performing an update would
uncordon all nodes unconditionally, without taking into account if a
user had a node cordoned for some reason (e.g. hardware failures or
other reasons). Do not uncordon those nodes, keep them cordoned.

Fixes: bsc#1050017

### DIFF
--- a/salt/kubelet/stop.sls
+++ b/salt/kubelet/stop.sls
@@ -4,6 +4,8 @@
 include:
   - kubectl-config
 
+{% set should_uncordon = salt['cmd.run']("kubectl --kubeconfig=" + pillar['paths']['kubeconfig'] + " get nodes " + grains['caasp_fqdn'] + " -o=jsonpath='{.spec.unschedulable}' 2>/dev/null") != "true" %}
+
 # If this fails we should ignore it and proceed anyway as Kubernetes will recover
 drain-kubelet:
   cmd.run:
@@ -13,6 +15,10 @@ drain-kubelet:
       - /bin/true
     - require:
       - file: {{ pillar['paths']['kubeconfig'] }}
+  grains.present:
+    - name: kubelet:should_uncordon
+    - value: {{ should_uncordon }}
+    - force: True
 
 kubelet:
   service.dead:
--- a/salt/kubelet/init.sls
+++ b/salt/kubelet/init.sls
@@ -81,7 +81,7 @@ kubelet:
     - require:
       - service: kubelet
 
-{% if not "kube-master" in salt['grains.get']('roles', []) %}
+{% if not "kube-master" in salt['grains.get']('roles', []) and salt['grains.get']('kubelet:should_uncordon', false) %}
   caasp_cmd.run:
     - name: |
         kubectl uncordon {{ grains['caasp_fqdn'] }}
@@ -93,6 +93,11 @@ kubelet:
     - require:
       - file: {{ pillar['paths']['kubeconfig'] }}
 {% endif %}
+
+remove-should-uncordon-grain:
+  grains.absent:
+    - name: kubelet:should_uncordon
+    - destructive: True
 
 #######################
 # config files


ereslibre:
How I tested it:

Deployed a cluster with 1 master and 3 workers. Cordoned one of the workers manually. Ran:

- `kubernetes` orchestration, the cordoned node is kept cordoned. This orchestration affects for example when adding a new node.

- `update` orchestration, the cordoned node is kept cordoned after the update has finished. The rest of the nodes are in `Ready` status.

salt/kubelet/stop.sls
```diff
@@ -4,6 +4,14 @@
 include:
   - kubectl-config
 
+{% if not "kube-master" in salt['grains.get']('roles', []) %}
+set-should-uncordon-grain:
+  grains.present:
+    - name: kubelet:should_uncordon
+    - value: {{ salt['cmd.run']("kubectl --kubeconfig=" + pillar['paths']['kubeconfig'] + " get nodes " + grains['caasp_fqdn'] + " -o=jsonpath='{.spec.unschedulable}' 2>/dev/null") != "true" }}
+    - force: True
+{% endif %}
```
inercia: See [this comment](https://github.com/kubic-project/salt/pull/339/files#r158726435)

salt/kubelet/stop.sls
```diff
@@ -4,6 +4,8 @@
 include:
   - kubectl-config
 
+{% set should_uncordon = salt['cmd.run']("kubectl --kubeconfig=" + pillar['paths']['kubeconfig'] + " get nodes " + grains['caasp_fqdn'] + " -o=jsonpath='{.spec.unschedulable}' 2>/dev/null") != "true" %}
+
 # If this fails we should ignore it and proceed anyway as Kubernetes will recover
 drain-kubelet:
```
inercia: Shouldn't we do this only if the node is currently _scheduable_?

ereslibre: Yes, that's what we are doing, `.spec.unschedulable != true`.

ereslibre: If you are talking about the drain specifically, we need to drain. A `unschedulable` node will not accept new workloads, but will happily keep workloads that were running before it was set as `unschedulable`. The drain is necessary always.

inercia: Then how can we detect if the node has already been drained?

ereslibre: There's no need to check if a node has been drained. You want to drain the node, drain will mark the node as unschedulable and evict all running pods. If it was drained before there will be no difference, this operation will succeed.

inercia: By the way, will this mark master nodes as _scheduable_ after the reboot? Because the are _unscheduable_, so this will `drain` them and mark as `should_uncordon=True`, and then they will be `undordon`ed after reboot, so I don't know if hey will be marked `scheduable`...

ereslibre: `kubelet.stop` will only be executed on the worker nodes. Also, we never ever uncordon master nodes (https://github.com/kubic-project/salt/blob/master/salt/kubelet/init.sls#L84)

inercia: > kubelet.stop will only be executed on the worker nodes. Also, we never ever uncordon master nodes (https://github.com/kubic-project/salt/blob/master/salt/kubelet/init.sls#L84)

Ok, makes sense.

Anyway, this will not skip the `cordon` when the node has already be `cordon`ed by users, but maybe not a big deal...

salt/kubelet/stop.sls
```diff
@@ -13,6 +15,10 @@ drain-kubelet:
       - /bin/true
     - require:
       - file: {{ pillar['paths']['kubeconfig'] }}
+  grains.present:
+    - name: kubelet:should_uncordon
+    - value: {{ should_uncordon }}
```
MaximilianMeister: why not just executing the statement here instead of caching it in a variable, afaics it's just used once

ereslibre: It's basically the same thing, but thought it would be a bit more readable. However I can inline it if you think it's better.

PR #339, opened 2017-12-26
Only uncordon nodes that were cordoned because of our own processes

ereslibre:
Fix kubelet highstate to uncordon the node only if we did cordon it by
one of our processes (like an update).

Without this patch, adding new nodes or performing an update would
uncordon all nodes unconditionally, without taking into account if a
user had a node cordoned for some reason (e.g. hardware failures or
other reasons). Do not uncordon those nodes, keep them cordoned.

Fixes: bsc#1050017

Backport of https://github.com/kubic-project/salt/pull/338

### DIFF
--- a/salt/kubelet/stop.sls
+++ b/salt/kubelet/stop.sls
@@ -4,6 +4,8 @@
 include:
   - kubectl-config
 
+{% set should_uncordon = salt['cmd.run']("kubectl --kubeconfig=" + pillar['paths']['kubeconfig'] + " get nodes " + grains['caasp_fqdn'] + " -o=jsonpath='{.spec.unschedulable}' 2>/dev/null") != "true" %}
+
 # If this fails we should ignore it and proceed anyway as Kubernetes will recover
 drain-kubelet:
   cmd.run:
@@ -15,6 +17,10 @@ drain-kubelet:
       - /bin/true
     - require:
       - file: {{ pillar['paths']['kubeconfig'] }}
+  grains.present:
+    - name: kubelet:should_uncordon
+    - value: {{ should_uncordon }}
+    - force: True
 
 kubelet:
   service.dead:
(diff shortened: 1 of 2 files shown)


salt/kubelet/stop.sls
```diff
@@ -4,6 +4,14 @@
...
 
+{% if not "kube-master" in salt['grains.get']('roles', []) %}
+set-should-uncordon-grain:
+  grains.present:
+    - name: kubelet:should_uncordon
+    - value: {{ salt['cmd.run']("kubectl --kubeconfig=" + pillar['paths']['kubeconfig'] + " get nodes " + grains['caasp_fqdn'] + " -o=jsonpath='{.spec.unschedulable}' 2>/dev/null") != "true" }}
+    - force: True
+{% endif %}
+
 # If this fails we should ignore it and proceed anyway as Kubernetes will recover
```
inercia: If `drain-kubelet` fails in the orchestration for some reason, `kubelet:should_uncordon` will be set anyway (because it was run before `drain-kubelet`). I would do instead:

```yaml

{{ set schedulable = salt['cmd.run']("kubectl --kubeconfig=" + pillar['paths']['kubeconfig'] + " get nodes " + grains['caasp_fqdn'] + " -o=jsonpath='{.spec.unschedulable}' 2>/dev/null") != "true" }} }}

drain-kubelet:
    cmd.run:
     - name: |
         kubectl drain {{ grains['caasp_fqdn'] }} --ignore-daemonsets --grace-period=300 --timeout=340s
     - env:
       - KUBECONFIG: {{ pillar['paths']['kubeconfig'] }}
     - check_cmd:
       - /bin/true
     - require:
       - file: {{ pillar['paths']['kubeconfig'] }}
   grains.present:
     - name: kubelet:should_uncordon
     - value: {{ scheduable }}
     - force: True
     - require:
       - cmd: drain-kubelet
```

ereslibre: Good catch, will fix.

ereslibre: Taking a look more in depth at your comment I will take the idea, but even if a node was cordoned we still want to drain it, since some workloads could be running there from before the cordoning.

flavio:
I think the code is aligned with what has been approved inside of #338. Am I right? If that's the case we should merge it too.

ereslibre:
@flavio Yes, it's a clean cherry pick.

flavio:
@inercia: can you approve the PR like you did with the other one?

PR #342, opened 2018-01-10
Remove discovered IP addresses from certs

kiall:
As the discovered IP addresses are not static, that we don't maintain that
the certs are updated+services are reloaded upon IP change, that we're
including all IPs - even 127.0.0.1 - in this list, and that we don't make use
of any of these SAN's, we should remove them.

### DIFF
--- a/salt/_macros/certs.jinja
+++ b/salt/_macros/certs.jinja
@@ -2,11 +2,6 @@
 {% macro alt_names(lst=[]) -%}
   {#- add all the names and IPs we know about -#}
   {%- set altNames = ["DNS: " + grains['caasp_fqdn'] ] -%}
-  {%- for _, interface_addresses in grains['ip4_interfaces'].items() -%}
-    {%- for interface_address in interface_addresses -%}
-      {%- do altNames.append("IP: " + interface_address) -%}
-    {%- endfor -%}
-  {%- endfor -%}
   {#- append all the names/IPs provided (if not empty) -#}
   {%- for name in lst -%}
     {%- if name and name|length > 0 -%}


kiall:
@inercia / @ereslibre - thoughts on this? Do either of you see any potential issues with this?

ereslibre:
Off the top of my head there's one impact: the `velum` certificate will be reissued with salt (https://github.com/kubic-project/salt/blob/de4888e548d4986bf2980261534efbb7b5b3681c/salt/velum/init.sls#L9-L14), and salt will access the pillar through `localhost` (https://github.com/kubic-project/salt/blob/de4888e548d4986bf2980261534efbb7b5b3681c/config/master.d/50-master.conf#L27). 

However, I see that this issue exists today too:

```
caasp-admin:~ # curl --cacert /etc/pki/ca.crt https://localhost/
curl: (51) SSL: no alternative certificate subject name matches target host name 'localhost'
```

```
caasp-admin:~ # openssl x509 -in /etc/pki/velum.crt -noout -text | grep -A1 "X509v3 Subject Alternative Name"
            X509v3 Subject Alternative Name: 
                DNS:caasp-admin, DNS:caasp-admin.devenv.caasp.suse.net, IP Address:10.17.1.0
```

So it looks like `salt` is ignoring this in a really unsafe way... we should fix this :(

There was also this on the apiserver:

```
--etcd-servers stringSlice                                List of etcd servers to connect with (scheme://ip:port), comma separated.
```

what I always understood as mandatory IP address, but we are using:

```
salt/kube-apiserver/apiserver.jinja
21:                   --etcd-servers=https://{{ grains['caasp_fqdn'] }}:2379"
```

and everything seems to work fine... So right now, I can't think of more potential problems...

PR #345, opened 2018-01-15
Add swap disabling

FedericoCeratto:
Unmount existing swaps; comment out swaps from fstab file.
No changes to swap partitions on disk. No reboot required.

### DIFF
--- a/salt/swap/init.sls
+++ b/salt/swap/init.sls
@@ -1,0 +1,14 @@
+coreutils:
+  pkg:
+    - installed
+
+unmount-swaps:
+  cmd.run:
+    - name: /sbin/swapoff -a
+
+remove-swap-from-fstab:
+  file.line:
+    - name: /etc/fstab
+    - content:
+    - match: ' swap '
+    - mode: delete
--- a/salt/top.sls
+++ b/salt/top.sls
@@ -5,6 +5,7 @@ base:
   'roles:(admin|kube-(master|minion))':
     - match: grain_pcre
     - hostname
+    - swap
     - etc-hosts
     - proxy
     - rebootmgr


salt/swap/init.sls
```diff
@@ -0,0 +1,12 @@
+coreutils:
+  pkg:
+    - installed
+
+unmount swaps:
+  cmd.run:
+    - name: /sbin/swapoff -a
+
+comment out swap from fstab:
```
inercia: I think we have been using `this-kind-of-salt-ids` instead of `this kind of salt ids`.

FedericoCeratto:
@inercia the swap disabling is expected to work independently of reboots and it's idempotent.

kiall:
@FedericoCeratto have we confirmed that, once swap has been disabled using this, that rebooting the node does not reactivate swap? There was a concern raised that systemd *might not* care about swap lines in /etc/fstab, and does things differently.

kiall:
Additionally, while this will apply to the admin node using salt. Swap must be disabled on the admin node prior to salt having a chance to run, as salt-master starts inside kubelet, which will fail due to swap being on.

Adding the `swapoff -a`+removal of /etc/fstab swap line into https://github.com/kubic-project/caasp-container-manifests/blob/master/activate.sh will allow swap to be disabled on the admin node, prior to kubelet starting.

Note that this file is ran in two very different environments - for ISO based installs, the script is ran inside installer / YaST environment in a chroot of the installed system, and then the node is rebooted into the installed system. For VM image based installs, this script is ran directly inside the installed system. This may mean swapoff -a fails on ISO based installs (or maybe not - hard to tell without testing!)

FedericoCeratto:
With the current kernel / systemd versions, on the admin node, the swap is being configured from /etc/fstab and "systemctl --type swap --all list-unit-files" is not showing it as a unit. Removing the line in /etc/fstab is enough. Later on we can also scan for systemd units and mask them for future compatibility.

mjura:
Should we backport it to `release-2.0` ?

kiall:
@mjura yes, once the second half over @ https://github.com/kubic-project/caasp-container-manifests/pull/140 is merged - we should backport to unblock the 1.8 update for 2.0.

PR #350, opened 2018-01-19
A very basic README on the file naming conventions

inercia:
Feel free to suggest any change or addition...

### DIFF
--- a/salt/README.md
+++ b/salt/README.md
@@ -1,0 +1,19 @@
+# State files
+
+The main state files are
+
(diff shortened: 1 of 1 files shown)


MaximilianMeister:
> I like this - the patterns aren't perfect but there way better than what we had before ;)
> 
> Can we:
> 
> Move this to the main repo README.md so it's way more obvious to people?
> Update with some concrete examples to make it easier to grok rather than just the pattern.

:+1:

PR #351, opened 2018-01-19
Node removal: first steps...

inercia:
Implement the node removal, trying to find a replacement and doing all the migrations that are necessary.

Machines with no role assigned will be used as replacement by default. 

Optionally, a replacement can be provided through the pillar (in the `replacement` variable).

# Integration

From velum, this orchestration muct be invoked with `pillar='{"target":"node-id"}`, where `node-id` is the node ID of the node that must be removed from the cluster. This would require to implement:
  * parameters in orchestrations 
  * some node removal elements in the user interface
  * some consistency checks in Velum

# Testing

Run in the salt container:

```
salt run -l debug --force-color \
  state.orchestrate "orch.removal" pillar='{"target":"4eace3637a7d4c3b93781b545b13032c"}'
```

replacing the machine ID with the ID of the node you want to remove...

You can force a specific replacement with

```
salt run -l debug --force-color \
  state.orchestrate "orch.removal" pillar='{"target":"4eace3637a7d4c3b93781b545b13032c", "replacement":"aa443a26a7d8483a8bbe52d74d91d18f"}'
```

# TODO

* [ ] certificates revocation and CRLs (probably in a different PR)


salt/orch/removal.sls
```diff
@@ -0,0 +1,146 @@
...
+      {%- endif %}
+      - docker.stop
+      - etcd.stop
+    - require:
+      - salt: update_grains_again
+
+# revoke certificates
+# TODO
+
+# revoke Salt keys
```
mjura: Hmmm I think that revoking salt keys should goes after reboot.

salt/orch/removal.sls
```diff
@@ -0,0 +1,146 @@
...
+      - kubelet.remove-pre-reboot
+      - kube-proxy.remove-pre-reboot
+      {%- endif %}
+      - docker.remove-pre-reboot
+      - cni.remove-pre-reboot
+      - etcd.remove-pre-reboot
+    - require:
+      - salt: stop_services
+
+# reboot the node
```
kiall: Should this be shutting down the node rather than rebooting it? The intent is that this machine will be wiped before it can be reused, so shutting down seems better IMO.

salt/orch/removal.sls
```diff
@@ -0,0 +1,146 @@
...
+# revoke Salt keys
+# TODO
+
+# remove any other configuration in the machines
+# TODO: all the files in /etc/kubernetes/*, /var/lib/kubernetes/* ??
+remove-pre-reboot:
+  salt.state:
+    - tgt: {{ target }}
+    - sls:
+      - container-feeder.remove-pre-reboot
```
kiall: Why are we removing the config files etc? Is this actually necessary given we won't accept a node back into the cluster until it's been wiped?

I would just use the various stop sls's to disable each service and leave the config's etc alone.

I'd worry we start trying (and failing - because it's all but impossible to do right) to have "undo" playbooks for everything we install.

salt/orch/removal.sls
kiall: IMO we shouldn't reboot - we should shutdown.. So the revokes probably belong here.

inercia: Yeah, we should probably shutdown...

salt/orch/removal.sls
inercia: Config files could leave sensitive information on disk (for example, if we do not add the node's keys to a CRL, the node could join again). And, in general, any file we leave could be a source of information for a possible attacker.

And it is not so hard to `rm -rf` some dirs and make sure everything is cleaned up...

mjura: Agree

salt/etcd/remove-pre-reboot.sls
```diff
@@ -0,0 +1,19 @@
...
+
+/etc/sysconfig/etcd:
+  file.absent
+
+/etc/sysconfig/etcdctl:
+  file.absent
+
+/etc/systemd/system/etcd.service.d/etcd.conf:
+  file.absent
+
```
mjura: Here is still missing:
```
/var/lib/etcd:
  file.absent
```

inercia: Fixed. Added in `etcd-wipe-var-lib`.

salt/orch/removal.sls
```diff
@@ -0,0 +1,151 @@
...
+      - kube-apiserver.remove-pre-reboot
+      - kube-controller-manager.remove-pre-reboot
+      - kube-scheduler.remove-pre-reboot
+      {%- else %}
+      - kubelet.remove-pre-reboot
+      - kube-proxy.remove-pre-reboot
+      {%- endif %}
+      - docker.remove-pre-reboot
+      - cni.remove-pre-reboot
+      - etcd.remove-pre-reboot
```
mjura: Please add also 
```
- etc-hosts.remove-pre-reboot
```
I've added it to https://github.com/kubic-project/salt/pull/369

salt/_macros/certs.jinja
```diff
@@ -88,6 +88,7 @@
     - ST: {{ pillar['certificate_information']['subject_properties']['ST']|yaml_dquote }}
     - basicConstraints: "critical CA:false"
     - keyUsage: nonRepudiation, digitalSignature, keyEncipherment
+    - extendedKeyUsage: serverAuth, clientAuth
```
kiall: This is making every cert a client cert, why are we doing this?

salt/orch/update-etc-hosts.sls
```diff
@@ -1,4 +1,4 @@
-{%- set updates_all_target = 'P@roles:(admin|kube-(master|minion)) and G@bootstrap_complete:true and not G@bootstrap_in_progress:true and not G@update_in_progress:true' %}
+{%- set updates_all_target = 'P@roles:(admin|kube-(master|minion)) and G@bootstrap_complete:true and not ( G@bootstrap_in_progress:true or G@update_in_progress:true or G@removal_in_progress:true )' %}
```
kiall: Unrelated to your PR: This is getting out of hand - we need to find a better solution.

inercia: Indeed. :-)

inercia: It can be _reduced_ to:

```
'P@roles:(admin|kube-(master|minion)) and G@bootstrap_complete:true and not P@*_in_progress:true'
```

salt/_macros/certs.jinja
inercia: Yeah, I will remove this just in case...

mjura:
Small nitpick, can we apply etc-hosts state module at the end `salt/orch/removal.sls` orchestration. This will clean old hostname data from nodes.

salt/etcd/init.sls
```diff
@@ -39,9 +59,9 @@ etcd:
     - proto: tcp
   caasp_service.running_stable:
     - name: etcd
-    - successful_retries_in_a_row: 50
-    - max_retries: 300
-    - delay_between_retries: 0.1
+    - successful_retries_in_a_row: 10
+    - max_retries: 30
+    - delay_between_retries: 1
```
ereslibre: Why we changed this? In a healthy cluster this will take 10 seconds at least per etcd instance, whereas before it was 5 seconds. Also, before we had a wider window to check if `etcd` was running in a stable manner.

salt/orch/removal.sls
```diff
@@ -0,0 +1,394 @@
...
+      - true
+
+{%- if replacement %}
+
+assign-addition-grain:
+  salt.function:
+    - tgt: {{ replacement }}
+    - name: grains.setval
+    - arg:
+      - addition_in_progress
```
ereslibre: `addition_in_progress` looks way too generic, could we rename this to something more specific, like `etcd_member_addition_in_progress` or something similar?

salt/orch/removal.sls
```diff
@@ -0,0 +1,394 @@
...
+    {%- if replacement in bootstrapped_masters %}
+      {%- do salt.caasp_log.warn('CaaS: error!! the replacement for an k8s master %s cannot be %s: another k8s master is already running there', target, replacement) %}
+      {%- if replacement_provided %}
+        {%- do salt.caasp_log.abort('CaaS: fatal!! the user provided replacement %s cannot be used', replacement) %}
+      {%- endif %}
+      {%- set replacement = '' %}
+    {%- elif replacement in minions %}
+      {%- do salt.caasp_log.warn('CaaS: warning!! will not replace the k8s master at %s: the replacement found/provided is the k8s minion %s', target, replacement) %}
+      {%- if replacement_provided %}
+        {%- do salt.caasp_log.abort('CaaS: fatal!! the user provided replacement %s cannot be used', replacement) %}
```
ereslibre: Do we want to allow to provide a replacement? I think for a first implementation it would make sense to avoid this, and stick to the easiest implementation possible, as this kind of edge cases add unnecessary complexity IMO, while nothing will call to the orchestration with a replacement (at least Velum won't).

salt/orch/removal.sls
inercia: It is used for any addition, not only etcd members.

salt/orch/removal.sls
inercia: We discussed in the planning meeting about this: a replacement could be provided, specially for masters.

salt/etcd/init.sls
inercia: In my expericne, 5 seconds is not enough time for knowing if the server is up and running (for example, it can timeout not reaching consensus).

And retrying these checks after 100ms does not provide anything (besides overloading the machine). It is better to just check less but for more time.

And adding 5 seconds per etcd instance is not so bad. They are all provisioned in parallel, so we are adding 5 seconds to the whole orchestration...

salt/orch/removal.sls
ereslibre: A replacement refers basically to `run this etcd member on this replacement when it was running in this node that was to be removed`? In the context of this PR, and in general in our setup, the replacement is strictly referred to `etcd` as far as I can see. No?

salt/etcd/init.sls
ereslibre: 100ms is not overloading the machine at all (in fact it's still asking for more work), but OK.

salt/orch/removal.sls
inercia: The replacement is the node that replaces the removed one. This replacement can be provided by the user or auto-calculated. It is not only for etcd: you could replace a k8s master (with no etcd running), so the replacement for this master would be marked with `addition_in_progress`... (so far we are only using this flag for etcd, but it coud be used in other roles too)

PR #354, opened 2018-01-22
Early mark nodes requiring update reboot as update in progress.

ereslibre:
This will allow us to reduce the timeframe in which the update-etc-hosts
orchestration can pop up, eventually running states on minions effectively
taking their lock and making this orchestration fail. We don't want the
update-etc-hosts orchestration to interfere with the main update
orchestration.

We'll release minion per minion grain when they are done, but let's
block all of them at the very beginning.

Fixes: bsc#1077086

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -1,3 +1,14 @@
+# Ensure all nodes with updates are marked as upgrading. This will reduce the time window in which
+# the update-etc-hosts orchestration can run in between machine restarts.
+set-update-grain:
+  salt.function:
+    - tgt: G@roles:kube-* and G@tx_update_reboot_needed:true
+    - tgt_type: compound
+    - name: grains.setval
+    - arg:
+      - update_in_progress
+      - true
+
 # Generate sa key (we should refactor this as part of the ca highstate along with its counterpart
 # in orch/kubernetes.sls)
 generate_sa_key:
@@ -58,17 +69,6 @@ pre-orchestration-migration:
 {%- set masters = salt.saltutil.runner('mine.get', tgt='G@roles:kube-master and G@tx_update_reboot_needed:true', fun='network.interfaces', tgt_type='compound') %}
 {%- for master_id in masters.keys() %}
 
-# Ensure the node is marked as upgrading
-{{ master_id }}-set-update-grain:
-  salt.function:
-    - tgt: {{ master_id }}
-    - name: grains.setval
-    - arg:
-      - update_in_progress
-      - true
-    - require:
-      - salt: update_modules
-
 {{ master_id }}-clean-shutdown:
   salt.state:
     - tgt: {{ master_id }}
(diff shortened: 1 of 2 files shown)


ereslibre:
The way I tested this was to deploy a cluster with `3.0` (`master`). After the bootstrap has finished, what I run on the hypervisor are two scripts:

```bash
#!/usr/bin/env bash

DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"

ssh -i $DIR/../misc-files/id_shared root@10.17.1.0 <<'EOF'
  docker exec -i $(docker ps | grep salt-master | awk '{print $1}') salt -P "roles:kube-(master|minion)" grains.setval tx_update_reboot_needed true
  docker exec -i $(docker ps | grep salt-master | awk '{print $1}') salt '*' saltutil.refresh_grains
  docker exec -i $(docker ps | grep salt-master | awk '{print $1}') salt-run state.orchestrate orch.update -l debug
EOF
```

Right when the `orch.update` starts, I run another script that will call to the `update-etc-hosts` orchestration in a loop, forcing conflict:

```bash
#!/usr/bin/env bash

DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"

ssh -i $DIR/../misc-files/id_shared root@10.17.1.0 <<'EOF'
  docker exec -i $(docker ps | grep salt-master | awk '{print $1}') bash -c 'while true; do salt-run state.orchestrate orch.update-etc-hosts; done'
EOF
```

What I can confirm is that without this patch on every run the `orch.update` orchestration failed because the `update-etc-hosts` got in between (something that can happen in real life because we are rebooting machines during the `orch.update` orchestration).

The idea of this PR is to "lock" all machines that are going to be part of the update process, and then:

* If the machine is a worker and it finished updating, release it.
* If the machine is a master, keep it locked until we run the post-update process, then release all masters. (We use one master [called `super_master`] to run some specific logic on it after everything is back from rebooting).

When I say "release" in this context I mean releasing from the `update-etc-hosts` perspective. A not released machine will continue having the `update_in_progress: True` grain while a released machine will not have this grain. This grain blocks `update-etc-hosts` orchestration from rendering that machine on that orchestration.

salt/orch/update.sls
```diff
@@ -282,3 +269,17 @@ services_setup:
       - dex
     - require:
       - cni_setup
+
+{% if masters|length > 0 %}
+masters-remove-update-grain:
+  salt.function:
+    - tgt: L@{{ masters.keys()|join(',') }}
```
inercia: Can't we just use?

```
- tgt: G@roles:kube-master and G@update_in_progress:true
- tgt_type: compound
```

ereslibre: We can, it was just to make sure to use the exact same subset that was used in the rest of the orchestration. Do you want me to change it or keep it as is?

PR #356, opened 2018-01-23
Early mark nodes requiring update reboot as update in progress.

ereslibre:
This will allow us to reduce the timeframe in which the update-etc-hosts
orchestration can pop up, eventually running states on minions effectively
taking their lock and making this orchestration fail. We don't want the
update-etc-hosts orchestration to interfere with the main update
orchestration.

We'll release minion per minion grain when they are done, but let's
block all of them at the very beginning.

Fixes: bsc#1077086

Backport of https://github.com/kubic-project/salt/pull/354

### DIFF
--- a/salt/_states/caasp_retriable.py
+++ b/salt/_states/caasp_retriable.py
@@ -30,7 +30,10 @@ def retry(name, target, retry={}, **kwargs):
     ret = None
 
     for attempt in xrange(retry_['attempts']):
-        ret = __states__[target](name=name, **kwargs)
+        try:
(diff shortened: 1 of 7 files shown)


ereslibre:
In this case, the backport is slightly different as we aren't performing special operations on any machine after all of them have been updated, so we just lock them at the very beginning of the orchestration and release all of them one by one.

No need to lock any of them until the end.

ereslibre:
[Reproduce SE's issues with updates](https://trello.com/c/bLLDZWYa/579-reproduce-ses-issues-with-updates)

PR #358, opened 2018-01-24
Add pillar root for public cloud specific config

jgleissner:
This adds a second pillar root for cloud specific configuration. See also [Add mount for public cloud pillar](https://github.com/kubic-project/caasp-container-manifests/pull/142) in caasp-container-manifests.


flavio:
I think the failure is unrelated, I would go ahead and merge it. Any objection?

PR #362, opened 2018-01-24
Catch any exceptions on `caasp_retriable` state, and treat it as a regular failure

ereslibre:
Explicitly wait for `caasp_retriable` states.

Fixes: bsc#1070989


ereslibre:
CI not passing on `master`, looks like Velum is broken.

PR #366, opened 2018-01-25
Try to be idempotent again...

inercia:


inercia:
Jenkins jobs are failing because etcd is not running in all nodes now...

salt/_macros/kubectl.jinja
```diff
@@ -63,7 +63,25 @@
...
                   watch=["file: " + manifest] + kwargs.pop('watch', []),
                   **kwargs) }}
-{%- endmacro %}
+    {#- make this idempotent by using a 'onchanges' on the file: #}
+    {#- we don't run 'kubectl apply' unless the file changes     #}
+    - onchanges:
+      - file: {{ manifest }}
+  {#- make sure we remove the manifest if something fails,     #}
+  {#- so we could run kubectl again if we orchestrate again... #}
+  cmd.run:      {#- cannot use file.absent to avoid duplicates #}
```
mjura: I am curious why we can't write:
```
{{ manifest }}:
  file.absent
```

salt/swap/init.sls
```diff
@@ -5,6 +5,7 @@ coreutils:
 unmount-swaps:
   cmd.run:
     - name: /sbin/swapoff -a
+    - onlyif: test `tail -n +2 /proc/swaps | wc -l` != 0
```
mjura: Small nitpick this should goes as other PR

salt/_macros/kubectl.jinja
inercia: Because there is a `file.managed` in the same block, and Salt does not allow two `file.*` statements in the same block... :-(

salt/swap/init.sls
inercia: This is for making `unmount-swaps` idempotent: it will not try to `swapoff` if there are no swap partitions.

salt/_macros/kubectl.jinja
mjura: Ok, thank you for explanation

salt/swap/init.sls
mjura: Ok, I've got your point

PR #368, opened 2018-01-31
Fix breakage introduced by docker update

kiall:
* Docker will no longer accept a `docker cp` over /etc/hosts
* Fix docker package name

### DIFF
--- a/salt/etc-hosts/init.sls
+++ b/salt/etc-hosts/init.sls
@@ -24,17 +24,35 @@ update-velum-hosts:
     - name: |-
         velum_id=$(docker ps | grep velum-dashboard | awk '{print $1}')
         if [ -n "$velum_id" ]; then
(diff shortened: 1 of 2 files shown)


kiall:
This is layering on another hack on top of the existing hacks. We should, when things aren't 100% broken, spend some time reevaluating this section of the code.

salt/etc-hosts/init.sls
```diff
@@ -24,17 +24,35 @@ update-velum-hosts:
     - name: |-
         velum_id=$(docker ps | grep velum-dashboard | awk '{print $1}')
         if [ -n "$velum_id" ]; then
-            docker cp /etc/hosts $velum_id:/etc/hosts
+            docker cp /etc/hosts $velum_id:/etc/hosts-caasp
```
mjura: Can you please elaborate why it has to be copied to /etc/hosts-caasp first?

PR #375, opened 2018-02-02
Update flannel image tag to match flannel version

kiall:
Must be accepted at the same time as https://build.suse.de/request/show/153722


kiall:
Merged, as https://build.suse.de/request/show/153722 was accepted.

PR #377, opened 2018-02-02
Namespace the roles and cluster roles we create

kiall:
When we create a role, rolebinding etc, we should namespace the names
in order to make it obvious these are deployed as part of CaaSP, as well
as to help ensure these are obviously part of CaaSP, not a stock part of
Kubernetes.

I've gone with a "suse:caasp:" prefix, which matches the "system:" prefix
for built in roles/rolebindings/etc.

Additionally, move a few "kubectl create clusterrolebinding" shell outs over
to the matching YAML files, keeping everything together.

### DIFF
--- a/salt/addons/dns/init.sls
+++ b/salt/addons/dns/init.sls
@@ -10,11 +10,10 @@ include:
                           "/etc/kubernetes/addons/kubedns.yaml",
                           check_cmd="kubectl get deploy kube-dns -n kube-system | grep kube-dns") }}
 
-{{ kubectl("create-dns-clusterrolebinding",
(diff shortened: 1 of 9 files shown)


kiall:
the "onlyif" condition on the removals depends on https://github.com/kubic-project/salt/pull/317 merging to function.

kiall:
Once this goes in - I'll look at removing everyone's cluster-admin access. That seems like just a little bit of a blunt instrument to get things working -_-

kiall:
I've included the "Remove old mis-named tiller deployment" commit here to let this work - the other PR should still be reviewed and merged first though

PR #380, opened 2018-02-06
Use separate Dex clients for each actual client

kiall:
Previously Velum, CaaSP CLI, and Kubernetes all shared a single Dex
client. From a security perspective, this was far from ideal.

Update Dex with 3 clients, one for each actual client. Both the Velum
and CaaSP CLI clients are allowed to issue tokens for the Kubernetes
client.

Depends-On: https://github.com/kubic-project/velum/pull/439
Depends-On: https://github.com/kubic-project/caasp-cli/pull/9

### DIFF
--- a/salt/dex/dex.yaml
+++ b/salt/dex/dex.yaml
@@ -69,13 +69,30 @@ data:
       skipApprovalScreen: true
 
     staticClients:
+    - id: kubernetes
+      redirectURIs:
+      - 'urn:ietf:wg:oauth:2.0:oob'
+      name: "Kubernetes"
+      secret: "{{ pillar['dex']['client_secrets']['kubernetes'] }}"
+      trustedPeers:
+      - caasp-cli
+      - velum
+
     - id: caasp-cli
       redirectURIs:
+      - 'urn:ietf:wg:oauth:2.0:oob'
       - 'http://127.0.0.1'
-      - 'https://{{ pillar['dashboard'] }}/oidc/done'
-      - 'https://{{ pillar['dashboard_external_fqdn'] }}/oidc/done'
+      - 'http://localhost'
       name: "CaaSP CLI"
       secret: "swac7qakes7AvucH8bRucucH"
+      public: true
+
+    - id: velum
+      redirectURIs:
+      - 'https://{{ pillar['dashboard'] }}/oidc/done'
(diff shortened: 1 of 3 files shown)


kiall:
This will fail CI without the dependant commits

salt/dex/dex.yaml
```diff
@@ -69,13 +69,30 @@ data:
...
       name: "CaaSP CLI"
       secret: "swac7qakes7AvucH8bRucucH"
+      public: true
+
+    - id: velum
+      redirectURIs:
+      - 'https://{{ pillar['dashboard'] }}/oidc/done'
+      - 'https://{{ pillar['dashboard_external_fqdn'] }}/oidc/done'
+      name: "Velum"
+      secret: "{{ pillar['dex']['client_secrets']['velum'] }}"
```
kiall: Docs on most these values is here https://github.com/coreos/dex/blob/master/Documentation/custom-scopes-claims-clients.md

salt/dex/dex.yaml
```diff
@@ -69,13 +69,30 @@ data:
       skipApprovalScreen: true
 
     staticClients:
+    - id: kubernetes
+      redirectURIs:
+      - 'urn:ietf:wg:oauth:2.0:oob'
+      name: "Velum"
```
ereslibre: I guess we want a different name here, only for descriptive purposes?

kiall: Whoops - Nice catch

kiall: Fixed

salt/dex/dex.yaml
```diff
@@ -69,13 +69,30 @@ data:
...
+
     - id: caasp-cli
       redirectURIs:
+      - 'urn:ietf:wg:oauth:2.0:oob'
       - 'http://127.0.0.1'
-      - 'https://{{ pillar['dashboard'] }}/oidc/done'
-      - 'https://{{ pillar['dashboard_external_fqdn'] }}/oidc/done'
+      - 'http://localhost'
       name: "CaaSP CLI"
       secret: "swac7qakes7AvucH8bRucucH"
```
mjura: Is this hardcoded secret value?

mjura: Can we move it also to pillar and don't keep it in template? It can be also randomly generated

PR #381, opened 2018-02-06
Mark the haproxy as critical pod

flavio:
Flag the haproxy pods providing connectivity to the API server as critical ones.

This should force kubelet and the scheduler to never ever get rid of them. If these pods are killed to make more space for other ones, the node would not be able to talk with the API server making it useless.

More details inside [upstream doc](https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/).

### DIFF
--- a/salt/haproxy/haproxy.yaml.jinja
+++ b/salt/haproxy/haproxy.yaml.jinja
@@ -6,6 +6,8 @@ metadata:
   namespace: kube-system
   labels:
     name: haproxy
+  annotations:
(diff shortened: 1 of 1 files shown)


flavio:
I tested the change locally and it works, meaning a `kubectl describe pod` returns all the expected fields.

inercia:
@flavio Shouldn't we backport this to 2.0?

kiall:
@inercia - yea, we probably should!

PR #382, opened 2018-02-07
Resist 'None' certs in pillars

inercia:
Make sure we do not crash on pillars that are not properly formatted by surrounding the code with some `try..except` blocks.

### DIFF
--- a/salt/_modules/caasp_docker.py
+++ b/salt/_modules/caasp_docker.py
@@ -40,44 +40,53 @@ def get_registries_certs(lst, default_port=5000):
 
     LOG.debug('Finding certificates in: %s', lst)
     for registry in lst:
-        url = registry.get('url')
-        cert = registry.get('cert', '')
-        if len(cert) > 0:
+        try:
+            url = registry.get('url')
 
-            # parse the name as an URL or "host:port", and return <HOST>[:<PORT>]
-            hostname, port = _get_hostname_and_port(url)
-            host_port = hostname
-            if port:
-                host_port += ":" + str(port)
+            cert = registry.get('cert', '')
+            if cert:
 
-            LOG.debug('Adding certificate for: %s', host_port)
-            certs[host_port] = cert
+                # parse the name as an URL or "host:port", and return <HOST>[:<PORT>]
+                hostname, port = _get_hostname_and_port(url)
+                host_port = hostname
+                if port:
+                    host_port += ":" + str(port)
 
-            if port:
-                if port == default_port:
-                    # When using the standar port (5000), if the user introduces
-                    # "my-registry:5000" as a trusted registry, he/she will be able
-                    # to do "docker pull my-registry:5000/some/image" but not
-                    # "docker pull my-registry/some/image".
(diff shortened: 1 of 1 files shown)


salt/_modules/caasp_docker.py
```diff
@@ -40,44 +40,53 @@ def get_registries_certs(lst, default_port=5000):
...
-                    # to do "docker pull my-registry:5000/some/image" but not
-                    # "docker pull my-registry/some/image".
-                    # So we must also create the "ca.crt" for "my-registry"
-                    # as he/she could just access "docker pull my-registry/some/image",
-                    # and Docker would fail to find "my-registry/ca.crt"
-                    name = hostname
+        try:
+            url = registry.get('url')
+
+            cert = registry.get('cert', '')
```
ereslibre: So, if the `cert` is `None` it will prevail as this `''` fallback is just in case the `cert` entry does not exist, so the `''` case never gets used. Another way to get around the issue would be to do `registry.get('cert', '') or ''`. That way the cert would never be a `None` (note that the `''` on the `registry.get` could very well be optional, but in any case, it's good to have).

Wouldn't that fix the issue alone?

salt/_modules/caasp_docker.py
```diff
@@ -40,44 +40,53 @@ def get_registries_certs(lst, default_port=5000):
...
-                certs[name] = cert
-
-        mirrors = registry.get('mirrors', [])
-        if len(mirrors) > 0:
-            LOG.debug('Looking recursively for certificates in mirrors')
-            certs_mirrors = get_registries_certs(mirrors,
-                                                 default_port=default_port)
-            certs.update(certs_mirrors)
+
+        except Exception, e:
```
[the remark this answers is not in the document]
inercia: > Ideally you shouldn't catch Exception, at least not on its own.

I agree that ideallly we should not do that, but in practice we don't want to fail an orchestration just beacuse a certificate is wrong. Fine-grain exception filtering would make sense if we would handle specific errors in specific ways, but this is a all-or-nothing thing

salt/_modules/caasp_docker.py
inercia: The `if cert` in the next line will fix the issue...

flavio: Isn't it better to fail an orchestration rather than make it work and have some stuff not configured properly (and hence causing harder to debug problems?)

inercia: At the end, the difference will be 1) orchestration succedes but no certificate is added and an error is logged or 2) orchestration just fails. Developers usually prefer 1), while users prefer 2). Both situations will lead to users filling a new issue in Bugzilla, and they are both easy to debug.

In this case I prefer 1) too (as I think it is an overkill to  abort an orchestration by a failure on the certificates), but I guess this is more a philoshophical question...

flavio: Fair enough, I don't have a strong opinion on that. I'm fine leaving things as they currently are implemented (catch exception)

PR #384, opened 2018-02-08
When executing a highstate of `apiserver` make sure that we check the local `apiserver` instance

ereslibre:
When executing the highstate make sure the `apiserver` we are checking
is the local one, not *any* master through haproxy.

Fixes: bsc#1079460

### DIFF
--- a/salt/haproxy/haproxy.cfg.jinja
+++ b/salt/haproxy/haproxy.cfg.jinja
@@ -22,11 +22,13 @@ defaults
 listen kubernetes-master
         bind {{ bind_ip }}:{{ pillar['api']['ssl_port'] }}
         mode tcp
-        default-server inter 10s fall 3
+        default-server inter 10s fall 2
         balance roundrobin
+        option redispatch
+        option httpchk GET /healthz
 
 {%- for minion_id, nodename in salt['mine.get']('roles:kube-master', 'nodename', 'grain').items() %}
-        server master-{{ minion_id }} {{ nodename }}:{{ pillar['api']['int_ssl_port'] }} check
+        server master-{{ minion_id }} {{ nodename }}:{{ pillar['api']['int_ssl_port'] }} check check-ssl port {{ pillar['api']['int_ssl_port'] }} verify none
 {% endfor -%}
 
 {%- if "admin" in salt['grains.get']('roles', []) %}
@@ -34,8 +36,9 @@ listen kubernetes-master
 listen kubernetes-dex
         bind {{ bind_ip }}:32000
         mode tcp
-        default-server inter 10s fall 3
+        default-server inter 10s fall 2
         balance roundrobin
+        option redispatch
 
 {%- for minion_id, nodename in salt['mine.get']('roles:kube-master', 'nodename', 'grain').items() %}
         server master-{{ minion_id }} {{ nodename }}:32000 check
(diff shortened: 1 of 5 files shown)


ereslibre:
WIP as I still need to check if it fixes the upgrade issues to 1.8 with multi master. If it does, I will add a `Fixes: bsc#` to the commit, so I'll drop any reviews at that time.

salt/haproxy/haproxy.cfg.jinja
```diff
@@ -24,9 +24,10 @@ listen kubernetes-master
         mode tcp
         default-server inter 10s fall 3
         balance roundrobin
+        option httpchk GET /healthz
 
 {%- for minion_id, nodename in salt['mine.get']('roles:kube-master', 'nodename', 'grain').items() %}
-        server master-{{ minion_id }} {{ nodename }}:{{ pillar['api']['int_ssl_port'] }} check
+        server master-{{ minion_id }}:{{ pillar['api']['int_ssl_port'] }} {{ nodename }}:{{ pillar['api']['int_ssl_port'] }} check-ssl verify none
```
kiall: Should this not be `check ssl verify none` instead of `check-ssl verify none` ?

And - any reason we're adding the port the backend name? That name is just an arbitrary label haproxy asks for - it could be anything, so long as it's unique - for consistency - we should either leave the port off, or add it to all the names for all the services fronted by haproxy.

flavio:
So far LGTM

salt/haproxy/haproxy.cfg.jinja
ereslibre: According to HAProxy's documentation I see `check-ssl`: https://www.haproxy.com/documentation/aloha/7-0/traffic-management/lb-layer7/health-checks/.

I will drop the port on the name, yeah.

ereslibre:
I have done **a lot** of testing. A little sum up, but I've tried lots of different combinations:

- Healthcheck is not optimal. It will let a window in which things can fail (e.g. 10s each check, 2 required failed checks to determine server is down -- 20s in which we could try to go to a server that is not available).

A small comment that I found out, `SIGHUP` will **not** really reread the haproxy config, it will just print some stats on the terminal. I was under the assumption that it did, and realized that the configs I were modifying were not being applied by the `SIGHUP`. So, we'll need to apply the haproxy config changes in some other way...

I found redispatch to be extremely awesome. While with `healthcheck` we could miss requests in important moments (on an update process we do some `kubectl` commands that will round robin the server we reach on one specific machine -- could happen that on a certain point in time a request will not succeed and the healthcheck did not even realize). With `redispatch` I was completely unable to make the thing fail as long as there's one master responding at all.

ereslibre:
The health check will help to ensure that we only consider as UP a server that has been completely initialized. On the update process: 

- Stop all kube-master related services
- Sleep 15 seconds and reboot
- HAproxy on the rest of the nodes should realize the machine is down (after 20 seconds max, at this point this leaves us with 5 seconds in which the machine would have to have rebooted and started the api server in an incomplete fashion for a `kubectl` command to fail -- by incomplete I mean accepting connections but not having loaded all apiserver plugins).

The good part is that `redispatch` covers us when one or more masters are not listening at all on the socket, while the health check (with its time windows) will allow us to only attack healthy apiservers.

salt/haproxy/haproxy.cfg.jinja
mjura: Yes, `check-ssl` is valid for our version of Haproxy v1.6
https://cbonte.github.io/haproxy-dconv/configuration-1.6.html#5.2-check-ssl

PR #386, opened 2018-02-08
Dex pods: introduce anti-affinity rule

flavio:
Our dex deployment creates 3 pods running the dex service. There are really high chances (or even certainty in the case of clusters made by 1or 2 worker nodes) that all the dex pods end up running on the same node.

This is bad from a HA perspective, plus we end up taking away resources from small clusters.

With the following change we enforce the kubernetes scheduler to always spread the dex pods over different nodes.

On small clusters (1 or 2 nodes) the deployment will be running with a lower number of replicas until new nodes are added. This doesn't cause our orchestration to fail.

Adding new nodes at a later stage will allow the deployment to reach the desired replica size without any intervention from us or the user.

### DIFF
--- a/salt/dex/dex.yaml
+++ b/salt/dex/dex.yaml
@@ -118,6 +118,19 @@ spec:
       tolerations:
       - key: "CriticalAddonsOnly"
         operator: "Exists"
+
+      # ensure dex pods are running on different hosts
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - dex
+            topologyKey: "kubernetes.io/hostname"
+
       containers:
       - image: sles12/caasp-dex:2.7.1
         name: dex


salt/dex/dex.yaml
```diff
@@ -118,6 +118,19 @@ spec:
       tolerations:
       - key: "CriticalAddonsOnly"
```
kiall: Should we also update the tolerations to all dex to run on the masters? which will give us some extra copies in smaller clusters? With a smaller test / POC cluster, we can easily have just a single worker node - giving us just 1 dex pod. If we allow it masters, we get the 3 we request.

flavio: That's a good question.

If we want to do that we have to start label the master nodes accordingly. Upstream usually sets `node-role.kubernetes.io/master=true`.

By doing that we could write a new toleration rule like:

```yaml
tolerations:
  - key: node-role.kubernetes.io/master
    value: "true"
    effect: PreferNoSchedule
```

kiall: I put this together which sets the taints and labels etc: https://github.com/kubic-project/salt/pull/388

kiall: Lets move this discussion over to https://github.com/kubic-project/salt/pull/388 - I've added it there, as well as correctly tainting the masters.

PR #388, opened 2018-02-08
Taint and Label Masters

kiall:
Masters should be tainted and labelled as masters, rather than setting
these nodes as unschedulable.

### DIFF
--- a/salt/kube-apiserver/init.sls
+++ b/salt/kube-apiserver/init.sls
@@ -2,6 +2,7 @@ include:
   - repositories
   - ca-cert
   - cert
+  - haproxy
   - kubernetes-common
   - kubernetes-common.serviceaccount-key
 
--- a/salt/kubelet/init.sls
+++ b/salt/kubelet/init.sls
@@ -11,10 +11,12 @@ include:
     - source: salt://kubelet/kubelet-initial.jinja
     - template: jinja
     - defaults:
-{% if not "kube-master" in salt['grains.get']('roles', []) %}
-      schedulable: "true"
+{% if "kube-master" in salt['grains.get']('roles', []) %}
+      nodeLabels: "node-role.kubernetes.io/master="
+      nodeTaints: "node-role.kubernetes.io/master=:NoSchedule"
 {% else %}
-      schedulable: "false"
+      nodeLabels: ""
+      nodeTaints: ""
 {% endif %}
 
 {% from '_macros/certs.jinja' import certs with context %}
@@ -97,7 +99,7 @@ kubelet:
     - require:
       - service: kubelet
 
-{% if not "kube-master" in salt['grains.get']('roles', []) and salt['grains.get']('kubelet:should_uncordon', false) %}
+{% if salt['grains.get']('kubelet:should_uncordon', false) %}
   caasp_cmd.run:
     - name: |
         kubectl uncordon {{ grains['nodename'] }}
--- a/salt/top.sls
+++ b/salt/top.sls
@@ -10,10 +10,19 @@ base:
     - rebootmgr
     - transactional-update
     - haproxy
+    - kubectl-config
   'roles:admin':
     - match: grain
     - velum
     - ldap
+  'roles:etcd':
+    - match: grain
+    - etcd
+  'roles:kube-master':
+    - match: grain
+    - kube-apiserver
+    - kube-controller-manager
+    - kube-scheduler
   'roles:kube-(master|minion)':
     - match: grain_pcre
     - ca-cert
@@ -23,14 +32,5 @@ base:
     - cert
     - docker
     - container-feeder
-    - kubectl-config
     - kubelet
     - kube-proxy
-  'roles:etcd':
-    - match: grain
-    - etcd
-  'roles:kube-master':
-    - match: grain
-    - kube-apiserver
-    - kube-controller-manager
-    - kube-scheduler
--- a/salt/_macros/kubectl.jinja
+++ b/salt/_macros/kubectl.jinja
@@ -1,10 +1,8 @@
 # NOTE: There are some implicit "requires" in these macros, so you
 # should start your SLS with:
 #
-# - include:
-#   - kube-apiserver
+# include:
 #   - kubectl-config
-#
 
 #####################################################################
 
@@ -18,7 +16,6 @@
         attempts: 10
         interval: 1
     - require:
-      - kube-apiserver
       - {{ pillar['paths']['kubeconfig'] }}
 {%- if 'require' in kwargs %}
   {%- for r in kwargs['require'] %}
--- a/salt/addons/dns/kubedns.yaml.jinja
+++ b/salt/addons/dns/kubedns.yaml.jinja
@@ -47,6 +47,9 @@ spec:
         scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
       tolerations:
+      - key: node-role.kubernetes.io/master
+        operator: Exists
+        effect: NoSchedule
       - key: "CriticalAddonsOnly"
         operator: "Exists"
       volumes:
--- a/salt/addons/tiller/tiller.yaml.jinja
+++ b/salt/addons/tiller/tiller.yaml.jinja
@@ -48,6 +48,9 @@ spec:
         scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
       tolerations:
+      - key: node-role.kubernetes.io/master
+        operator: Exists
+        effect: NoSchedule
       - key: "CriticalAddonsOnly"
         operator: "Exists"
       containers:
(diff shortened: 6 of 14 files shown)


salt/kubelet/init.sls
```diff
@@ -12,10 +12,12 @@ include:
     - source: salt://kubelet/kubelet-initial.jinja
     - template: jinja
     - defaults:
-{% if not "kube-master" in salt['grains.get']('roles', []) %}
-      schedulable: "true"
+{% if "kube-master" in salt['grains.get']('roles', []) %}
+      nodeLabels: "node-role.kubernetes.io/master="
```
flavio: Shouldn't we set that to be:
```
 nodeLabels: "node-role.kubernetes.io/master=true"
```

To simplify label matching later on (see the dex PR)

kiall: No, for some reason I can't fathom, master=<empty string> was chosen upstream.

See https://github.com/kubernetes/kubeadm/blob/master/docs/design/design_v1.9.md#mark-master

salt/kubelet/init.sls
```diff
@@ -116,6 +120,24 @@ remove-should-uncordon-grain:
     - name: kubelet:should_uncordon
     - destructive: True
 
+# TODO: onlyif's need doing
+
+{% if "kube-master" in salt['grains.get']('roles', []) %}
```
kiall: Ah. So - This will fail given kube-api isnt running - it only works in dev as I had already bootstrapped. Humm.

kiall:
Node Listing:

	caasp-master-0:~ # kubectl get node -o wide
	NAME             STATUS    ROLES     AGE       VERSION   EXTERNAL-IP   OS-IMAGE                                   KERNEL-VERSION        CONTAINER-RUNTIME
	caasp-master-0   Ready     master    1h        v1.8.7    <none>        SUSE Container as a Service Platform 3.0   4.4.92-6.18-default   docker://17.9.1
	caasp-worker-0   Ready     <none>    1h        v1.8.7    <none>        SUSE Container as a Service Platform 3.0   4.4.92-6.18-default   docker://17.9.1
	caasp-worker-1   Ready     <none>    1h        v1.8.7    <none>        SUSE Container as a Service Platform 3.0   4.4.92-6.18-default   docker://17.9.1

Master Node Dump:

	caasp-master-0:~ # EDITOR=cat kubectl get node -o yaml
	See https://gist.github.com/kiall/0f0aa24abbde6ba23115b3fb7468c83f

All pods, and where they are running:

	caasp-master-0:~ # kubectl get pods --all-namespaces -o wide
	NAMESPACE     NAME                             READY     STATUS    RESTARTS   AGE       IP           NODE
	kube-system   dex-7696cbdf7c-4qxzf             1/1       Running   0          6m        172.16.4.3   caasp-worker-0
	kube-system   dex-7696cbdf7c-hdwnh             1/1       Running   6          6m        172.16.2.3   caasp-worker-1
	kube-system   dex-7696cbdf7c-kx27g             1/1       Running   0          6m        172.16.0.2   caasp-master-0
	kube-system   haproxy-caasp-master-0           1/1       Running   0          1h        10.17.2.0    caasp-master-0
	kube-system   haproxy-caasp-worker-0           1/1       Running   0          1h        10.17.3.0    caasp-worker-0
	kube-system   haproxy-caasp-worker-1           1/1       Running   0          1h        10.17.3.1    caasp-worker-1
	kube-system   kube-dns-6d446f9754-bkn7w        3/3       Running   0          6m        172.16.2.2   caasp-worker-1
	kube-system   kube-flannel-9cmkj               1/1       Running   0          7m        10.17.3.0    caasp-worker-0
	kube-system   kube-flannel-pktsx               1/1       Running   0          7m        10.17.3.1    caasp-worker-1
	kube-system   kube-flannel-xsz6f               1/1       Running   0          7m        10.17.2.0    caasp-master-0
	kube-system   tiller-deploy-78c59cfc85-prnqv   1/1       Running   0          6m        172.16.4.2   caasp-worker-0

All pods running on the master (notice Dex):

	caasp-master-0:~ # kubectl get pods --all-namespaces -o wide | grep master
	kube-system   dex-7696cbdf7c-kx27g             1/1       Running   0          6m        172.16.0.2   caasp-master-0
	kube-system   haproxy-caasp-master-0           1/1       Running   0          1h        10.17.2.0    caasp-master-0
	kube-system   kube-flannel-xsz6f               1/1       Running   0          7m        10.17.2.0    caasp-master-0

Scale Kube-DNS up:

	caasp-master-0:~ # kubectl -n kube-system scale --replicas=3 deployment/kube-dns
	deployment "kube-dns" scaled

Check one of the new kube-dns's landed on the master:

	caasp-master-0:~ # kubectl get pods --all-namespaces -o wide | grep dns
	kube-system   kube-dns-6d446f9754-4bdjx        3/3       Running   0          16s       172.16.0.3   caasp-master-0
	kube-system   kube-dns-6d446f9754-bkn7w        3/3       Running   0          10m       172.16.2.2   caasp-worker-1
	kube-system   kube-dns-6d446f9754-zdnnv        3/3       Running   0          16s       172.16.4.4   caasp-worker-0

Scale Tiller up:

	caasp-master-0:~ # kubectl -n kube-system scale --replicas=3 deployment/tiller-deploy
	deployment "tiller-deploy" scaled

Check one of the new tillers's landed on the master:

	caasp-master-0:~ # kubectl get pods --all-namespaces -o wide | grep tiller
	kube-system   tiller-deploy-78c59cfc85-cb8br   1/1       Running   0          20s       172.16.2.4   caasp-worker-1
	kube-system   tiller-deploy-78c59cfc85-mg5tp   1/1       Running   0          20s       172.16.0.4   caasp-master-0
	kube-system   tiller-deploy-78c59cfc85-prnqv   1/1       Running   0          12m       172.16.4.2   caasp-worker-0

salt/top.sls
```diff
@@ -10,10 +10,19 @@ base:
     - rebootmgr
     - transactional-update
     - haproxy
+    - kubectl-config
```
ereslibre: Can this go in a group `roles:kube-(master|minion)`? Does the admin need this config?

salt/top.sls
```diff
@@ -10,10 +10,19 @@ base:
...
   'roles:admin':
     - match: grain
     - velum
     - ldap
+  'roles:etcd':
+    - match: grain
+    - etcd
+  'roles:kube-master':
```
ereslibre: Why the change of order of this specific bit?

salt/kube-apiserver/init.sls
```diff
@@ -2,6 +2,7 @@ include:
   - repositories
   - ca-cert
   - cert
+  - haproxy
```
ereslibre: If we rely on the `top.sls` order as before, maybe this can be removed as well as the include in `haproxy/init.sls`, no?

salt/top.sls
kiall: Actually, while it's unrelated, it's really nice to have the config there - kubectl works on the admin node all of a sudden ;)

salt/kube-apiserver/init.sls
kiall: `top.sls` ordering still comes into effect, but `top.sls` ordering is just a hint to salt - it doesn't enforce anything. The only way to be sure state A has ran on a machine before state B is to encode that into an orchestration (which spreads the dependency chain out over the whole repo), or encode it into the states directly allowing any state to be called, and the right dependency ordering is generated by salt.

In this case, the `kube-apiserver-up` at the bottom of this file will fail if `haproxy` hasn't been setup - so we declare the dependency.

salt/top.sls
kiall: The initial starting point for salt as it was ordering the states was wrong, I believe the requires alone are probably enough to resolve this, but, still - it's better to put the "starting order" salt uses to seed the dependency graph into the rough order we expect things to be deployed.

salt/kube-apiserver/init.sls
ereslibre: However, the `include` alone also doesn't enforce any order (https://docs.saltstack.com/en/latest/ref/states/include.html, see the NOTE), we should also add a `require`.

salt/top.sls
ereslibre: It's tricky because visually a kube-master on the `top.sls` looks like this:

```
base:
  'roles:(admin|kube-(master|minion))':
    - match: grain_pcre
    - swap
    - etc-hosts
    - proxy
    - rebootmgr
    - transactional-update
    - haproxy
    - kubectl-config
  'roles:etcd': (optional, depends if it has the `etcd` role too)
    - match: grain
    - etcd
  'roles:kube-master':
    - match: grain
    - kube-apiserver
    - kube-controller-manager
    - kube-scheduler
  'roles:kube-(master|minion)':
    - match: grain_pcre
    - ca-cert
    - repositories
    - motd
    - users
    - cert
    - docker
    - container-feeder
    - kubelet
    - kube-proxy
```

So visually we are also applying `kube-apiserver` before `ca-cert`, for example... but anyways, as you say this is not an enforcement of order. In fact, ordering semantics in salt are... "special".

salt/kube-apiserver/init.sls
kiall: Ugh - salt's ordering, combined with how our states/highstates/orchestrations are laid out makes this *really* hard. heh

PR #391, opened 2018-02-12
[2.0] Make sure we do not crash on pillars that are not properly formatted.

inercia:
Backport of #382


mjura:
As tip for future, just try to cherry-pick patches with -x, it will source original commit id
```
git cherry-pick -x <commit_id>
```

PR #394, opened 2018-02-16
Velum Dash and API both attempt to bind to the same port

kiall:
It's not possible to reliably bind to 0.0.0.0:443 for one service,
and 127.0.0.1:443 for another service.

As such, we'll move velum-api over to 127.0.0.1:444

Depends-On: https://github.com/kubic-project/caasp-container-manifests/pull/155

### DIFF
--- a/config/master.d/50-master.conf
+++ b/config/master.d/50-master.conf
@@ -37,5 +37,5 @@ module_dirs:
 
 ext_pillar:
   - velum:
-      url: https://localhost/internal-api/v1/pillar.json
+      url: https://localhost:444/internal-api/v1/pillar.json
       ca_bundle: /etc/pki/ca.crt
--- a/salt/haproxy/haproxy.cfg.jinja
+++ b/salt/haproxy/haproxy.cfg.jinja
@@ -57,7 +57,7 @@ listen velum
         server velum unix@/var/run/puma/dashboard.sock
 
 listen velum-api
-        bind 127.0.0.1:443 ssl crt /etc/pki/velum.pem ca-file /etc/pki/ca.crt
+        bind 127.0.0.1:444 ssl crt /etc/pki/velum.pem ca-file /etc/pki/ca.crt
         mode http
         option forwardfor
         http-request set-header X-Forwarded-Proto https


ereslibre:
Interesting, they should be bound on different ifaces (localhost, or 127.0.0.1 for the internal API), but if it's being problematic LGTM.

kiall:
> Interesting, they should be bound on different ifaces (localhost, or 127.0.0.1 for the internal API), but if it's being problematic LGTM.

0.0.0.0 includes 127.0.0.1 as it's means "bind to everything"

kiall:
> LGTM, I'm not sure only about bad FQDN name as localhost and SSL cert verification

TLS ignores the port numbers (I kinda feel like thats a gap in TLS.. but.. thats how it is)

ereslibre:
Yes, thought 0.0.0.0 would get the packet if there's no interface more specific. Specific should win before 0.0.0.0, but anyway LGTM, it's safer.

ereslibre:
My comment is probably not accurate, I don't know how HAProxy binds this internally. I'd be curious just to check, at least I saw that the internal api server requests ended up in the api instance and the external ones to the 0.0.0.0 bound to all interfaces. Maybe I was just wrong and it was working for some reason I don't know at the moment. Without magic socket options it's true that it fails to bind anything and 0.0.0.0 to the same port :/

kiall:
> My comment is probably not accurate, I don't know how HAProxy binds this internally. I'd be curious just to check, at least I saw that the internal api server requests ended up in the api instance and the external ones to the 0.0.0.0 bound to all interfaces. Maybe I was just wrong and it was working for some reason I don't know at the moment. Without magic socket options it's true that it fails to bind anything and 0.0.0.0 to the same port :/

The only way I can see this "working" is if HAProxy is smart enough to bind the most specific first, which will actually work from past experience - but it's still confusing and *potentially* haproxy was just eating an error and we never noticed :)

ereslibre:
Thanks Kiall.

kiall:
Now - either get back to your holiday, or come back to work - pick one and stick to it! ;)

PR #396, opened 2018-02-21
add ldap.infra.caasp.local to the certificate

MaximilianMeister:
feature#net-ldap-cert

## related to
https://github.com/kubic-project/caasp-container-manifests/pull/153

### DIFF
--- a/salt/etc-hosts/hosts.jinja
+++ b/salt/etc-hosts/hosts.jinja
@@ -5,6 +5,8 @@
 # and forwarding connections to the real master
 {% if "kube-master" in salt['grains.get']('roles', []) %}
 127.0.0.1 api api.{{ pillar['internal_infra_domain'] }}{% if not salt['caasp_filters.is_ip'](external_fqdn) %} {{ external_fqdn }}{% endif %}
+{% elif "admin" in salt['grains.get']('roles', []) %}
+127.0.0.1 api api.{{ pillar['internal_infra_domain'] }} ldap.{{ pillar['internal_infra_domain'] }}
 {% else %}
 127.0.0.1 api api.{{ pillar['internal_infra_domain'] }}
 {% endif %}
--- a/salt/ldap/init.sls
+++ b/salt/ldap/init.sls
@@ -2,7 +2,7 @@ include:
   - ca-cert
   - cert
 
-{% set names = [salt.caasp_pillar.get('dashboard')] %}
+{% set names = [salt.caasp_pillar.get('dashboard'), 'ldap.' + pillar['internal_infra_domain']] %}
 
 {% from '_macros/certs.jinja' import alt_names, certs with context %}
 {{ certs("ldap:" + grains['nodename'],


ereslibre:
LGTM, but shouldn't we also add it to the etc-hosts orchestration so we maintain this entry with salt too and never lose it?

MaximilianMeister:
> LGTM, but shouldn't we also add it to the etc-hosts orchestration so we maintain this entry with salt too and never lose it?

added

salt/etc-hosts/hosts.jinja
```diff
@@ -12,7 +12,7 @@
 ### admin nodes ###
 {%- set admins = salt['mine.get']('roles:admin', 'network.interfaces', 'grain') %}
 {%- for admin_id, ifaces in admins.items() %}
-{{ salt.caasp_net.get_primary_ip(host=admin_id, ifaces=ifaces) }} {{ salt.caasp_net.get_nodename(host=admin_id) }} {{ salt.caasp_net.get_nodename(host=admin_id) }}.{{ pillar['internal_infra_domain'] }} {{ admin_id }} {{ admin_id }}.{{ pillar['internal_infra_domain'] }}
+{{ salt.caasp_net.get_primary_ip(host=admin_id, ifaces=ifaces) }} {{ salt.caasp_net.get_nodename(host=admin_id) }} {{ salt.caasp_net.get_nodename(host=admin_id) }}.{{ pillar['internal_infra_domain'] }} {{ admin_id }} {{ admin_id }}.{{ pillar['internal_infra_domain'] }} ldap.{{ pillar['internal_infra_domain'] }}
```
ereslibre: Adding the same entry several times won't have the desired effect. Instead of this, I think we want a separate line outside of this loop with something like: `127.0.0.1 ldap.{{ pillar['internal_infra_domain'] }}`.

ereslibre: And only in the admin role, since the rest of the nodes won't have ldap running locally.

MaximilianMeister: updated

salt/etc-hosts/hosts.jinja
```diff
@@ -14,6 +14,9 @@
 {%- for admin_id, ifaces in admins.items() %}
 {{ salt.caasp_net.get_primary_ip(host=admin_id, ifaces=ifaces) }} {{ salt.caasp_net.get_nodename(host=admin_id) }} {{ salt.caasp_net.get_nodename(host=admin_id) }}.{{ pillar['internal_infra_domain'] }} {{ admin_id }} {{ admin_id }}.{{ pillar['internal_infra_domain'] }}
 {%- endfor %}
+{% if "admin" in salt['grains.get']('roles', []) %}
+127.0.0.1 ldap.{{ pillar['internal_infra_domain'] }}
```
kiall: This duplicates line 9 above - and should be folded into that if/else branch.

MaximilianMeister: updated

PR #398, opened 2018-02-23
Restart instead of reload container-feeder

kiall:
container-feeder is a oneshot service, where reload makes no sense
and in unsupported. If this triggers, we ended up getting:

	salt-minion[2454]: [ERROR   ] Command '['systemd-run', '--scope', 'systemctl', 'reload', 'container-feeder.service']' failed with return code: 3
	salt-minion[2454]: Failed to reload container-feeder.service: Job type reload is not applicable for unit container-feeder.service.

### DIFF
--- a/salt/container-feeder/init.sls
+++ b/salt/container-feeder/init.sls
@@ -9,7 +9,6 @@ include:
 container-feeder:
   service.running:
     - enable: True
-    - reload: True
     - require:
       - cmd: docker
     - watch:


salt/container-feeder/init.sls
```diff
@@ -9,7 +9,7 @@ include:
 container-feeder:
   service.running:
```
ereslibre: Since it's a one-shot service maybe we can completely remove `reload` and `restart` from here. If that makes `container-feeder` run it'll be fine. Also, I see related past issues (https://github.com/saltstack/salt/issues/37842) with one shot services, so worst case for `container-feeder` we replace this with a plain `cmd.run` that runs `systemctl start container-feeder`.

kiall:
WIP as the salt docs make no reference to either a reload or restart param. Need to dig into whats actually supported.

salt/container-feeder/init.sls
kiall: I think restart was the original intent  - if CF was started before docker, it could be failing out, when we restart docker, we need to restart CF to be sure it starts clean from the beginning.

ereslibre: I think the magic is on the `mod_watch` on `service.py`: https://github.com/saltstack/salt/blob/2016.11/salt/states/service.py#L596-L615

It should be `full_restart`. Even if we dropped to `cmd.run` it should be ran after docker has been (re)started since it's requiring it and also watching it.

PR #403, opened 2018-02-27
Don't allow docker restart/kill failures to fail the orch

kiall:
This avoids a race condition between docker ps and docker kill/restart.

### DIFF
--- a/salt/haproxy/init.sls
+++ b/salt/haproxy/init.sls
@@ -53,7 +53,9 @@ haproxy_restart:
     - name: |-
         haproxy_id=$(docker ps | grep -E "k8s_haproxy.*_kube-system_" | awk '{print $1}')
         if [ -n "$haproxy_id" ]; then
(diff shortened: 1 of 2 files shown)


kiall:
this one concerns me a bit - what if docker is starting the container and is already using the old copy of the file that triggered the restart/kill - e.g. haproxy.cfg?

Anyone have a better option?

ereslibre:
There's another option, not sure if better. We can loop until we certainly kill the container ourselves, until then, we loop and for us to succeed, our own `docker kill` command should have succeeded.

The only path that can end up with haproxy running an old config would be:

1. docker spawns haproxy
2. We update the config file
3. `docker ps | grep haproxy` does not succeed (haproxy container status: Created) -> we skip the `docker kill`
4. haproxy container is finally up with the old config (this also depends on *when* haproxy loads the config from the filesystem, as we are mounting haproxy config as directory)

PR #407, opened 2018-02-28
Mark nodes with the `etcd` roles in the update process too.

ereslibre:
Not only on the bootstrap, but when updating from a 2.0 cluster to a 3.0
we need to mark explicitly the `etcd` roles.

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -1,3 +1,17 @@
+{%- set additional_etcd_members = salt.caasp_etcd.get_additional_etcd_members() %}
+
+{% if additional_etcd_members|length > 0 %}
+# Mark some machines as new etcd members
+set-etcd-roles:
+  salt.function:
+    - tgt: {{ additional_etcd_members|join(',') }}
+    - tgt_type: list
+    - name: grains.append
+    - arg:
+      - roles
+      - etcd
+{% endif %}
+
 admin-apply-haproxy:
   salt.state:
     - tgt: 'roles:admin'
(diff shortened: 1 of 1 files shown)


salt/orch/update.sls
```diff
@@ -1,3 +1,19 @@
...
+  salt.function:
+    - tgt: {{ additional_etcd_members|join(',') }}
+    - tgt_type: list
+    - name: grains.append
+    - arg:
+      - roles
+      - etcd
+{% endif %}
+
 admin-apply-haproxy:
```
kiall: Should we add a "require" to this state, so that if setting etcd roles fails, the upgrade doesn't proceed?

ereslibre: Yes, will update.

salt/orch/update.sls
```diff
@@ -1,3 +1,19 @@
+{# the number of etcd masters that should be in the cluster #}
+{%- set num_etcd_members = salt.caasp_etcd.get_cluster_size() %}
```
inercia: `num_etcd_members` is not used, right?

salt/orch/update.sls
ereslibre: Done.

salt/orch/update.sls
ereslibre: Yes, updated.

ereslibre:
This patch **should not be necessary**. I'm staring at https://github.com/kubic-project/salt/blob/07aada2d4fbf0dd08a7562cdac6ab156376e9fb8/salt/etcd/update-pre-orchestration.sls wondering why this didn't work.

PR #408, opened 2018-02-28
Only remove the `kubelet:should_uncordon` grain when we actually uncordon the node.

ereslibre:
As part of the update process, we are cordoning the nodes, so they don't get new jobs when
we are planning to reboot them. If an update fails for whatever reason, it might happen that
we didn't uncordon the node, but removed the `kubelet:should_uncordon` grain. This would
cause that subsequent retries will never uncordon the worker node again, because without this
grain we'll think that this node was cordoned by the user and will not take any action.


ereslibre:

PR #410, opened 2018-03-01
reload on daemon.json changes

MaximilianMeister:
## Depends on
https://github.com/kubic-project/velum/pull/447

### DIFF
--- a/salt/docker/init.sls
+++ b/salt/docker/init.sls
@@ -78,18 +78,22 @@ docker:
     - append_if_not_found: True
     - require:
       - pkg: docker
-  cmd.run:
-    - name: systemctl restart docker.service
-    - onlyif: systemctl status docker.service
-    - onchanges:
-      - /etc/systemd/system/docker.service.d/proxy.conf
-      - /etc/docker/daemon.json
-    - require:
-      - file: /etc/sysconfig/docker
-      - file: /etc/docker/daemon.json
   service.running:
     - enable: True
     - watch:
       - pkg: docker
       - file: /etc/sysconfig/docker
+
+docker-restart-config:
+  cmd.run:
+    - name: systemctl restart docker.service
+    - onlyif: systemctl status docker.service
+    - onchanges:
       - /etc/systemd/system/docker.service.d/proxy.conf
+
+docker-reload-config:
+  service.running:
+    - name: docker
+    - reload: True
+    - onchanges:
+      - file: /etc/docker/daemon.json
--- a/salt/docker/init.sls
+++ b/salt/docker/init.sls
@@ -84,12 +84,13 @@ docker:
       - pkg: docker
       - file: /etc/sysconfig/docker
 
-docker-restart-config:
+docker-proxy-config:
   cmd.run:
-    - name: systemctl restart docker.service
+    - name: systemctl daemon-reload && systemctl restart docker.service
     - onlyif: systemctl status docker.service
     - onchanges:
       - /etc/systemd/system/docker.service.d/proxy.conf
+      - /etc/sysconfig/docker
 
 docker-reload-config:
   service.running:
--- a/salt/container-feeder/init.sls
+++ b/salt/container-feeder/init.sls
@@ -10,6 +10,6 @@ container-feeder:
   service.running:
     - enable: True
     - require:
-      - cmd: docker
+      - pkg: docker
     - watch:
       - service: docker


MaximilianMeister:
[velum: allow users to configure on-premise registry mirrors](https://trello.com/c/VlUFFwDn/477-velum-allow-users-to-configure-on-premise-registry-mirrors)

kiall:
Are we going to end up with a million different orchestrations for different things to apply? e.g. salt/orch/update-registries.sls - salt/orch/update-proxies.sls salt/orch/update-networking.sls etc etc.

Should we consider calling this something like "apply-configurations.sls" ?

MaximilianMeister:
> Should we consider calling this something like "apply-configurations.sls" ?

updated

salt/orch/apply-configurations.sls
```diff
@@ -0,0 +1,16 @@
...
+      - docker
+
+reload-docker-daemon:
+  salt.state:
+    - tgt: 'roles:kube-(master|minion)'
+    - tgt_type: grain_pcre
+    - sls:
+      - docker.reload
+    - require:
+      - salt: apply-docker-config
```
inercia: The whole file could be reduced to:

```SaltStack
# update registries
apply-docker-config:
  salt.state:
    - tgt_type: grain_pcre
    - sls:
      - docker
      - docker.reload
```

And that makes me wonder if we really need a new orchestration for this. Maybe we could do the `docker.reload` with some call to the Salt API...

MaximilianMeister: thx for the hint. looking at it again i also believe we should include the admin in the target...

inercia (dismissed):

ereslibre (commented):
Isn't this already covered by https://github.com/MaximilianMeister/salt/blob/39a187123c657f2857d038b1178faa34ebd38bce/salt/docker/init.sls#L81-L89 ? (just saw it)

MaximilianMeister:
> Isn't this already covered by https://github.com/MaximilianMeister/salt/blob/39a187123c657f2857d038b1178faa34ebd38bce/salt/docker/init.sls#L81-L89 ? (just saw it)

ouh that's a bug right there. good catch. fixed in another comment with some explanation

salt/docker/init.sls
```diff
@@ -83,13 +83,20 @@ docker:
...
       - file: /etc/sysconfig/docker
       - /etc/systemd/system/docker.service.d/proxy.conf
+
+docker-reload-config:
+  cmd.run:
+    - name: systemctl reload docker.service
+    - onlyif: systemctl status docker.service
+    - onchanges:
+      - file: /etc/docker/daemon.json
+    - require:
```
inercia: The `require` here means that the `docker:` state will be executed and, on success, the `docker-reload-config` will be executed too, and that is not what we want: we want to do the reload _iff_ `/etc/docker/daemon.json` changes. So I would remove the `require` and only leave the `onchanges`

MaximilianMeister: @inercia fixed, thanks

salt/docker/init.sls
```diff
@@ -83,13 +83,18 @@ docker:
     - onlyif: systemctl status docker.service
     - onchanges:
       - /etc/systemd/system/docker.service.d/proxy.conf
```
ereslibre: I wonder if we can also use `reload` when `proxy.conf` changes, and we can basically use the same state for any change on the docker configuration.

salt/docker/init.sls
```diff
@@ -83,13 +83,18 @@ docker:
...
-      - /etc/docker/daemon.json
     - require:
       - file: /etc/sysconfig/docker
-      - file: /etc/docker/daemon.json
```
inercia: I would take this block out and create a `docker-restart-config` state, very similar to the `docker-reload-config`:

```SaltStack
docker-restart-config:
   cmd.run:
     - name: systemctl restart docker.service
     - onlyif: systemctl status docker.service
     - onchanges:
       - /etc/systemd/system/docker.service.d/proxy.conf
```

ereslibre: I'd have to test it, but we can use `service.running` I guess:

```SaltStack
docker:
  service.running:
    - onchanges:
      - /etc/systemd/system/docker.service.d/proxy.conf
```

This should issue a `restart` when `/etc/systemd/system/docker.service.d/proxy.conf` changes.

Also,

```SaltStack
docker:
  service.running:
    - reload: True
    - onchanges:
      - /etc/systemd/system/docker.service.d/proxy.conf
```

This should issue a `reload` when `/etc/systemd/system/docker.service.d/proxy.conf` changes.

ereslibre: Reference: https://github.com/saltstack/salt/blob/2016.11/salt/states/service.py#L596-L615

inercia: I think that to have `service.running`, `reload` and `onchanges` in the same state is rather cryptic...

salt/docker/init.sls
MaximilianMeister: we need to do a daemon-reload and a restart, see: https://docs.docker.com/config/daemon/systemd/#httphttps-proxy

updated the code

inercia (dismissed):
LGTM

salt/docker/init.sls
ereslibre: I didn't mean to have them on the same state but rather use `service.running` than direct `systemctl` commands.

ereslibre (dismissed):
LGTM, despite `docker-reload-config` could look like:

```SaltStack
docker-reload-config:
  service.running:
    - name: docker
    - reload: True
    - onchanges:
      - file: /etc/docker/daemon.json
```

MaximilianMeister:
@ereslibre updated

ereslibre (dismissed):
LGTM, thanks!

inercia (dismissed):

salt/docker/init.sls
```diff
@@ -78,18 +78,22 @@ docker:
...
     - watch:
       - pkg: docker
       - file: /etc/sysconfig/docker
+
+docker-proxy-config:
+  cmd.run:
+    - name: systemctl daemon-reload && systemctl restart docker.service
+    - onlyif: systemctl status docker.service
+    - onchanges:
       - /etc/systemd/system/docker.service.d/proxy.conf
```
inercia: Should we add `/etc/sysconfig/docker` here too? I guess the docker daemon should be restarted when that file changes...

MaximilianMeister: yes, it got lost in all the revisions ;) thanks. only thing is it probably doesnt need a `daemon reload`, so should we put it in yet another section or keep it in the same one?

inercia: Bah, I would just put it here...

ereslibre: If a reload of docker is enough (I don't know), we could just add it to the onchanges section of `docker-reload-config`

MaximilianMeister: it needs a restart due to the possible cmd line changes, so a reload is not enough unfortunately.
added it back to the `docker-proxy-config`

ereslibre: I agree with @inercia, we can just put it on `docker-proxy-config`... the `daemon-reload` shouldn't be a problem... not needed at all but...

inercia (dismissed):

MaximilianMeister:
@inercia i needed to add another commit to fix the container-feeder requisite because we moved the `cmd` section out of the `docker` state.

ereslibre (approved):
LGTM

inercia (approved):

PR #413, opened 2018-03-01
Discover everything (_states, _modules...) when we update, not only modules

ereslibre:
When we update from 2.0 to 3.0, not only `_modules` have changed, but for example,
`caasp_cmd` has been introduced in `_states`. This can make the orchestration fail:

```
              ID: uncordon-node
        Function: caasp_cmd.run
            Name: kubectl uncordon caasp-worker-0

          Result: False
         Comment: State 'caasp_cmd.run' was not found in SLS 'kubelet'
                  Reason: 'caasp_cmd.run' is not available.
         Changes:
```

So, make sure that we run `sync_all` on the update process too, and we discover all
the new modules in general, not only `_modules`. After forcefully running `sync_all`
some minions discovered new states.

```
dbf5c05dc8c74fbb8df1fccb105fe3bc:
    ----------
    beacons:
    engines:
    grains:
    log_handlers:
    modules:
    output:
    proxymodules:
    renderers:
    returners:
    sdb:
    states:
        - states.caasp_cmd
        - states.caasp_service
    utils:
```


ereslibre:

PR #415, opened 2018-03-02
Ensure external_fqdn is not rendered to /etc/hosts if it's an IP

kiall:


salt/etc-hosts/hosts.jinja
```diff
@@ -1,10 +1,13 @@
 {% set internal_fqdn = "api api." + pillar['internal_infra_domain'] -%}
+{% set external_fqdn = "" -%}
+{% if not salt['caasp_filters.is_ip'](pillar['api']['server']['external_fqdn']) %}
```
inercia: This is not the most efficient way to do it, as we are querying the pillar twice...

kiall: Pillar data is a in memory dictionary at this point - readability is significantly more important than an extra dictionary lookup IMO.

inercia:
@kiall this would need to be rebased...

mjura:
[Bug 1080608 - Caasp v3 installation failure on openstack](https://trello.com/c/y1FxdkcY/656-bug-1080608-caasp-v3-installation-failure-on-openstack)

PR #417, opened 2018-03-03
Update `etcd` certificates before updating any machine

ereslibre:
We need to include the new SAN on all the certificates before restarting
the first machine. Otherwise, this machine (a master) can find itself
isolated without being able to contact any etcd member with the name it
has (as the rest of the nodes haven't updated their certificates yet to
also include the new name on the SAN).

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -83,6 +83,33 @@ pre-orchestration-migration:
     - require:
       - update-modules
 
+# NOTE: Remove me for 4.0
+#
+# During an upgrade from 2.0 to 3.0, as we go master by master first, the first master will not
+# succeed on the orchestration if it doesn't have an etcd member. Assume M{1,2,3}, W{1,2}. Assume
+# etcd members are running on M2, W1 and W2.
+#
+# M1 updates its configurations on highstate and refers to the etcd nodes with the new names
+# (hostnames) instead of machine-ids, but M2, W1 and W2 still didn't run anything to refresh their
+# certificates and their etcd instances, what will make M1 fail because it cannot connect to any
+# etcd instance (as all certificates look invalid at this point for M2.hostname, W1.hostname and
+# W2.hostname). This makes the apiserver on M1 fail restarting itself until the orchestration reaches
+# M2 [there's no hard dependency on states between masters], but the orchestration already failed on
+# M1, so the global result will be failure nevertheless.
+#
+# Let's force etcd to refresh certificates on all machines, restarting the etcd service so we can
+# continue with the upgrade, as certificates will be valid for the old and the new SAN.
+etcd-setup:
+  salt.state:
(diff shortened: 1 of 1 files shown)


ereslibre:

salt/orch/update.sls
```diff
@@ -83,6 +83,33 @@ pre-orchestration-migration:
...
+#
+# Let's force etcd to refresh certificates on all machines, restarting the etcd service so we can
+# continue with the upgrade, as certificates will be valid for the old and the new SAN.
+etcd-setup:
+  salt.state:
+    - tgt: 'roles:etcd'
+    - tgt_type: grain
+    - sls:
+      - etcd
+    - batch: 1
```
inercia: Any reason for updating them one by one? Couldn't we run this step with `batch: {{ salt.caasp_etcd.get_cluster_size() }}`?

salt/orch/update.sls
```diff
@@ -83,6 +83,33 @@ pre-orchestration-migration:
...
+# succeed on the orchestration if it doesn't have an etcd member. Assume M{1,2,3}, W{1,2}. Assume
+# etcd members are running on M2, W1 and W2.
+#
+# M1 updates its configurations on highstate and refers to the etcd nodes with the new names
+# (hostnames) instead of machine-ids, but M2, W1 and W2 still didn't run anything to refresh their
+# certificates and their etcd instances, what will make M1 fail because it cannot connect to any
+# etcd instance (as all certificates look invalid at this point for M2.hostname, W1.hostname and
+# W2.hostname). This makes the apiserver on M1 fail restarting itself until the orchestration reaches
+# M2 [there's no hard dependency on states between masters], but the orchestration already failed on
+# M1, so the global result will be failure nevertheless.
```
inercia: Cound't we just update `/etc/hosts` in all the masters in the cluster prior to starting the update orchestration? The new `/etc/hosts` will contain both hostnames and machine-ids...

salt/orch/update.sls
ereslibre: It's better to do it safely so we don't restart all etcd instances at once. In any case, the `batch: 1` while being safer has negligible impact in terms of timing overall.

kiall: That would mean all etcd's could restart at the same time - we want to keep the cluster up and functional during the upgrade.

salt/orch/update.sls
ereslibre: It wouldn't help, as the certificates on the running etcd members would still not contain the new name in the SANs, what would still isolate M1 when upgrading (in the presented case in the comment).

salt/orch/update.sls
inercia: Right.

salt/orch/update.sls
inercia: I see.

PR #418, opened 2018-03-03
Make kubelet rename migration idempotent.

ereslibre:
If the new name already exists, also do nothing. A faulty update could
make this script fail over and over again because of its `set -e` and
the `kubectl create -f` command failing as the new node name already
exists.


ereslibre:

PR #419, opened 2018-03-05
Run the highstate on the admin after `sync_all` has been called.

ereslibre:
The admin node might use features not yet discovered, make sure we
run `sync_all` before we enforce a `highstate` on the admin node too.

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -6,14 +6,6 @@ admin-apply-haproxy:
     - sls:
       - haproxy
 
-admin-setup:
-  salt.state:
-    - tgt: 'roles:admin'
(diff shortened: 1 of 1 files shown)


salt/orch/update.sls
```diff
@@ -68,6 +60,14 @@ update-modules:
     - require:
```
inercia: No really related to your PR, but I was wondering, doesn't this `sync_all` imply all the `refresh_*` we are doing before? Maybe we could remove those stages...

ereslibre: I'll check if they are really duplicated and if we can remove some of those (but not on this PR). Thanks!

PR #420, opened 2018-03-05
Remove the TODO message for using the standard `/opt/cni/bin`.

ereslibre:
Internal constraints won't allow us to use `/opt`, so we'll stick
to `/var/lib/kubelet/cni/bin`.


pillar/cni.sls
```diff
@@ -20,7 +20,5 @@ flannel:
 cni:
   plugin:         'flannel'
   dirs:
-    #bin:          '/opt/cni/bin'
-    # TODO: use the standard directory
```
inercia: The idea of using the standard directory is valid, So I would leave the comment or at least say why we are not using the standard directory...

ereslibre: Good point, updating.

PR #423, opened 2018-03-07
When populating the cache, don't fail if this fails for some reason.

ereslibre:
There's a race condition in which the cache directory does not exist, but
when tried to be created it has already been created by something else, and
an exception is raised, stopping the execution.

When populating the cache, we don't really care if it was correctly populated
or not in that *specific* call, so move on.

Fixes: bsc#1084441

### DIFF
--- a/salt/_pillar/velum.py
+++ b/salt/_pillar/velum.py
@@ -74,7 +74,10 @@ def ext_pillar(minion_id,
                                   decode_type='json')
 
     if 'dict' in data:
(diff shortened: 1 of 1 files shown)


ereslibre:
Pushed again to add the `Fixes` line on the commit message.

ereslibre:
[Bug 1084441 - [Build 0038] Bootstrap fails on missing cache directory](https://trello.com/c/v68Z2mXR/700-bug-1084441-build-0038-bootstrap-fails-on-missing-cache-directory)

PR #424, opened 2018-03-08
Make HAProxy work as an http proxy instead of a tcp proxy.

ereslibre:
This allows us to add fine-grained timeouts depending on the endpoint being
accessed or with what parameters (e.g. /log?follow=true should have no
timeout as happens on the apiserver). /exec is another example, but in this
case the protocol is upgraded to spdy.

Depends on https://github.com/kubic-project/caasp-container-manifests/pull/164
Depends on https://github.com/kubic-project/automation/pull/377

Fixes: bsc#1071994

### DIFF
--- a/salt/haproxy/haproxy.cfg.jinja
+++ b/salt/haproxy/haproxy.cfg.jinja
@@ -9,28 +9,56 @@ global
 
 defaults
         log     global
-        mode    tcp
-        option  tcplog
+        mode    http
+        option  redispatch
         option  dontlognull
-        timeout connect 5000
-        timeout client 120000
-        timeout server 120000
+        balance roundrobin
+        timeout connect      5s
+        timeout client       120s
+        timeout http-request 30s
+        timeout client-fin   30s
+        timeout server       120s
+        timeout tunnel       0
+        default-server inter 10s fall 2
 
 # Listen on the standard Kube-API Public port, 6443 by default, and proxy to the masters on
 # the Kube-API internal port, 6444 by default.
-listen kubernetes-master
-        bind {{ bind_ip }}:{{ pillar['api']['ssl_port'] }}
-        mode tcp
-        default-server inter 10s fall 2
-        balance roundrobin
-        option redispatch
+frontend kubernetes-master
+        bind {{ bind_ip }}:{{ pillar['api']['ssl_port'] }} ssl crt {{ pillar['ssl']['kube_apiserver_proxy_bundle'] }} ca-file /etc/pki/ca.crt verify optional
+        timeout client 0
+        http-request set-header X-Remote-User  %{+Q}[ssl_c_s_dn(cn)] if { ssl_c_used ssl_c_verify 0 }
+        http-request set-header X-Remote-Group %{+Q}[ssl_c_s_dn(o)] if { ssl_c_used ssl_c_verify 0 }
+        # If no certificate is passed, or if it's invalid, remove the auth headers
+        http-request del-header X-Remote-User  unless { ssl_c_used ssl_c_verify 0 }
+        http-request del-header X-Remote-Group unless { ssl_c_used ssl_c_verify 0 }
+        http-request del-header X-Remote-Extra unless { ssl_c_used ssl_c_verify 0 }
+        acl streaming_logs url_reg [?&]follow=true
+        acl streaming_logs url_reg [?&]watch=true
+        use_backend no-timeout-backend if streaming_logs
+        acl interactive_session url_reg [?&]tty=true
+        use_backend no-timeout-backend if interactive_session
+        default_backend default-backend
+
+backend default-backend
+        option forwardfor
         option httpchk GET /healthz
+{% for minion_id, nodename in salt['mine.get']('roles:kube-master', 'nodename', 'grain').items() %}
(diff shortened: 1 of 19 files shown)


ereslibre:
I will resume this work when I have more time. This is interesting for the conformance tests as we wouldn't have to do tricks in order to find out whether the tests have finished yet or not. Also, it's fixing a bug with `kubectl exec` (not this specific PR at the moment, but I plan to improve it to also cover the `kubectl exec` case) in which the connection timeouts on the proxy.

ereslibre:
[Bug 1071994 - kubectl run --attach exits early](https://trello.com/c/nNNdxJAn/704-bug-1071994-kubectl-run-attach-exits-early)

ereslibre:
The reason why this is failing on CI is because when rebooting the admin node to update, `admin-node-setup.sh` runs on the admin node on bootup writing old manifests, so haproxy cannot start (interrupting the real update of the cluster as Velum will be down):

https://github.com/kubic-project/caasp-container-manifests/blob/2ce28547afef995aabe6fc132f585d79ece3009a/admin-node-setup.sh#L33-L48

Basically the issue is that we have new `haproxy` configuration with old manifests (with old mounts), what means that `haproxy` cannot find on the new locations the configuration as we copied over the old manifests with the old mounts.

salt/top.sls
```diff
@@ -9,6 +9,7 @@ base:
     - proxy
     - rebootmgr
     - transactional-update
+    - kube-apiserver-proxy
```
inercia: Do we need this new top-level dir? It looks to me the code in  `salt/kube-apiserver-proxy/init.sls` (it is just a certificca generation) could be just moved to `salt/haproxy/init.sls`

salt/haproxy/haproxy.cfg.jinja
```diff
@@ -9,58 +9,77 @@ global
...
+        http-request set-header X-Remote-User  %{+Q}[ssl_c_s_dn(cn)] if { ssl_c_used ssl_c_verify 0 }
+        http-request set-header X-Remote-Group %{+Q}[ssl_c_s_dn(o)] if { ssl_c_used ssl_c_verify 0 }
+        # If no certificate is passed, or if it's invalid, remove the auth headers
+        http-request del-header X-Remote-User  unless { ssl_c_used ssl_c_verify 0 }
+        http-request del-header X-Remote-Group unless { ssl_c_used ssl_c_verify 0 }
+        http-request del-header X-Remote-Extra unless { ssl_c_used ssl_c_verify 0 }
+        acl streaming_logs url_reg [?&]follow=true
+        use_backend no-timeout-backend if streaming_logs
+        acl interactive_session url_reg [?&]tty=true
+        use_backend no-timeout-backend if interactive_session
```
inercia: (I'm not an haproxy expert, so) is this setting the `haproxy` as the SSL termination? Will the `haproxy`<->`master` traffic be in plain HTTP?

kiall:
> The reason why this is failing on CI is because when rebooting the admin node to update, admin-node-setup.sh runs on the admin node on bootup writing old manifests, so haproxy cannot start (interrupting the real update of the cluster as Velum will be down):

So - do we need a matching change to the haproxy manifest in caasp-container-manifests?

salt/top.sls
ereslibre: Yes, I'll move it.

salt/haproxy/haproxy.cfg.jinja
ereslibre: It's TLS terminated.

ereslibre:
> So - do we need a matching change to the haproxy manifest in caasp-container-manifests?

No. My idea was to fall back to `/etc/pki/ca.crt` on the `haproxy` config, as we cannot use `/etc/pki/trust/anchors/SUSE_CaaSP_CA.crt` (`ssl:ca_file`) on caasp-container-manifests because on the first run of the admin node this file doesn't exist and I don't want to add more glue to `caasp-container-manifests`.

The fix should be trivial, but I just set this aside despite almost done because we have a number of other pressing issues :(

ereslibre:
That being said, updating `caasp-container-manifests` might be needed, but not for fixing this issue in particular, for the certificate I'd just fall back to `/etc/pki/ca.crt`.

salt/haproxy/haproxy.cfg.jinja
inercia: Ok, TLS, but I mean that the `haproxy`<->`master` communication will be in plain HTTP, right?

ereslibre: No, that communication happens also over TLS.

ereslibre:
This won't pass on CI until https://github.com/kubic-project/caasp-container-manifests/pull/164 is on the image.

ereslibre:
Rebased and conflicts fixed. Will address comments at a later time.

ereslibre:
This also depends on https://github.com/kubic-project/automation/pull/377

ereslibre:
Please, do not merge: figuring out why the two PR's together aren't passing in Jenkins: https://github.com/kubic-project/jenkins-library/pull/170

ereslibre:
Passing with: https://github.com/kubic-project/jenkins-library/pull/170

ereslibre:
@flavio @kiall: We need to force merge this one. Passing on https://github.com/kubic-project/jenkins-library/pull/170

PR #427, opened 2018-03-12
Ensure Dex certificate rotations work as expected

kiall:
When the Dex config / certs are changed, nothing would trigger
the restart of the already running Dex pods in order to have Dex
load it's new config, or certificates.

A common pattern for this is to use checksum annotations on the pods,
when the annotation changes, Kubernetes will start the rolling upgrade
of the pods - restarting them according to their rolling upgrade
policy.

Fixes bsc#1082996


salt/dex/dex.yaml
```diff
@@ -1,3 +1,79 @@
...
+  - id: velum
+    redirectURIs:
+    - 'https://{{ pillar['dashboard'] }}/oidc/done'
+    - 'https://{{ pillar['dashboard_external_fqdn'] }}/oidc/done'
+    name: "Velum"
+    secret: "{{ pillar['dex']['client_secrets']['velum'] }}"
+{%- endmacro -%}
+
+{%- macro dex_secret() -%}
+{%- import_text pillar['ssl']['dex_crt'] as crt -%}
```
inercia: I think you can just do:

```Jinja
{%- set crt = pillar['ssl']['dex_crt'] %}
```

salt/dex/init.sls
```diff
@@ -5,7 +5,7 @@ include:
   - kube-apiserver
 
 {% from '_macros/certs.jinja' import alt_master_names, certs with context %}
-{% from '_macros/kubectl.jinja' import kubectl, kubectl_apply_template with context %}
+{% from '_macros/kubectl.jinja' import kubectl, kubectl_apply_template, kubectl_apply_dir_template with context %}
```
ereslibre: We can get rid of `kubectl_apply_template` here.

salt/orch/removal.sls
```diff
@@ -294,10 +294,10 @@ cleanups-in-target-before-rebooting:
       - kube-apiserver.remove-pre-reboot
       - kube-controller-manager.remove-pre-reboot
       - kube-scheduler.remove-pre-reboot
-      - dex.remove-pre-reboot
       - addons.remove-pre-reboot
       - addons.dns.remove-pre-reboot
       - addons.tiller.remove-pre-reboot
+      - addoms.dex.remove-pre-reboot
```
ereslibre: Typo.

kiall: Fixed

PR #428, opened 2018-03-13
Add PodSecurityPolicy Support

kiall:
Add support for PodSecurityPolicy's, allowing us to disable use of the
hostPath volume type.

This change adds 2 PSP's:

* unprivileged (Default assigned to all users)

The unprivileged PodSecurityPolicy is intended to be a
reasonable compromise between the reality of Kubernetes workloads, and
suse:caasp:psp:privileged. By default, we'll grant this PSP to all
users and service accounts.

* privileged

The privileged PodSecurityPolicy is intended to be given
only to trusted workloads. It provides for as few restrictions as possible
and should only be assigned to highly trusted users.

Fixes bsc#1047535

### DIFF
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -67,6 +67,7 @@ kube_log_level:   '2'
 
 # install the addons (ie, DNS)
 addons:
+  psp:    'true'
   dns:    'true'
   tiller: 'false'
 
--- a/salt/addons/dns/init.sls
+++ b/salt/addons/dns/init.sls
@@ -1,4 +1,4 @@
-{% if salt.caasp_pillar.get('addons:dns', False) %}
+{% if salt.caasp_pillar.get('addons:dns', True) %}
 
 include:
   - kube-apiserver
--- a/salt/addons/psp/manifests/10-podsecuritypolicy-privileged.yaml
+++ b/salt/addons/psp/manifests/10-podsecuritypolicy-privileged.yaml
@@ -1,0 +1,77 @@
+# The privileged PodSecurityPolicy is intended to be given
+# only to trusted workloads. It provides for as few restrictions as possible
+# and should only be assigned to highly trusted users.
+---
+apiVersion: extensions/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: suse.caasp.psp.privileged
+  annotations:
+    seccomp.security.alpha.kubernetes.io/defaultProfileName: docker/default
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: '*'
+    apparmor.security.beta.kubernetes.io/defaultProfileName: runtime/default
+spec:
+  # Privileged
+  privileged: true
+  # Volumes and File Systems
+  volumes:
+    # Kubernetes Pseudo Volume Types
+    - configMap
+    - secret
+    - emptyDir
+    - downwardAPI
+    - projected
+    - persistentVolumeClaim
+    # Kubernetes Host Volume Types
+    - hostPath
+    # Networked Storage
+    - nfs
+    - rbd
+    - cephFS
+    - glusterfs
+    - fc
+    - iscsi
+    # Cloud Volumes
+    - gcePersistentDisk
+    - awsElasticBlockStore
+    - azureDisk
+    - azureFile
+    - vsphereVolume
+  allowedFlexVolumes: []
+  #allowedHostPaths: []
+  readOnlyRootFilesystem: false
+  # Users and groups
+  runAsUser:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  fsGroup:
+    rule: RunAsAny
+  # Privilege Escalation
+  allowPrivilegeEscalation: true
+  defaultAllowPrivilegeEscalation: true
+  # Capabilities
+  allowedCapabilities:
+    - '*'
+  defaultAddCapabilities: []
+  requiredDropCapabilities: []
+  # Host namespaces
+  hostPID: true
+  hostIPC: true
+  hostNetwork: true
+  hostPorts:
+  - min: 0
+    max: 65535
+  seLinux:
+    # SELinux is unsed in CaaSP
+    rule: 'RunAsAny'
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: suse:caasp:psp:privileged
+rules:
+  - apiGroups: ['extensions']
+    resources: ['podsecuritypolicies']
+    verbs: ['use']
+    resourceNames: ['suse.caasp.psp.privileged']
--- a/salt/addons/psp/manifests/10-podsecuritypolicy-unprivileged.yaml
+++ b/salt/addons/psp/manifests/10-podsecuritypolicy-unprivileged.yaml
@@ -1,0 +1,82 @@
+# The unprivileged PodSecurityPolicy is intended to be a
+# reasonable compromise between the reality of Kubernetes workloads, and
+# suse:caasp:psp:privileged. By default, we'll grant this PSP to all
+# users and service accounts.
+---
+apiVersion: extensions/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: suse.caasp.psp.unprivileged
+  annotations:
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: docker/default
+    seccomp.security.alpha.kubernetes.io/defaultProfileName: docker/default
+    apparmor.security.beta.kubernetes.io/allowedProfileNames: runtime/default
+    apparmor.security.beta.kubernetes.io/defaultProfileName: runtime/default
+spec:
+  # Privileged
+  privileged: false
+  # Volumes and File Systems
+  volumes:
+    # Kubernetes Pseudo Volume Types
+    - configMap
+    - secret
+    - emptyDir
+    - downwardAPI
+    - projected
+    - persistentVolumeClaim
+    # Networked Storage
+    - nfs
+    - rbd
+    - cephFS
+    - glusterfs
+    - fc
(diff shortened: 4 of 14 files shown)


salt/addons/dns/init.sls
```diff
@@ -1,4 +1,4 @@
-{% if salt.caasp_pillar.get('addons:dns', False) %}
+{% if salt.caasp_pillar.get('addons:dns', True) %}
```
kiall: This was defaulting in the static pillar to True - sync'd it up while I noticed it.

inercia:
I think this should reference [this bugzilla issue](https://bugzilla.suse.com/show_bug.cgi?id=1047535)

salt/addons/psp/podsecuritypolicy-privileged.yaml.jinja
```diff
@@ -0,0 +1,72 @@
...
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  fsGroup:
+    rule: RunAsAny
+  # Privilege Escalation
+  allowPrivilegeEscalation: true
+  defaultAllowPrivilegeEscalation: true
+  # Capabilities
+  allowedCapabilities: []
```
ereslibre: As a completely open PSP, this setting could be `'*'` in case a container is not `privileged` but still needs certain capabilities?

kiall: Done

vrothberg:
Some background information for capabilities in Kubernetes. There are three fields in the config:

### DefaultAddCapabilities
"The capabilities which are added to containers by default, in addition to the runtime defaults." This means that any capability we specify here will be added on top of what our runtime of choice is adding by default.

### AllowedCapabilities
This is a list of capabilities that will be granted to a container if it requests it. None of them will be added by default but on request.

### RequiredDropCapabilities
All specified capabilities *must* be dropped from the container. It is important that no capability in this list is also referenced in the two above ones. Notice: if our runtime of choice adds `CAP_FOO` by default, it will be removed if specified in this list.

In the context of capabilities, having (only) three profiles for different security levels is really tough. That's just my 2 cents, but I would prefer to provide concrete profiles for specific use cases and workloads (e.g., web server, database).

salt/addons/psp/manifests/10-podsecuritypolicy-privileged.yaml
```diff
@@ -0,0 +1,75 @@
+# The privileged PodSecurityPolicy is intended to be given
+# only to trusted workloads. It provides for as few restrictions as possible
+# and should only be assigned to highly trusted users.
+---
+apiVersion: extensions/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: privileged
+  annotations: {}
```
cyphar: I believe you need to explicitly specify `seccomp.security.alpha.kubernetes.io/allowedProfileNames: '*'`. According to the docs, not including this annotation means the default cannot be changed.

kiall:
@vrothberg

> In the context of capabilities, having (only) three profiles for different security levels is really tough. That's just my 2 cents, but I would prefer to provide profiles tailored for specific use cases and workloads (e.g., web server, database).

Agree - however, we need to provide some generic defaults that suit "most" workloads. Operators are free to add extra policies with more specific restrictions or allowances. I see these policies as existing to provide some defaults which can either be used directly, or better yet - can be used as a template for more specific requirements.

salt/addons/psp/manifests/10-podsecuritypolicy-lockdown.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - downwardAPI
+    - projected
+    - persistentVolumeClaim
+  allowedFlexVolumes: []
+  allowedHostPaths:
+    # Note: We don't allow hostPath volumes above, but set this to a path we
+    # control anyway as a belt+braces protection. /dev/null may be a better
+    # option, but the implications of pointing this towards a device are
+    # unclear.
+    - pathPrefix: /opt/kubernetes-hostpath-volumes
```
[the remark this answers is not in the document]
kiall: > If we have to point this at a path, I would point it at one that we know is going to never be used (and is obvious should never be used). Something like /tmp/caasp-kubernetes-locked-down-hostpath

Yea - I like the notion of making it clear this is not intended to be useable. I'll make it something like `/invalid-kubernetes-hostpaths/` - given the RO root FS, and that name, it will be unusable and obviously unusable  - even if the hostPath volume type restriction is somehow bypassed.

> As for pointing it at a device, there theoretically shouldn't be an issue if pathPrefix handles path targets properly.

Yea, the special-caseness of host devices made me rethink my initial setting of `/dev/null` - I think you're agreeing with me here, that we should avoid referencing the device as it's unclear what that would end up allowing should something, somewhere, have a bug...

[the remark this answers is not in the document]
cyphar: (As an aside, I just took a very quick look how `pathPrefix` is implemented and I'm a little concerned. I am going to take a closer look tomorrow...)

salt/addons/psp/manifests/10-podsecuritypolicy-unprivileged.yaml
```diff
@@ -0,0 +1,84 @@
...
+    # unclear.
+    - pathPrefix: /opt/kubernetes-hostpath-volumes
+  readOnlyRootFilesystem: false
+  # Users and groups
+  runAsUser:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  fsGroup:
+    rule: RunAsAny
```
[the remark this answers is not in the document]
kiall: > Ideally we would want these to be RunAsNonRoot -- but this might break users who (IMO incorrectly) are running containers as root.

For better or worse (okay - just worse), 99% of containers run as root. We want to provide a reasonable default for "unprivileged containers" where operators won't immediately jump to using the `privileged` policy as nothing works.

> But as I mention below, I'm not entirely sure who the target is for the unprivileged PSP.

I'll update the PR commit messages to be more clear on this. In short - unprivileged is meant to be a reasonable set of default privileges for the real-world of workloads our users have. For example, 99% of workloads have no need to drop the network/IPC/PID namespacing, so we don't allow that. However, 99% of real world workloads expect to start up as `root`, so we allow it (e.g. go look at the Helm charts repo - maybe 1% allow specifying the necessary non-root account details)

salt/addons/psp/manifests/10-podsecuritypolicy-unprivileged.yaml
```diff
@@ -0,0 +1,84 @@
...
+  # Privilege Escalation
+  allowPrivilegeEscalation: true
+  defaultAllowPrivilegeEscalation: true
+  # Capabilities
+  allowedCapabilities: []
+  defaultAddCapabilities: []
+  requiredDropCapabilities:
+    - ALL
```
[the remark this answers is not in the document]
kiall: > This is a little bit extreme if you also want to allow users to run as root. Most of the properties of "running as root" are actually capabilities that need to be allowed (such as being able to change the owner of a file).

This is exactly the sort of info I need :) What are a reasonable set of allowances here?

> However there also appears to be a fairly large shortcoming in the capabilities support in PSPs. You cannot whitelist the absolute set of capapbilities, you can only:

Yea, that annoyed me too. How do other runtimes handle this? And - What are the default set docker keeps no matter what?

salt/addons/psp/manifests/10-podsecuritypolicy-unprivileged.yaml
```diff
@@ -0,0 +1,84 @@
...
+  # Host namespaces
+  hostPID: false
+  hostIPC: false
+  hostNetwork: false
+  hostPorts:
+  - min: 0
```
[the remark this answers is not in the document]
kiall: I really wanted to do that :( - but for example, the pretty common apache/nginx containers from dockerhub go straight to port 80. There is no need for them to do this, especially in the context of Kubernetes, but they do. If we make this only allow unprivileged IP ports - then, most workloads will need to be ran under the "privileged" policy.

Maybe this points to a flaw in my policy naming.

salt/addons/psp/manifests/10-podsecuritypolicy-privileged.yaml
kiall: Updated to be explicit about this.

cyphar:
I've added my comments on the `hostPath` issue in kubernetes/kubernetes#61043 (which is tracking the security issue).

pillar/params.sls
```diff
@@ -67,6 +67,7 @@ kube_log_level:   '2'
 
 # install the addons (ie, DNS)
 addons:
+  psp:    'true'
```
inercia: Is the PodSecurityPolicy really a addon? Is it optional? Are we going to expose this flag to the user?

salt/orch/kubernetes.sls
```diff
@@ -181,29 +181,23 @@ kubelet-setup:
...
 
+# we must start CNI before any other pods, as nodes will be NotReady until
+# the CNI DaemonSet is loaded and running...
 services-setup:
   salt.state:
     - tgt: {{ super_master }}
     - sls:
       - addons
+      - addons.psp
+      - cni
```
inercia: Maybe we should create a `/services` folder and move all these things there...

kiall: Yea, I agree - we should rename the addons -> services folder. I won't do that in this PR, but I will create a followup once this merges to do the refactor.

pillar/params.sls
kiall: I'd like it to be optional yes, although I don't think we want to expose that in Velum - I just want is as a "advanced" option for anyone who happens to need it disabled for whatever reason.

PR #430, opened 2018-03-14
Wipe out our /etc/hosts changes before reboot

kiall:
This ensures the systemd/wicked logic is unaffected by our /etc/hosts
changes.

### DIFF
--- a/salt/etc-hosts/init.sls
+++ b/salt/etc-hosts/init.sls
@@ -2,6 +2,7 @@
 {% if salt['grains.get']('virtual_subtype', None) != 'Docker' %}
 /etc/hosts:
   file.blockreplace:
+    # If markers are changed, also update etc-hosts/update-pre-reboot.sls
     - marker_start: "#-- start Salt-CaaSP managed hosts - DO NOT MODIFY --"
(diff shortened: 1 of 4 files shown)


kiall:
> I think it's safer to explicitly run etc-hosts right after the salt-minion has come back if we do this before rebooting, so we don't wait for the highstate to set up /etc/hosts again (ordering would be a mess).

At that point, it's too late. We no longer have the right data to re-render etc-hosts, and systemd/wicked has already changed the hostname to match what was in /etc/hosts between "power on" and salt-minion finally starting.

ereslibre:
> At that point, it's too late. We no longer have the right data to re-render etc-hosts, and systemd/wicked has already changed the hostname to match what was in /etc/hosts between "power on" and salt-minion finally starting.

Yes, what I mean is that in addition to the current change we can also run the `etc-hosts` forcefully right after the minion is back, so we are sure that when we run a `highstate` the `/etc/hosts` is well formed.

kiall:
> Yes, what I mean is that in addition to the current change we can also run the etc-hosts sls forcefully right after the minion is back, so we are sure that when we run a highstate the /etc/hosts is well formed.

Ahh - makes sense. Will update tomorrow.

PR #432, opened 2018-03-16
PCRE grain expressions only allow the regexp on the value side.

ereslibre:
Fix PCRE grain query expressions so they are matching what we expect.

```
caasp-admin:~ # docker exec -it 06bf salt -P 'bootstrap_complete:.*' cmd.run hostname
admin:
    caasp-admin
6b5cb85d20f94f6eb813449b228cfe13:
    caasp-worker-1
4c0e4d31bc754369940ffcbae28e2f0a:
    caasp-worker-0
cb92123fa85d4170807e0aa24573501b:
    caasp-master-0
66d5844bc5f14d1480896b1bc234dd92:
    caasp-master-1
3f3f505c6eb3464e8a08cc0ae6fbc8f4:
    caasp-master-2
caasp-admin:~ # docker exec -it 06bf salt -P 'bootstrap_.*:true' cmd.run hostname
No minions matched the target. No command was sent, no jid was assigned.
ERROR: No return received
```


ereslibre:
Issue spotted by @kiall.

salt/orch/update-etc-hosts.sls
```diff
@@ -1,4 +1,4 @@
-{%- set updates_all_target = 'P@roles:(admin|kube-(master|minion)) and G@bootstrap_complete:true and not P@.*_in_progress:true' %}
+{%- set updates_all_target = 'P@roles:(admin|kube-(master|minion)) and G@bootstrap_complete:true and not G@update_in_progress:true and not G@removal_in_progress:true' %}
```
kiall: Missing bootstrap_in_progress:true

salt/orch/removal.sls
```diff
@@ -381,7 +381,8 @@ remove-target-salt-key:
   {%- endif %}
 
   {%- set affected_expr = 'G@bootstrap_complete:true' +
-                          ' and not P@.*_in_progress:true' +
+                          ' and not G@update_in_progress:true' +
```
kiall: Missing bootstrap_in_progress:true ?

salt/orch/update-etc-hosts.sls
ereslibre: Oh I skipped it because of `bootstrap_complete:true` but it won't work when re-running the orchestration because of a settings change for example... Fixing.

salt/orch/removal.sls
ereslibre: Yeah... same as the previous one. Fixing.

ereslibre:
Fixed.

ereslibre:
Despite at this point we should add a method `any_action_in_progress` or something similar that concatenates these for us... only having to update one place would be nicer and this pattern will just grow.

inercia:
It seems this is a known issue saltstack/salt#30936

PR #433, opened 2018-03-16
Fix undefined variable

ereslibre:
This fixes master removal on the salt states. I could remove a master with this patch.

* Salt was complaining that this variable didn't exist in the `orch.removal`
orchestration when removing a master when no replacement was provided.

* If no replacement is provided, `sync-all` was trying to refer to states
that didn't exist because those states also were wrapped with a `replacement`
guard.

### DIFF
--- a/salt/orch/removal.sls
+++ b/salt/orch/removal.sls
@@ -4,9 +4,7 @@
 # ... and we can provide an optional replacement node, and
 # this Salt code will always trust that node as a valid replacement
 {%- set replacement = salt['pillar.get']('replacement', '') %}
-{%- if replacement %}
-  {%- set replacement_provided = True %}
-{%- endif %}
+{%- set replacement_provided = (replacement != '') %}
 {%- set replacement_roles = [] %}
 
 
--- a/salt/orch/removal.sls
+++ b/salt/orch/removal.sls
@@ -49,7 +49,7 @@
     {%- do salt.caasp_log.debug('CaaS: setting %s as the replacement for the etcd member %s', replacement, target) %}
     {%- do replacement_roles.append('etcd') %}
   {%- elif etcd_members|length > 1 %}
-    {%- do salt.caasp_log.warn('CaaS: numnber of etcd members will be reduced to %d, as no replacement for %s has been found (or provided)', etcd_members|length, target) %}
+    {%- do salt.caasp_log.warn('CaaS: number of etcd members will be reduced to %d, as no replacement for %s has been found (or provided)', etcd_members|length, target) %}
   {%- else %}
     {#- we need at least one etcd server #}
     {%- do salt.caasp_log.abort('CaaS: cannot remove etcd member %s: too few etcd members, and no replacement found or provided', target) %}
@@ -147,12 +147,18 @@
 {%- endif %} {# target in minions #}
 
 {#- other consistency checks... #}
+
 {%- if replacement %}
   {#- consistency check: if there is a replacement, it must have some (new) role(s) #}
   {%- if not replacement_roles %}
     {%- do salt.caasp_log.abort('CaaS: %s cannot be removed: too few etcd members, and no replacement found', target) %}
   {%- endif %}
-{%- endif %} {# replacement #}
+{%- else %}
+  {#- consistency check: if some of the previous checks set replacement to an empty string, clear replacement_roles #}
+  {%- set replacement_roles = [] %}
+{%- endif %}
+
+{#- end other consistency checks... #}
 
 {##############################
  # set grains


salt/orch/removal.sls
```diff
@@ -204,9 +206,11 @@ sync-all:
       - saltutil.sync_all
     - require:
       - assign-removal-grain
+{%- if replacement %}
   {%- for role in replacement_roles %}
       - assign-{{ role }}-role-to-replacement
   {%- endfor %}
+{%- endif %}
```
inercia: So was this failing beacuse there was no `replacement` but there were `replacement_roles`?

salt/orch/removal.sls
```diff
@@ -6,6 +6,8 @@
 {%- set replacement = salt['pillar.get']('replacement', '') %}
 {%- if replacement %}
   {%- set replacement_provided = True %}
+{%- else %}
+  {%- set replacement_provided = False %}
```
inercia: Maybe Jinja would accept `set replacement_provided = (replacement != '')`

salt/orch/removal.sls
ereslibre: Yes, exactly.

salt/orch/removal.sls
ereslibre: I didn't try that one specifically, will try.

salt/orch/removal.sls
inercia: In that case it doesn't make sense to have some roles for a replacement that does not exist. Maybe it would be better to fix it [here](https://github.com/kubic-project/salt/blob/master/salt/orch/removal.sls#L152-L157), with something like:

```Jinja
{%- if replacement %}
  {#- consistency check: if there is a replacement, it must have some (new) role(s) #}
  {%- if not replacement_roles %}
    {%- do salt.caasp_log.abort('CaaS: %s cannot be removed: too few etcd members, and no replacement found', target) %}
  {%- endif %}
{%- else %}
    {#- remove the roles for a replacement that could not be found #}
  {%- set replacement_roles = [] %}
{%- endif %} {# replacement #}
```

ereslibre: What you propose would make sense, let me check the original problem again. Will update the PR.

ereslibre: When removing a master, the commit https://github.com/kubic-project/salt/pull/433/commits/0196d659fc1e5565706f53a37022149e055a68ff is fixing this error:

```
caasp-admin_master:
    Data failed to compile:
----------
    Rendering SLS 'base:orch.removal' failed: Jinja variable 'replacement_provided' is undefined
```

ereslibre:
When removing a master (without providing a replacement), the commit https://github.com/kubic-project/salt/pull/433/commits/0196d659fc1e5565706f53a37022149e055a68ff is fixing this error:

```
caasp-admin_master:
    Data failed to compile:
----------
    Rendering SLS 'base:orch.removal' failed: Jinja variable 'replacement_provided' is undefined
```

Called like: `docker exec -it 570ba salt-run state.orchestrate orch.removal pillar='{"target":"861617b404364357a87af375aa5b3a16"}'`

ereslibre:
When removing a master (after having fixed the first issue), what was shown was:

```
caasp-admin_master:                                                                                                                                                                                                                                                                                                                                                                                                                 [21/1969]
----------                          
          ID: assign-removal-grain
    Function: salt.function
        Name: grains.setval
      Result: True                                                      
     Comment: Function ran successfully. Function grains.setval ran on 746581d3e86a40da9e792e87f2f133f4.
     Started: 14:43:55.565980
    Duration: 264.095 ms                                                       
     Changes:           
              746581d3e86a40da9e792e87f2f133f4:
                                                                               
              Summary for 746581d3e86a40da9e792e87f2f133f4
              -----------
              Succeeded: 0    
              Failed:   0
              -----------
              Total states run:    0
              Total run time:  0.000 ms
----------              
          ID: sync-all     
    Function: salt.function                                   
        Name: mine.update
      Result: False
     Comment: The following requisites were not found:
                                 require:
                                     id: assign-etcd-role-to-replacement
     Changes:   
(and many others with the same failed require)
```

This is what the second commit fixes.

salt/orch/removal.sls
```diff
@@ -149,12 +147,20 @@
...
 {%- if replacement %}
-  {#- consistency check: if there is a replacement, it must have some (new) role(s) #}
   {%- if not replacement_roles %}
     {%- do salt.caasp_log.abort('CaaS: %s cannot be removed: too few etcd members, and no replacement found', target) %}
   {%- endif %}
-{%- endif %} {# replacement #}
+{%- endif %}
+
+{#- consistency check: if some of the previous checks set replacement to an empty string, clear replacement_roles #}
+{%- if not replacement %}
```
inercia: Why not a simple `{% else %}` instead of `{% endif %} \n {% if not replacement %}` ?

ereslibre: The idea was to keep the consistency checks separate, but I'll change it.

inercia:
LGTM, thanks!

PR #441, opened 2018-03-19
When we explicitly run `haproxy` sls in the update, run `etc-hosts` too.

ereslibre:
* During a rename, it might happen that `haproxy` refuses to start because it cannot
resolve the new names `nodename.infra.caasp.local` in the configuration because its
`/etc/hosts` file hasn't been updated yet.

* Refresh modules before we call to any `sls`, they might use undiscovered modules.

### DIFF
--- a/salt/haproxy/init.sls
+++ b/salt/haproxy/init.sls
@@ -53,7 +53,7 @@ haproxy:
 # Send a SIGTERM to haproxy when the config changes
 # TODO: There should be a better way to handle this, but currently, there is not. See
 # kubernetes/kubernetes#24957
-haproxy_restart:
+haproxy-restart:
   cmd.run:
     - name: |-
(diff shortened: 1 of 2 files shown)


ereslibre:
Issue caused by https://github.com/kubic-project/salt/pull/431. Whereas `nodename` was resolvable, `nodename.caasp.infra.local` wasn't until `/etc/hosts` is updated first.

ereslibre:
In a `2.0` to `3.0` upgrade, running `etc-hosts` without updating modules first failed because `caasp_net` was not yet discovered, so let's first discover everything and then run the required sls/highstates.

ellisab:
Successfully upgraded from v2 devel to v3 devel  (3 masters + 3 minions) with this patch.

ereslibre:
Rebased.

PR #443, opened 2018-03-20
Try to resist transient node failures on updates

inercia:
# Description

Try to resist transient node failures on updates, so users can updte the cluster even if some node X is down.

# Testing

This can be tested by
1) orchestrating
2) stopping a Salt minion in a node `X` (it can be any master or worker)
3) then trying to run the update orchestration (it can be forced with a _fake_ orchestration, setting the `tx_reboot_needed:true` grain in all those nodes)

The update orchestration sould succeed and all the nodes but `X` should be udpated.

See https://trello.com/c/irviWd1m

feature#update_on_node_failures

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -1,50 +1,72 @@
+{#- Get a list of nodes seem to be down or unresponsive #}
+{#- This sends a "are you still there?" message to all #}
+{#- the nodes and wait for a response, so it takes some time. #}
+{#- Hopefully this list will not be too long... #}
+{%- set nodes_down = salt.saltutil.runner('manage.down') %}
+{%- if nodes_down|length >= 1 %}
+# {{ nodes_down|join(',') }} seem to be down: skipping
+  {%- do salt.caasp_log.debug('CaaS: nodes "%s" seem to be down: ignored', nodes_down|join(',')) %}
+  {%- set is_responsive_node_tgt = 'not L@' + nodes_down|join(',') %}
+{%- else %}
+# all nodes seem to be up
+  {%- do salt.caasp_log.debug('CaaS: all nodes seem to be up') %}
+  {#- we cannot leave this empty (it would produce many " and <empty>" targets) #}
+  {%- set is_responsive_node_tgt = '*' %}
+{%- endif %}
+
+{#- some other targets: #}
+
+{#- the regular nodes (ie, not the CA or the admin node) #}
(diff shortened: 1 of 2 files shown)


salt/orch/update.sls
```diff
@@ -26,39 +33,26 @@ generate-sa-key:
     - sls:
       - kubernetes-common.generate-serviceaccount-key
 
-# Generic Updates
-sync-pillar:
-  salt.runner:
-    - name: saltutil.sync_pillar
```
ereslibre: We need this `saltutil.sync_pillar` explicitly with `salt.runner`, that's how the master will load the `_pillars/velum.py` on the master.

inercia: Oh, I missed it. I will add it in the `update-data` block. Thanks.

ereslibre:
I know this logic was implemented for the update, but if we isolate the list of unresponsive minions in a module, we could also use this for the `orch.kubernetes` orchestration (this orchestration is also called when adding nodes, or when applying new settings), so I think we can get advantage of this logic also there. If you think we should do that on a follow-up PR I'd be OK with it though.

inercia:
> I know this logic was implemented for the update, but if we isolate the list of unresponsive minions in a module, we could also use this for the orch.kubernetes orchestration (this orchestration is also called when adding nodes, or when applying new settings), so I think we can get advantage of this logic also there. If you think we should do that on a follow-up PR I'd be OK with it though.

I think it will be difficult to reuse this code as it is very specific to the orchestration in question. Besides that, I'm not sure we can do the same for the `kubernetes` orchestration: we must be very careful with the nodes that can be ignored (specially for some roles), what to do when they are not responsive and so on. And, anyway, any further work on unresponsive nodes should go in subsequent PRs...

salt/orch/update.sls
```diff
@@ -410,7 +433,7 @@ remove-caasp-fqdn-grain:
 
 masters-remove-update-grain:
   salt.function:
-    - tgt: G@roles:kube-master and G@update_in_progress:true
+    - tgt: '{{ is_updateable_master_tgt }}'
```
ereslibre: This one is a little different: `is_updateable_master_tgt` expands to `G@tx_update_reboot_needed:true and * and G@roles:kube-master` whereas this was `G@roles:kube-master and G@update_in_progress:true`. `update_in_progress:true` vs `tx_update_reboot_needed: true`.

inercia: Good catch. I'll fix it. Thanks.

inercia: Done.

PR #449, opened 2018-03-21
Add flannel readiness/liveness probe

kiall:
This makes sure flannel has at least reached the point where it starts the
healthz API endpoint. However, that point in the flannel code is *very* early
and not all that useful for actual health checking. Additionally, as long as
the HTTP gorouting is running, healthz will *always* respond with a 200. It
performs no actual health checking.

Even still, lets include the probe. If flannel gets better health checking,
it will be enabled for us, on the other hand, if flannel doesn't get better
health checking, it's still *very slightly* useful to know that flannel has
at least reached this point in it's code.

### DIFF
--- a/salt/cni/kube-flannel.yaml.jinja
+++ b/salt/cni/kube-flannel.yaml.jinja
@@ -103,8 +103,21 @@ spec:
           - "--kube-subnet-mgr"
           - "--v={{ pillar['flannel']['log_level'] }}"
           - "--iface=$(POD_IP)"
+          - "--healthz-ip=$(POD_IP)"
+          - "--healthz-port={{ pillar['flannel']['healthz_port'] }}"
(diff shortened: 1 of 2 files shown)


salt/cni/kube-flannel.yaml.jinja
```diff
@@ -103,8 +103,21 @@ spec:
           - "--kube-subnet-mgr"
           - "--v={{ pillar['flannel']['log_level'] }}"
           - "--iface=$(POD_IP)"
+          - "--healthz-ip=$(POD_IP)"
+          - "--healthz-port=8081"
```
ereslibre: Nitpick: we might want to pillar this port.

kiall: Fixed

PR #461, opened 2018-03-27
Wait for deployments during the orchestration time.

ereslibre:
Additionally to other checks, we should also consider the orchestration
done once that the expected pods are running.

feature#deployment-stability


ereslibre:
Still not working locally, removing review requests and adding WIP.

ereslibre:
This has succeeded three times in a row in CI: http://jenkins.caasp.suse.net/blue/organizations/jenkins/salt/activity?branch=PR-461

(ignoring http://jenkins.caasp.suse.net/blue/organizations/jenkins/salt/detail/PR-461/10/pipeline, that has nothing to do with this change)

PR #462, opened 2018-03-27
Introduce cri-o support

flavio:
This PR allow to deploy new SUSE CaaS Platform clusters using cri-o as a container runtime instead of docker.

The cluster will keep using docker on the admin node, while all the other nodes are going to use cri-o.

It's not possible to have mixed environments, all nodes have to use the same container runtime.

The CRI can be chosen by setting the value of the `cri:name` pillar, which is defined inside of the `pillar/cri.sls` file. By default `docker` is being used.


pillar/params.sls
```diff
@@ -172,3 +172,6 @@ ldap:
   admin_group_name: ''
   tls_method: ''
   mail_attribute: ''
+
+# infra container to use instead of downloading gcr.io/google_containers/pause
+pod_infra_container_image: sles12/pause:1.0.0
```
flavio: I wonder if that should be `docker.io/sles12/pause:1.0.0` for crio, but apparently it doesn't make a difference

vrothberg: Currently, it's working because we're configuring `/etc/containers/registries.conf` to search in `docker.io` when using unqualified images. It would be safer to always use fully qualified images, but both ways are fine for me atm.

salt/_modules/caasp_cri.py
```diff
@@ -0,0 +1,389 @@
...
+    '''
+    if 'admin' in __salt__['grains.get']('roles', []):
+        return '/var/run/dockershim.sock'
+
+    cri = __salt__['pillar.get']('cri:name', 'docker').lower()
+
+    if cri == 'docker':
+        return '/var/run/dockershim.sock'
+    elif cri == 'crio':
+        return '/var/run/crio/crio.sock'
```
vrothberg: Nitpick: This looks like something we could configure with grains? We reference the endpoints in some config files as well.

salt/_modules/caasp_cri.py
```diff
@@ -0,0 +1,389 @@
...
+        return (
+                False,
+                'Error mounting container fs with podman: {}'.format(
+                    result['stderr']))
+
+    try:
+        if os.path.isdir(source):
+            shutil.copytree(source, destination)
+        else:
+            shutil.copy(source, destination)
```
vrothberg: Currently, we ignore the returned mount point from `podman mount`. I guess we need to copy to `mountpoint/destination`.

flavio:
Velum enablement done via https://github.com/kubic-project/velum/pull/478

salt/_modules/caasp_cri.py
flavio: Good point, I'll tackle that.

salt/_modules/caasp_cri.py
flavio: Grains are about collecting facts from the nodes (eg: arch, cpu,...). In this case I think it's fine to have the code structured in that way. But maybe I misunderstood your comment...

flavio:
mhhh, looks like some containers are scheduled on the master node

```
kubectl get pods --all-namespaces -o wide                                                                                                                                                                               1 ↵
NAMESPACE                          NAME                                READY     STATUS              RESTARTS   AGE       IP           NODE
e2e-tests-pod-network-test-v6ckn   netserver-0                         0/1       Running             0          20s       172.16.2.4   caasp-worker-0
e2e-tests-pod-network-test-v6ckn   netserver-1                         0/1       ContainerCreating   0          20s       <none>       caasp-worker-1
e2e-tests-pod-network-test-v6ckn   netserver-2                         0/1       ContainerCreating   0          20s       <none>       caasp-worker-2
kube-system                        dex-7b45557ff8-chtsb                1/1       Running             4          8m        172.16.4.3   caasp-worker-1
kube-system                        dex-7b45557ff8-qvlqh                1/1       Running             4          8m        172.16.0.2   caasp-master-0
kube-system                        dex-7b45557ff8-wf5mz                1/1       Running             3          8m        172.16.2.3   caasp-worker-0
kube-system                        haproxy-caasp-master-0              1/1       Running             0          10m       172.30.2.0   caasp-master-0
kube-system                        haproxy-caasp-worker-0              1/1       Running             0          9m        172.30.3.0   caasp-worker-0
kube-system                        haproxy-caasp-worker-1              1/1       Running             0          10m       172.30.3.1   caasp-worker-1
kube-system                        haproxy-caasp-worker-2              1/1       Running             0          9m        172.30.3.2   caasp-worker-2
kube-system                        kube-dns-7488679ff9-cd48p           3/3       Running             0          8m        172.16.4.2   caasp-worker-1
kube-system                        kube-dns-7488679ff9-nlw4x           3/3       Running             0          8m        172.16.2.2   caasp-worker-0
kube-system                        kube-dns-7488679ff9-p9g6j           3/3       Running             0          8m        172.16.6.2   caasp-worker-2
kube-system                        kube-flannel-94f9s                  1/1       Running             0          9m        172.30.3.0   caasp-worker-0
kube-system                        kube-flannel-j2p7t                  1/1       Running             0          9m        172.30.3.1   caasp-worker-1
kube-system                        kube-flannel-qrbbt                  1/1       Running             0          9m        172.30.3.2   caasp-worker-2
kube-system                        kube-flannel-vncbb                  1/1       Running             0          9m        172.30.2.0   caasp-master-0
sonobuoy                           sonobuoy                            1/1       Running             0          3m        172.16.6.3   caasp-worker-2
sonobuoy                           sonobuoy-e2e-job-e8aa23a5cda74fd6   0/2       ContainerCreating   0          2m        <none>       caasp-master-0
```

salt/_modules/caasp_cri.py
vrothberg: Oups, sorry. I am not familiar with Salt and interpreted grains to be some kind of local key-value store.

flavio:
I've to keep an eye open on the pods scheduled on the master, in theory that shouldn't happen:
```
Taints:             node-role.kubernetes.io/master:NoSchedule
```

pillar/params.sls
flavio: I will leave the code as it is, to avoid clutter inside of the salt states.

flavio:
While looking at the remaining `TODOs` I found our codebase has an issue that's completely unrelated with crio. I've created a dedicated commit https://github.com/kubic-project/salt/pull/462/commits/77807a9a5135982be5670a3fcea151349f63a97d for that. Are you fine including that into this PR or should I create a new one?

ping @kiall who wrote the code affected by https://github.com/kubic-project/salt/pull/462/commits/77807a9a5135982be5670a3fcea151349f63a97d

BTW: I'll file a PR against salt upstream to fix the detection of containers started by kubernetes and by podman.

flavio:
**RFC:** on a node using cri-o we don't need to have the docker daemon running neither its images loaded.

I saw we have image removal (and some other cleanups) are done by [remove-pre-reboot.sls](https://github.com/kubic-project/salt/blob/master/salt/docker/remove-pre-reboot.sls) while disabling daemon is done by [docker/stop.sls](https://github.com/kubic-project/salt/blob/master/salt/docker/stop.sls).

How can I reuse these states? The easiest approach would be to extend the orchestrations states to include them. I could match the nodes using crio that are not the admin and the ca nodes via `not P@roles:(admin|ca) and J@cri:name:(crio)`. However that feels dirty to me, I would prefer to have them reused from the `crio.sls` file.

salt/crio/crio.conf.jinja
```diff
@@ -0,0 +1,160 @@
...
+# The "crio" table contains all of the server options.
+[crio]
+
+# root is a path to the "root directory". CRIO stores all of its data,
+# including container images, in this directory.
+root = "{{pillar['crio']['dirs']['root']}}"
+
+# run is a path to the "run directory". CRIO stores all of its state
+# in this directory.
+runroot = "{{pillar['crio']['dirs']['runroot']}}"
```
ereslibre: Minor: here as well as on `root`, we usually leave blank spaces between the double braces.

salt/etcd/create-or-update-etcd-pillar.jinja
```diff
@@ -0,0 +1,81 @@
...
+if velum_dashboard_container_id is None:
+    print('Error, cannot find the velum-dashboard running inside of the default namespace')
+    exit(1)
+
+rails_cmd = ['entrypoint.sh',
+             'bundle',
+             'exec',
+             'rails',
+             'runner',
+             "'ActiveRecord::Base.logger=nil; pillar = Pillar.find_or_create_by(pillar: \"api:etcd_version\"); pillar.value = \"etcd3\"; pillar.save'"]
```
ereslibre: I think we can avoid escaping `"` here:

`'ActiveRecord::Base.logger=nil; pillar = Pillar.find_or_create_by(pillar: "api:etcd_version"); pillar.value = "etcd3"; pillar.save'`

salt/transactional-update/init.sls
```diff
@@ -1,5 +1,5 @@
 {# In devenv, the transactional-update service does not exist on admin #}
```
ereslibre: I think this comment can go away too.

salt/motd/motd.jinja
```diff
@@ -1,7 +1,8 @@
-Welcome!
+Welcome to SUSE CaaS Platform!
 
 Machine ID: {{ salt['grains.get']('machine_id', 'ERROR') }}
 Hostname: {{ salt['grains.get']('nodename', 'ERROR') }}
+Container Runtime: {{ salt.caasp_cri.cri_name() }}
```
ereslibre: Maybe we can add yet another helper `cri_human_name` or something similar so we can output on the motd something like: "Docker Open Source Container Engine".

salt/etcd/create-or-update-etcd-pillar.jinja
```diff
@@ -0,0 +1,81 @@
...
+    with the CRI
+    '''
+    # check the crio socket first, there might be cases when docker and crio
+    # are running at the same time
+    if CRI == 'docker':
+        return '/var/run/dockershim.sock'
+    elif CRI == 'crio':
+        return '/var/run/crio/crio.sock'
+    else:
+        raise RuntimeError('Cannot find a suitable CRI socket')
```
inercia: This could be simplified:

```Python
def _cri_runtime_endpoint():
    return {{ salt.caasp_cri.cri_runtime_endpoint }}
```

and, as this would be a constant,

```Python
_CRI_RUNTIME_ENDPOINT = {{ salt.caasp_cri.cri_runtime_endpoint }}
```

and you could remove the previous `CRI` variable...

salt/haproxy/init.sls
```diff
@@ -1,6 +1,8 @@
 {% if "admin" not in salt['grains.get']('roles', []) %}
 include:
+  - {{ salt.caasp_cri.cri_salt_state_name() }}
```
inercia: I don't think we really need this dependency here and, if there is some dependency, it could be reversed: make X include `haproxy` and then add the dependency in X, with `require_in: haproxy`.

salt/kubelet/kubelet.jinja
```diff
@@ -31,6 +31,11 @@ KUBELET_ARGS="\
   {% endif -%}
 {% endif -%}
     {{ pillar['components']['kubelet']['args'] }} \
+{% if salt.caasp_cri.cri_name() == 'crio' -%}
+    --container-runtime=remote \
+    --container-runtime-endpoint=/var/run/crio/crio.sock \
```
inercia: There are many places where we are referencing this socket. Maybe we should add it to the pillar...

salt/motd/motd.jinja
inercia: I think a pillar value would be better, so we could do `{{ pillar['cri'][salt.caasp_cri.cri_name()]['description'] }}`

salt/orch/update.sls
```diff
@@ -229,7 +229,7 @@ etcd-setup:
       - container-feeder.stop
       - kubelet.stop
       - kube-proxy.stop
-      - docker.stop
+      - {{ salt.ccaasp_cri.cri_service_name() + '.stop' }}
```
inercia: Typo here.

salt/orch/update.sls
```diff
@@ -130,7 +130,7 @@ etcd-setup:
       - kube-apiserver.stop
       - kube-controller-manager.stop
       - kube-scheduler.stop
-      - docker.stop
+      - {{ salt.caasp_cri.cri_service_name() + '.stop' }}
```
inercia: All these dynamic calculations of SLS names do not look right to me. Maybe we should move all the different CRI drivers under the same hierarchy, something similar to what we have for the `addons`:

```Shell

    /cri
        init.sls
        stop.sls
        /docker
            init.sls
            stop.sls
        /crio
            init.sls
            stop.sls
```

So we could just invoke `cri.stop` and `cri.stop` would invoke the right `stop` in `docker` or `crio`...

vrothberg:
We also need to configure `/etc/crictl.yaml`, which currently defaults to `/var/run/crio/crio.sock`. For Docker, we need to use `/var/run/dockershim.sock`.

vrothberg:
If we want to use `crictl exec` we need to pass the `--sync` flag to execute the command synchronously. Otherwise, it will fail using `dockershim.sock` as a run-time endpoint. I opened an issue upstream to find out why this is happening.

salt/etcd/create-or-update-etcd-pillar.jinja
flavio: We have to escape that, otherwise python will choke.

salt/_modules/caasp_cri.py
```diff
@@ -0,0 +1,263 @@
...
+                                     output_loglevel='trace',
+                                     python_shell=False)
+
+    if result['retcode'] != 0:
+        raise CommandExecutionError(
+                'Could not invoke crictl',
+                info={'errors': [result['stderr']]}
+            )
+
+    ps_data = json.loads(result['stdout'])
```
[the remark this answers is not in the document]
flavio: The function returns the `ID` of the container of `None` when no running container is found. I decided to not handle the exception to let everything blow up. I just didn't want to swallow the exception into a log file.

Probably the best approach would be to handle the exception at a higher level (like the salt state actually using it) to return a better error message to the end user... Thoughts?

salt/crio/init.sls
```diff
@@ -0,0 +1,32 @@
...
+/etc/crio/crio.conf:
+  file.managed:
+    - source: salt://crio/crio.conf.jinja
+    - template: jinja
+    - require_in:
+      - kubelet
+  service.running:
+    - name: crio
+    - reload: True
+    - onchanges:
```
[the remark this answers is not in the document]
flavio: Right now there's a know bug that causes cri-o to lose track of the running containers when restarted. This is being worked on upstream.

Is the service going to be restarted all the times if we switch to a `watches`?

salt/_modules/caasp_cri.py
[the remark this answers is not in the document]
inercia: I agree with you about sending an exception to the upper layers, but I think it should be some exception specific to CRI (ie, `CRIRuntimeException`) instead of a JSON decoding error that means nothing to a user of this module. Anyway, it is a small detail...

salt/crio/init.sls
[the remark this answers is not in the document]
inercia: A `watches` will restart (or reload, if supported) the service if the _watched_ items are changed.

flavio:
I applied all the proposed changes except for the reorganization of the docker/crio folders. I'll take care of that tomorrow.

In the meantime feel free to take a look at the changes I introduced.

salt/cri/remove-pre-reboot.sls
```diff
@@ -0,0 +1,2 @@
+include:
+  - {{ pillar['cri'][salt.caasp_cri.cri_name()]['service'] }}.remove-pre-reboot
```
inercia: Nice hack.

flavio:
I squashed the final cleanup commits into a single one. I prefer to keep some of the intermediate commits around to leave some history (like 1bae9eb6b1b3e2847a0e1837274e5e2b95115b0f - that can be reverted later one day)

salt/haproxy/init.sls
```diff
@@ -62,8 +62,10 @@ haproxy-restart:
     - timeout: 60
     - onchanges:
       - file: /etc/caasp/haproxy/haproxy.cfg
+{% if not salt.caasp_cri.needs_docker() %}
     - require:
       - service: container-feeder
```
inercia: Why is the `container-feeder` service not required for docker?

flavio: Our images have docker enabled by default. On the 1st boot container-feeder will automatically import all the pre-built images into docker. That doesn't happen with cri-o. Hence we have to ensure the container-feeder sls file has been applied when docker is not being used.

PR #463, opened 2018-04-03
Add force removal orchestration

ereslibre:
This orchestration will try to unregister a node on a best-effort
basis, and is considered to always succeed.

feature#force-node-removal

### DIFF
--- a/salt/_modules/caasp_etcd.py
+++ b/salt/_modules/caasp_etcd.py
@@ -197,17 +197,25 @@ def get_etcdctl_args_str(**kwargs):
     return " ".join(get_etcdctl_args(**kwargs))
 
 
-def get_member_id():
+def get_member_id(nodename=None):
     '''
     Return the member ID (different from the node ID) for
-    this etcd member of the cluster.
+    a etcd member of the cluster.
+
+    Arguments:
+
+    * `nodename`: (optional) the nodename for the member we
+                  want the ID for. if no name is provided (or empty),
+                  the local node will be used.
     '''
     command = ["etcdctl"] + get_etcdctl_args() + ["member", "list"]
+
+    target_nodename = nodename or __salt__['caasp_net.get_nodename']()
 
     debug("getting etcd member ID with: %s", command)
     try:
         this_nodename = __salt__['caasp_net.get_nodename']()
-        this_url = 'https://{}:{}'.format(this_nodename, ETCD_CLIENT_PORT)
+        this_url = 'https://{}:{}'.format(target_nodename, ETCD_CLIENT_PORT)
         members_output = subprocess.check_output(command)
         for member_line in members_output.splitlines():
             if this_url in member_line:
--- a/salt/_states/caasp_etcd.py
+++ b/salt/_states/caasp_etcd.py
@@ -61,18 +61,25 @@ def member_add(name, **kwargs):
     # or, otherwise, etcd will refuse to join... (facepalm)
 
 
-def member_remove(name, **kwargs):
+def member_remove(name, nodename=None, **kwargs):
     '''
-    Remove this node from the etcd cluster
+    Remove a member from the etcd cluster
+
+    Arguments:
+
+    * `nodename`: (optional) the nodename for the member we
+                  want the ID for. if no name is provided (or empty),
+                  the local node will be used.
     '''
-    this_member_id = __salt__['caasp_etcd.get_member_id']()
-    if not this_member_id:
+    target_member_id = __salt__['caasp_etcd.get_member_id'](nodename=nodename)
+    if not target_member_id:
         return {
             'name': "member_remove.{0}".format(name),
             'result': False,
-            'comment': "Could not obtain member id."
+            'comment': "Could not obtain member id.",
+            'changes': {}
         }
 
-    name = 'member remove {}'.format(this_member_id)
-    log.debug('CaaS: removing etcd member %s', this_member_id)
+    name = 'member remove {}'.format(target_member_id)
+    log.debug('CaaS: removing etcd member %s', target_member_id)
     return etcdctl(name=name, **kwargs)
--- a/salt/orch/force-removal.sls
+++ b/salt/orch/force-removal.sls
@@ -1,0 +1,57 @@
+# must provide the node (id) to be removed in the 'target' pillar
+{%- set target = salt['pillar.get']('target') %}
+{%- set target_nodename = salt.saltutil.runner('mine.get', tgt=target, fun='nodename')[target] %}
+
+{%- set super_master = salt.saltutil.runner('manage.up', tgt='G@roles:kube-master and not ' + target, expr_form='compound')|first %}
+{%- set etcd_members = salt.saltutil.runner('manage.up', tgt='roles:etcd', expr_form='grain') %}
+
+set-cluster-wide-removal-grain:
+  salt.function:
+    - tgt: '*'
+    - name: grains.setval
+    - arg:
+      - force_removal_in_progress
+      - true
+
+sync-all:
+  salt.function:
+    - tgt: '*'
+    - names:
+      - saltutil.refresh_pillar
+      - saltutil.refresh_grains
+      - mine.update
+      - saltutil.sync_all
+
+{%- if target in etcd_members %}
+unregister-{{ target }}-etcd:
+  salt.state:
+    - tgt: {{ super_master }}
+    - sls:
+        - etcd.remove-pre-stop-services
+    - fail_minions: {{ super_master }}
+    - pillar:
+        nodename: {{ target_nodename }}
+{%- endif %}
+
+unregister-{{ target }}-kubelet:
+  salt.function:
+    - tgt: {{ super_master }}
+    - name: cmd.run
+    - fail_minions: {{ super_master }}
+    - arg:
+        - kubectl --kubeconfig={{ pillar['paths']['kubeconfig'] }} delete node {{ target_nodename }}
+
+remove-cluster-wide-removal-grain:
+  salt.function:
+    - tgt: '*'
+    - name: grains.delval
+    - arg:
+      - force_removal_in_progress
+    - kwarg:
+        destructive: True
+
+remove-target-salt-key:
+  salt.wheel:
+    - name: key.reject
+    - include_accepted: True
+    - match: {{ target }}
--- a/salt/etcd/remove-pre-stop-services.sls
+++ b/salt/etcd/remove-pre-stop-services.sls
@@ -5,6 +5,7 @@
 # so it is not considered a node suffering some
 # transient failure...
 etcd-remove-member:
-  caasp_etcd.member_remove
+  caasp_etcd.member_remove:
+    - nodename: {{ salt['pillar.get']('nodename', grains['nodename']) }}
   # NOTE: we are not requiring /etc/hosts or the certificates
   #       because we are assuming this node was on high state
--- a/salt/orch/update-etc-hosts.sls
+++ b/salt/orch/update-etc-hosts.sls
@@ -2,7 +2,8 @@
                              'G@bootstrap_complete:true and ' +
                              'not G@bootstrap_in_progress:true and ' +
                              'not G@update_in_progress:true and ' +
-                             'not G@removal_in_progress:true' %}
+                             'not G@removal_in_progress:true and ' +
+                             'not G@force_removal_in_progress:true' %}
 
 {%- if salt.saltutil.runner('mine.get', tgt=updates_all_target, fun='nodename', tgt_type='compound')|length > 0 %}
 update_pillar:


inercia:
I think this is not really necessary as the _force-removal_ is going to be implemented as a salt key deletion in the Salt master (as we assume that the node is unresponsive and we cannot really do anything there).

ereslibre:
@inercia While I agree in general terms, there are still things we can try, like trying a `kubectl delete node` on **other machines** so the node disappears from the kubectl listings. This orchestration allows every step to fail, and the orchestration itself should succeed.

I'll check how feasible this is, because only removing the salt key won't do anything with the cluster itself. Think about it as a last resort for cleaning and using other healthy machines to try to remove the node from existence.

inercia:
We discussed that, once the user presses the "force remove" for a node, we should warn the user that that node should be reinstalled and/or removed from the cluster, so it will not disrupt the cluster.

And, if the only thing we want to do is to do a `kubectl delete node`, we don't need a new orchestration: it can be done as part of the regular `removal.sls` orchestration...

ereslibre:
> And, if the only thing we want to do is to do a kubectl delete node, we don't need a new orchestration: it can be done as part of the regular removal.sls orchestration...

The `removal.sls` orchestration is an orchestration that tries to do things as cleanly as possible, adding replacements if necessary and all sort of logic of that kind. This orchestration should probably only be called if the `removal.sls` orchestration failed. If the user force removes a node we should do some other tasks apart from just revoking the key, but running the whole `removal.sls` orchestration might be too much.

inercia:
I'm already working on resisting unresponsive nodes on the node removal orchestration in #464 and adding the same resiliance for the `target` is trivial. I will add some code there instead of more orchestrations...

ereslibre:
> I'm already working on resisting unresponsive nodes on the node removal orchestration in #464 and adding the same resiliance for the target is trivial. I will add some code there instead of more orchestrations...

Adding that resiliance on the removal orchestration is nice, the main problem is that we can try things on the force_removal orchestration that really don't belong to the removal orchestration. For example: if a node is healthy and fully working and we just call `orch.removal` it makes sense for it to unregister itself. However, if we are force removing a node, it makes sense to make others try that unregistration, because we don't know in what state the target is, and we cannot assume anything about it.

Maintaining everything in a single removal orchestration will be hard to maintain and error prone when we use the "force removal" feature of the `removal` orchestration. This is why I introduced the `force_removal` orchestration, that will do *minor* tasks in order to clean the trace of the force removed node.

ereslibre:
This approach clashes with the `forced` argument addition present in https://github.com/kubic-project/salt/pull/464/. We need to pick our poison: dedicated force removal orchestration or argument `forced` on the regular `orch.removal` orchestration.

ereslibre:
The rationale behind this orchestration is that it's a best effort orchestration, and doesn't assume anything about the target nor the cluster.

* All steps can fail with `fail_minions` argument
* All sls calls are done individually (best effort, we don't care if some fail, we do our best)
* Since we cannot assume the target will be healthy nor alive, try to deregister it from Kubernetes using other nodes. Also we don't assume any living node is healthy (as in, the salt-minion being responsive doesn't mean that `kubectl` will succeed there)
* No requirements on steps. We always end up removing the target salt key

The main reasoning for a different orchestration on my view is that this one assumes nothing, and we do our best on each step to try to clean up. If that doesn't work, we don't care. This orchestration is a last resort orchestration, for use when the regular `orch.removal` has failed for whatever reason.

salt/orch/force_removal.sls
```diff
@@ -0,0 +1,72 @@
...
+{%- for service, sls_list in services_to_stop.items() %}
+  {%- for sls in sls_list %}
+stop-{{ service }}-{{ sls }}-stage:
+  salt.state:
+    - tgt: {{ target }}
+    - fail_minions: {{ target }}
+    - sls:
+        - {{ service }}.{{ sls }}
+  {%- endfor %}
+{%- endfor %}
```
inercia: I find this loop very confusing. This generates four only `salt.state`, so maybe it would be better to unroll the loop...

salt/orch/force_removal.sls
```diff
@@ -0,0 +1,72 @@
...
+    - name: cmd.run
+    - fail_minions: {{ node }}
+    - arg:
+        - kubectl --kubeconfig={{ pillar['paths']['kubeconfig'] }} delete node {{ target_nodename }}
+  {%- if previous_node %}
+    - onfail:
+        - try-to-unregister-{{ target }}-from-{{ previous_node }}
+  {%- endif %}
+  {%- set previous_node = node %}
+{%- endfor %}
```
inercia: Why don't you just run the `kubectl delete node` in a master that is up and running? Get all the masters that are up and running (with the `mine.get`, targeting `G@roles:kube-master and not L@node_down1,...`), and then running this salt function on `tgt: {{ responsive_masters|first }}`

salt/orch/force_removal.sls
```diff
@@ -0,0 +1,72 @@
...
+
```
inercia: I think we should also remove the node from the etcd cluster...

salt/orch/force_removal.sls
```diff
@@ -0,0 +1,72 @@
+# This is a best effort orchestration. Any operation run on the minions by this orchestration
+# could fail, as we don't know the current state of the cluster at this point. The idea is to
+# try to clean as much as possible where possible, but in the end the only thing we can really
+# ensure is that the salt-key of the target is rejected.
+
+# must provide the node (id) to be removed in the 'target' pillar
+{%- set target = salt['pillar.get']('target') %}
+{%- set target_nodename = salt.saltutil.runner('mine.get', tgt=target, fun='nodename')[target] %}
```
inercia: You can use [get_nodename()](https://github.com/kubic-project/salt/blob/master/salt/_modules/caasp_net.py#L64), so it would look like:

```
{%- set target_nodename = salt.caasp_net.get_nodename(host=target) %}
```

salt/orch/force_removal.sls
```diff
@@ -0,0 +1,72 @@
...
+# services that will be called when trying to unregister as clean as possible
+{%- set services_to_stop = { "kubelet": ["stop", "remove-pre-reboot"], "etcd": ["remove-pre-stop-services", "remove-pre-reboot"] } %}
+# nodes that are responding
+{%- set up_nodes = salt.saltutil.runner('manage.up', tgt='P@roles:kube-* and not ' + target, expr_form='compound') %}
```
inercia: It would be better to get the list of nodes that are down (it should be short) instead of the list of nodes that are up (it can be very long). Not necessary to get this list anyway...

salt/orch/force_removal.sls
inercia: I think you need to skip this whole block if `target in nodes_down`. Otherwise, the `salt.state` will fail (even with `fail_minions`) as the Salt minion is not responding...

ereslibre: It's easier to extend without repeating yourself.

salt/orch/force_removal.sls
ereslibre: Because we can't assume that a master that has salt-minion responding will succeed with running `kubectl delete node`.

salt/orch/force_removal.sls
ereslibre: We are, we are running: https://github.com/kubic-project/salt/blob/master/salt/etcd/remove-pre-stop-services.sls

salt/orch/force_removal.sls
ereslibre: Thanks, will fix!

salt/orch/force_removal.sls
inercia: If no masters have a Salt minion responding you better reinstall your cluster... :-)

Anyway, you can get any responsive machine in the cluster where `kubectl` can be run (a master or a minion), and just run it there instead of looping.

salt/orch/force_removal.sls
ereslibre: I will have a look at how I can improve this.

salt/orch/force_removal.sls
inercia: That can fail, so you should do it in some other machine. Any machine where `etcdctl` can be run should do the job...

ereslibre: Will fix

salt/orch/force_removal.sls
ereslibre: There's a long way between "no masters have a salt minion responding" and "I will just try with the first I got". Besides, my whole point is that the master you choose might have salt-minion responding, but `kubectl delete node` could fail there.

Please note that in this case as long as one succeeds running `kubectl delete node` the rest afterwards won't be executed because of the `onfail` requisite.

inercia: We must be prepared for the target not to respond, but we don't have to be prepared for "maybe all machines in the cluster will fail to run any command" because, if that is the case, you better reinstall the cluster...

ereslibre: As stated several times previously, this is a last resort orchestration. You could have two failing machines, you can only remove one by one, and it could happen that `masters|first` is the other machine failed. As a last resort orchestration I think this is the right way to go, go best effort everywhere you can.

ereslibre: Yes, I could have a deployment with say, 5 masters. I don't follow your reasoning here. 5 masters, 40 workers. 1 unresponsive master and 1 unresponsive worker. I click to force remove the worker, and `masters|first` is the unresponsive master...

ereslibre: By the way, you **can** add new masters.

inercia: It is not `masters|first` but `responsive_masters|first`, and

```SaltStack
# get nodes_down with manage.down()
{% responsive_masters = salt['mine.get'](
    'G@roles:kube-master and not L@' + nodes_down|join(','),
    'nodename', 'compound').keys() %}
```

so those masters will be responsive (at least for Salt).

ereslibre: I have stated several times before that a minion responding for salt does not mean anything more than its `salt-minion` is working and talking to the master. You don't know anything about that machine and if `kubectl delete node` will work or not.

inercia: ... and you needt least a kubernetes master with a salt minion that is responsive _and_ a working `kubectl` or otherwise your cluster is utterly broken and you will not be able to do anything. All our orchestrations depend on that, so if you remove an unresponsive node and you only leave a unresponsive master, the cluster is unsable.

flavio: Let's try to keep this code as simple as possible. Let's go to the first responsive master and issue the `kubectl delete` statement. If that proves to be to fragile we can always bring this code back.

salt/orch/force_removal.sls
ereslibre: This function doesn't work on the master because it got removed the case for `if __opts__['__role'] == 'master'` so it cannot be used when rendering as it is now.

salt/_modules/caasp_etcd.py
```diff
@@ -217,4 +226,36 @@ def get_member_id():
...
+    return __salt__['cmd.run_all'](cmd, **kwargs)
+
+
+def healthy():
+    return etcdctl(name='cluster-health')['retcode'] == 0
+
+
+def member_list():
+    return etcdctl(name='member list')['stdout']
+
```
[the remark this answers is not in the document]
ereslibre: `member_list` was more to debug easier. I can remove them though. In general, I'd like to start following the salt pattern of writing stuff in `modules` and `states` just call to those modules adding the result hash.

salt/_states/caasp_etcd.py
```diff
@@ -8,71 +8,71 @@
...
-    return etcdctl(name='cluster-health', **kwargs)
+    ret = {
+        'name': name,
+        'changes': {}
+    }
+    if __salt__['caasp_etcd.healthy']():
+        ret.update({
+            'result': True,
+            'comment': 'etcd cluster is healthy',
+        })
```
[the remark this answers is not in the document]
ereslibre: The latter is not very readable, I will update to the first one.

salt/orch/force-removal.sls
```diff
@@ -0,0 +1,130 @@
...
+# must provide the node (id) to be removed in the 'target' pillar
+{%- set target = salt['pillar.get']('target') %}
+{%- set target_nodename = salt.saltutil.runner('mine.get', tgt=target, fun='nodename')[target] %}
+# services that will be called when trying to unregister as clean as possible
+{%- set services_to_stop = { "kubelet":   ["stop", "remove-pre-reboot"],
+                             "haproxy":   ["remove-pre-reboot"],
+                             "etc-hosts": ["remove-pre-reboot"] } %}
+# nodes that are responding
+{%- set all_up_nodes = salt.saltutil.runner('manage.up', tgt='P@roles:(etcd|(kube-*))', expr_form='compound') %}
+{%- set up_nodes_except_target = salt.saltutil.runner('manage.up', tgt='P@roles:(etcd|(kube-*)) and not ' + target, expr_form='compound') %}
```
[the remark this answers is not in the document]
ereslibre: I'll check if I can just substract `target` to `all_up_nodes`, but this is jinja.

salt/orch/force-removal.sls
```diff
@@ -0,0 +1,130 @@
...
+{%- set etcd_members = salt.saltutil.runner('mine.get', tgt='G@roles:etcd', fun='network.interfaces', tgt_type='compound').keys() %}
+
+# Ensure we mark all nodes with the "as node is being removed" grain.
+# This will ensure the update-etc-hosts orchestration is not run.
+set-cluster-wide-removal-grain:
+  salt.function:
+    - tgt: {{ all_up_nodes|join(',') }}
+    - tgt_type: list
+    - fail_minions: {{ all_up_nodes|join(',') }}
```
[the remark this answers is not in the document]
ereslibre: Because any step on this force removal orchestration can fail. This just adds more resiliency to this orchestration, I don't think it's an issue.

salt/orch/force-removal.sls
```diff
@@ -0,0 +1,130 @@
...
+    - name: grains.setval
+    - arg:
+      - force_removal_in_progress
+      - true
+
+# Try to sync some information before going on. We might use modules that have not been
+# discovered yet on the minions.
+sync-all:
+  salt.function:
+    - tgt: {{ all_up_nodes|join(',') }}
```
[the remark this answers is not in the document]
ereslibre: Yes... in this specific orchestration I prefer to be more explicit to who we call and what.

salt/orch/force-removal.sls
```diff
@@ -0,0 +1,130 @@
...
+      - mine.update
+      - saltutil.sync_all
+
+{%- if target in all_up_nodes %}
+# Try to unregister the machine from cluster services and perform minimal housekeeping. It's
+# important to keep every sls in different toplevel stages, so each one will be tried even
+# if previous fails for whatever reason.
+{%- for service, sls_list in services_to_stop.items() %}
+  {%- for sls in sls_list %}
+stop-{{ service }}-{{ sls }}-stage:
```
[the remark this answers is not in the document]
ereslibre: If it's down we won't execute this part. If it's not, maybe the `orch.removal` orchestration failed and we can try to clean a little before giving up. Do you really think this adds complexity to completely reject this?

salt/orch/force-removal.sls
```diff
@@ -0,0 +1,130 @@
...
+    - arg:
+      - systemctl poweroff
+    - kwarg:
+        bg: True
+{%- endif %}
+
+{%- if target in etcd_members %}
+# Try to remove the node from the etcd cluster from any other working machine.
+{%- set previous_node = none %}
+{%- for node in up_nodes_except_target %}
```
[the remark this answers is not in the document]
ereslibre: If the `super-master` salt-minion is up but this fails there, we just give up? I think we can do better, try from all the machines that can run this. As long as one works, the rest will be omitted (because the `onfail`).

salt/orch/force-removal.sls
```diff
@@ -0,0 +1,130 @@
...
+        - try-to-unregister-etcd-{{ target }}-from-{{ previous_node }}
+  {%- endif %}
+  {%- set previous_node = node %}
+{%- endfor %}
+{%- endif %}
+
+# Try to remove the node from the kubernetes cluster from any other working machine.
+{%- set previous_node = none %}
+{%- for node in up_nodes_except_target %}
+try-to-unregister-kubelet-{{ target }}-from-{{ node }}:
```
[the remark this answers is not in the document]
ereslibre: Same here, I truly believe this adds value.

salt/orch/force-removal.sls
```diff
@@ -0,0 +1,130 @@
...
+remove-cluster-wide-removal-grain:
+  salt.function:
+    - tgt: {{ all_up_nodes|join(',') }}
+    - tgt_type: list
+    - fail_minions: {{ all_up_nodes|join(',') }}
+    - name: grains.delval
+    - arg:
+      - force_removal_in_progress
+    - kwarg:
+        destructive: True
```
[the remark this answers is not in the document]
ereslibre: According to https://docs.saltstack.com/en/latest/topics/orchestrate/orchestrate_runner.html#more-complex-orchestration: `Additionally, the states/functions will be executed in the order in which they are defined, unless prevented from doing so by any Requisites and Other Global State Arguments, as is the default in SLS files since 0.17.0`

salt/_modules/caasp_etcd.py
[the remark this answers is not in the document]
ereslibre: This is very simple actually, we need to call to `member_remove` on node `X` to remove node `Y` from etcd. It needs to be a function. Having it as a state passing a pillar as an argument is plain wrong and incorrect. If I move `member_remove`, it makes sense to also move `member_add`. If you are complaining about `member_list`, I'll remove it. TBH, I don't see the big deal.

salt/_states/caasp_etcd.py
```diff
@@ -8,71 +8,71 @@
...
-
-# default etcd peer port
-ETCD_PEER_PORT = 2380
-
-
-def etcdctl(name, retry={}, **kwargs):
-    '''
-    Run an etcdctl command
-    '''
-    retry_ = {'attempts': DEFAULT_ATTEMPTS,
```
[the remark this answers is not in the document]
ereslibre: The main problem with this is that we cannot call to states from within functions, and `caasp_cmd.run` is a state only (as many things we did that we should fix). So we need a further refactor to make `caasp_cmd.run` be a function that returns apart from the original `cmd.run_all` results the number of retries it did out of the total, so we could use that from other functions.

salt/orch/force-removal.sls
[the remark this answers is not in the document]
ereslibre: How so? Do you have facts that back up your theory?

salt/orch/force-removal.sls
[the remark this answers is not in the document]
ereslibre: Erm, no... read the state again. It will unroll to exactly 4 states on the orchestration (with 4 and with 4000 nodes).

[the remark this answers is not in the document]
inercia: Right, I misread the `if`. Anyway I think it is completely unnecessary to stop services in a machine we are going to nuke.

inercia:
LGTM. Thanks.

ereslibre:
[Node Remove: Force removal of dead/gone node](https://trello.com/c/u1J8JknO/628-node-remove-force-removal-of-dead-gone-node)

PR #464, opened 2018-04-03
Node removal: resist unresponsive nodes

inercia:
# Description

Try to resist any unresponsive nodes when removing a node `target`, so users should be able to remove the `target` even when some other node(s) are down. 

This implies that

* the `replacement` will not be chosen from the list of unresponsive nodes.
* the highstate on _affected nodes_ will exclude unresponsive nodes.

# Main changes

* some code for discovering unresponsive nodes, using that list of nodes as an `excluded` argument introduced in #442 in many functions that deal with nodes replacements.
* we do not try to perform the `target` removal from the kubernetes and etcd clusters _in_ the `target`. Instead of that, we perform the removal in some _supermaster_, so the `target` can be down and the removal will be performed anyway. 

# Testing

1. Boostrap the cluster by running the regular orchestration.

2. Stop the `salt-minion` service in some node `NODE_DOWN`, and get the `machine-id` of that node:

```
$ ssh NODE_DOWN
# systemctl stop salt-minion
# cat /etc/machine-id
4eace3637a7d4c3b93781b545b13032c
```

3. Check we can remove some other node `OTHER_NODE` and the orchestration doesn't fail because `NODE_DOWN` is not responding, by running in the Salt container:

```
$ ssh OTHER_NODE cat /etc/machine-id
a0848146a8854c519ce698d28901e824
# salt run -l debug --force-color \
  state.orchestrate "orch.removal" pillar='{"target":"a0848146a8854c519ce698d28901e824"}'
```

4. The orchestration should be able to remove the node from the cluster, `kubectl get nodes`  and `etcd member list` (if the node was part of the etcd cluster) should not show the node. The orchestration should perform a best-effort search for a replacement.

# References

See https://trello.com/c/O7daOErL

feature#node_removal


ereslibre:
The bit about `forced` argument on this orchestration clashes with the new orchestration added in https://github.com/kubic-project/salt/pull/463. We need to pick our poison.

salt/cleanup/remove-post-orchestration.sls
```diff
@@ -0,0 +1,33 @@
...
+  - kubectl-config
+
+{%- set target = salt['pillar.get']('target') %}
+{%- set nodename = salt.caasp_net.get_nodename(host=target) %}
+
+###############
+# k8s cluster
+###############
+
+{%- set k8s_nodes = salt['mine.get']('roles:(kube-master|kube-minion|etcd)', 'nodename', 'grain_pcre').keys() %}
```
ereslibre: Here we are usually explicitly saying `expr_form='grain_pcre'` (despite with a newer salt we'll have to change all occurences to `tgt_type`...)

salt/_states/caasp_etcd.py
```diff
@@ -27,7 +27,9 @@ def etcdctl(name, retry={}, **kwargs):
               'until': None}
     retry_.update(retry)
 
-    args = __salt__['caasp_etcd.get_etcdctl_args_str']()
+    skip_this = kwargs.pop('skip_this', False)
```
ereslibre: General question: why bury this kind of arguments inside `kwargs` inside of adding them as regular arguments with defaults?

salt/cleanup/remove-post-orchestration.sls
```diff
@@ -0,0 +1,33 @@
...
+{%- if target in k8s_nodes %}
+
+{%- from '_macros/kubectl.jinja' import kubectl with context %}
+
+{{ kubectl("remove-node",
+           "delete node " + nodename,
+           check_cmd="/bin/true") }}
```
ereslibre: `etcd` removal hasn't got this `check_cmd`. On one side I'd like to reuse this `sls` from the force removal orchestration, but I think in this regular `removal` orchestration, we should fail if this fails (or removing the etcd member), while on the force removal orchestration we shouldn't.

salt/orch/removal.sls
```diff
@@ -1,23 +1,52 @@
...
+{#- must provide the node (id) to be removed in the 'target' pillar #}
 {%- set target = salt['pillar.get']('target') %}
 
+{#- ... and we can provide an optional replacement node #}
+{%- set replacement = salt['pillar.get']('replacement', '') %}
+
+{#- we can skip the target (for example,e when it is unresponsive)  #}
+{#- in that case, we will not perform any cleanups in the target machine #}
+{#- (but it will be removed from the kubernetes and etcd clusters anyway) #}
+{%- set skip_target = salt.caasp_pillar.get('skip', False) %}
```
ereslibre: I would not add `skip` logic here... I think it's safe to assume the cluster is working fine, as a last resort there's the force removal orchestration. Nothing will set this flag on the pillar.

salt/cleanup/remove-post-orchestration.sls
inercia: etcd removal _must_ succeed: otherwise, the etcd cluster will be broken. If you ignore that you could leave a cluster that is unstable and unusable.

ereslibre: My main point being: why not remove `check_cmd` from here too?

salt/_states/caasp_etcd.py
inercia: Maybe not in this case in particular, but this kind of arguments are passed from function to function without being processed (and they are usually hidden/obscure arguments, so it is not very important)

salt/cleanup/remove-post-orchestration.sls
inercia: Yeah, I will change this just in case...

salt/orch/removal.sls
inercia: I would prefer to keep it. Even if nobody uses this in upper layers, it only adds 5 `if`s and brings some nice functionality.

salt/_states/caasp_etcd.py
ereslibre: I think where possible we should avoid this style (except when you need to forward unknown arguments to an inner function), this makes the code harder to read IMO (you need to read the body in order to find out what arguments it can take).

salt/cleanup/remove-post-orchestration.sls
inercia: Because removing a node from the API server is not so critical, so we don't want to abort the whole orchestration if this fails. In the worst case, the node will be kept in kubernetes but it will appear as not ready.

salt/orch/removal.sls
ereslibre: I overlooked `skip_this` in https://github.com/kubic-project/salt/pull/442. IMO if it's not used it should not be there, and we should remove it from the existing code that was merged today.

salt/cleanup/remove-post-orchestration.sls
ereslibre: IMO this doesn't look good from a product point of view. We should only succeed if all operations here did work as expected. Having leftovers is something we should avoid (I know it's minimal, but still, looks like a leftover from a customer point of view). If this fails IMO the orchestration should not succeed. If this orchestration does not succeed, the customer can trigger a force remove afterwards, that (as I was trying to explain in https://github.com/kubic-project/salt/pull/463/) will try this from all possible machines (`roles:kube=*`). I at least wouldn't like to see as a customer that the removal "went fine" (all green), and I can still see the removed node in `kubectl get nodes`.

inercia: We have been doing this for a long time (this code has been moved here from the `kubelet/init.sls` file), because we decided long ago it was a good idea to go ahead even if the `kubectl delete node` fails. And I still think it is the better solution, because users should not be forced to solve any problem in their clusters with a forced removal...

ereslibre: If you mean:

- [`kubelet/update-post-start-services.sls`](https://github.com/kubic-project/salt/blob/7a3c61f05ea09f5b938caa90d0b09d04a2f09662/salt/kubelet/update-post-start-services.sls): that's contained in the migration process, and it needs to be idempotent.
- [`kubelet/stop.sls`](https://github.com/kubic-project/salt/blob/master/salt/kubelet/stop.sls#L33-L39) (no `check_cmd = /bin/true`)

ereslibre: So leaving aside that we haven't been doing that for a long time; if something fails, we should not silently succeed. That's my 2 cents.

salt/_states/caasp_etcd.py
flavio: I agree with @ereslibre, but this is just my personal taste.

salt/_states/caasp_etcd.py
```diff
@@ -64,15 +66,23 @@ def member_add(name, **kwargs):
...
+
+    Arguments:
+
+    * `nodename`: (optional) the nodename for the member we
+                  want to remove. if no name is provided (or empty),
+                  the local node will be used.
     '''
-    this_member_id = __salt__['caasp_etcd.get_member_id']()
-    if not this_member_id:
+    nodename = kwargs.pop('nodename', '')
```
flavio: Same here, I would just add a `nodename` arg set to be `None` by default instead of relying on `kwargs`. But again, this is personal taste.

salt/cleanup/remove-post-orchestration.sls
flavio: I would prefer to not silently swallow the error. However I would like to confirm what I understood by reading the comments and the code.

The problem arises when we are removing an etcd node and something goes wrong. Am I right?

If the error on kubectl delete is ignored:

  * we have better chances to have a working etcd cluster, that's because the orchestration will find another node to assign the `etcd` role and the cluster will be healthy.
  * the user won't be notified that something went wrong. Even if we had the perfect UI/UX to show logs there's no way we can catch that because we simply ignored the error.
  * `kubectl get nodes` will still report the node. However kubernetes will realize the node is no longer reachable and will act accordingly (if that's a worker node no more workloads will be scheduled there).

If we don't swallow the error:

  * The orchestration won't assign another etcd node to the cluster. That means the etcd cluster might be close to breakage (eg: only 2 out of 3 nodes are running).
  * The node to be removed will still be shown into the UI.
  * The user is aware of the failure. One day it will also be easier for him to understand what went wrong, now he would have to look into the logs/open a ticket.
  * Can the user trigger another clean removal? What would happen then?
  * Is the user informed about the potential issue of the etcd cluster (eg: 2/3 nodes up)?
  * Is there a way for the user to grow the etcd cluster? Let's assume he leaves the failed to be removed node there and he adds a new worker/master node?

inercia: > The problem arises when we are removing an etcd node and something goes wrong. Am I right?

No, this is just for `kubectl delete node` failing. I think we shoud ignore it and continue with the removal orchestration (it will result in a node `NotReady` showing in Kubernetes) while @ereslibre thinks we should abort the orchestration.

salt/_states/caasp_etcd.py
inercia: This is a really subjective thing. In this case, most `**kwargs` go straigth to `caasp_cmd.run` (so we could use, for example, the `onlyif` argument when using this `caasp_etcd.etcdctl` state).

We must also extract some arguments used for for building the list of `etcdctl` arguments. `skip_if` is currently the only argument extracted. Adding it expllicitly in the arguments of this function would, in my view, give it more importance than what it really has, but as I said, this is very subjective.

However, as @flavio said, it should be documented in some way, so I will add some comment...

salt/_states/caasp_etcd.py
inercia: I agree.

salt/cleanup/remove-post-orchestration.sls
inercia: It seems we are using both `expr_form` and `tgt_type`

```
...
salt/orch/kubernetes.sls:{%- set etcd_members = salt.saltutil.runner('mine.get', tgt='G@roles:etcd', fun='network.interfaces', tgt_type='compound').keys() %}
salt/orch/update.sls:{%- set all_masters = salt.saltutil.runner('mine.get', tgt=is_master_tgt, fun='network.interfaces', tgt_type='compound').keys() %}
salt/reboot/init.sls:{%- set etcd_members = salt['mine.get']('G@roles:etcd', 'nodename', expr_form='compound').values() %}
...
```

I will use `tgt_type` as it seems to be fully supported..

salt/_states/caasp_etcd.py
inercia: Done.

salt/cleanup/remove-post-orchestration.sls
```diff
@@ -10,7 +10,7 @@ include:
 # k8s cluster
 ###############
 
-{%- set k8s_nodes = salt['mine.get']('roles:(kube-master|kube-minion)', 'nodename', expr_form='grain_pcre').keys() %}
+{%- set k8s_nodes = salt.caasp_nodes.get_with_expr('G@roles:kube-master', booted=True) %}
```
ereslibre: Missed this on previous PR, maybe `bootstrapped` is closer to what we mean here.

ereslibre: Yeah, I mean that we could rename it. Anyway, it's a minor thing.

inercia: Type `bootstrapped` five times in a row... Now you see why I named it `booted` :-D

PR #466, opened 2018-04-05
Reject keys of removed nodes instead of just deleting them.

inercia:
Reject keys of removed nodes instead of just deleting them.

https://bugzilla.suse.com/show_bug.cgi?id=1087062

bsc#1087062

### DIFF
--- a/salt/orch/removal.sls
+++ b/salt/orch/removal.sls
@@ -340,7 +340,8 @@ shutdown-target:
 # (it will appear as "unaccepted")
 remove-target-salt-key:
   salt.wheel:
-    - name: key.delete
+    - name: key.reject
+    - include_accepted: True
     - match: {{ target }}
     - require:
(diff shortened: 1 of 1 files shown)


salt/orch/removal.sls
```diff
@@ -340,7 +340,7 @@ shutdown-target:
 # (it will appear as "unaccepted")
 remove-target-salt-key:
   salt.wheel:
-    - name: key.delete
+    - name: key.reject
```
ereslibre: For this to work you need to include `include_accepted: True`.

ereslibre:
Oh missed the bug, sorry. Okay, then I think it's better to reject on salt and destroy on Velum.

salt/orch/removal.sls
inercia: Right. Fixed.

PR #467, opened 2018-04-06
Feature gates

flavio:
## Preamble

Right now this PR is made by two commits. The first one is pretty straight forward, the second one is more complicated and needs some discussion.

## The requirement

Kubernetes feature-gates are going to be controlled via a dedicated pillar. We won't expose this pillar inside of Velum because these settings are "dangerous" for the average user; plus they can enable features that we don't support.

The idea is to provide instructions to users who deliberately want to run into this unsupported area of kubernetes. These instructions will involve using some CLI commands.

## Being generic

The first commit allows any kind of feature-gate to be enabled or disabled. This is great because we don't have to provide dedicated boolean pillars for each kind of feature that is supported by kubernetes.

That makes our users independent from us :wink: 

## Going the extra mile (RFC)

The first commit isn't enough to make the feature consumable by our users: we have to expose the pillar through Velum to allow our users to override it.

That means:
  * Extend Velum database: create a new table to store feature-gates (we cannot reuse the generic table we currently have).
  * Extend Velum controller: read the contents of the new table and returns the new pillar through the API.

Finally, from a user perspective toggling feature-gates via the database isn't that fancy. I thought we could do something nicer, so here we go...

Thanks to the second commit users just have to create (or edit) a `/etc/salt/customizations.sls` file on the admin node (outside of any container!) with the custom values chosen by the user. Finally, the user has to issue a `saltutils.sync_all` statement on the salt master.

Basically we would allow our users to quickly overload our pillars by just writing a simple text file.

## Getting even more minimal

While working on this PR I found out we already have a way to provide generic arguments to all the kubernetes components (api-server, kubelet, ...):

https://github.com/kubic-project/salt/blob/85f8aed775b7e358e184a9f9ce78e62deff8c4d3/pillar/args.sls#L2-L17

Right now these pillars are not exported by Velum, which makes them unusable for regular users.

If we accept this `customization.sls` proposal we would allow users to provide custom args to kubernetes components. That would perfectly solve [this FATE request](https://fate.suse.com/324045).

We could even go one step further and remove the feature gates pillar and just instruct our users to take advantage of the `components` pillar.

**Note well:** as you can see the `/srv/pillar` directory is already used by the cloud guys. If you like this approach we would have to clean a little the code to ensure we don't overstep.

What are your thoughts?

### DIFF
--- a/pillar/args.sls
+++ b/pillar/args.sls
@@ -15,3 +15,9 @@ components:
   proxy:
     # Extra arguments to be passed to kube-proxy.
     args: ''
+
+# kubernetes feature gates to be enabled
+# https://kubernetes.io/docs/reference/feature-gates/
+# params passed to the --feature-gates cli flag.
+kubernetes:
+  feature_gates: ''
--- a/salt/kubernetes-common/config.jinja
+++ b/salt/kubernetes-common/config.jinja
@@ -24,3 +24,16 @@ KUBE_ENABLE_DAEMONSETS=true
 # How the controller-manager, scheduler, and proxy find the apiserver
 # note: we will configure everything with a kubeconfig file
 # KUBE_MASTER=""
+
+# Kubernetes feature gates, by default no feature gate is enabled.
+# Customers can enable feature gates **at their risk** by executing the
+# following command on the admin node:
+#
+#   docker exec $(docker ps | grep velum-dashboard | awk {'print $1'}) entrypoint.sh bundle exec rails runner "Pillar.apply(kubernetes_feature_gates: 'DevicePlugins=true')"
+#
+# This will set the KUBE_FEATURE_GATES flag to `DevicePlugins=true`
+{%- if salt.caasp_pillar.get('kubernetes:feature_gates', '') %}
+KUBE_FEATURE_GATES="--feature-gates={{ pillar['kubernetes']['feature_gates'] }}"
+{%- else %}
+KUBE_FEATURE_GATES=""
+{% endif %}


salt/_modules/caasp_pillar.py
```diff
@@ -26,3 +26,16 @@ def get(name, default=''):
         elif res.lower() in ["false", "no", "off"]:
             return False
     return res
+
+
+def get_feature_gates_flags():
```
inercia: Having this function only for the _feature_gates_ does not look right to me. I would generalize it, something like:

```Python
def get_map(name, separator):
    return ",".join(['{0}{2}{3}'.format(key, separator, fg[key]) for key in get(name, {})])
```

pillar/args.sls
```diff
@@ -15,3 +15,14 @@ components:
...
+# kubernetes feature gates to be enabled
+# https://kubernetes.io/docs/reference/feature-gates/
+# keys: name of the feature gate
+# values: boolean value that toogles the feature
+#
+# Example:
+#
+# feature_gates:
+#   DevicePlugins: true
+feature_gates: {}
```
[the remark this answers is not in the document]
ereslibre: +1

salt/_modules/caasp_pillar.py
ereslibre: Either generalize it here, or have another module `caasp_kubernetes` with specific kubernetes slang.

config/master.d/50-master.conf
```diff
@@ -21,6 +21,7 @@ file_roots:
 pillar_roots:
   base:
     - /usr/share/salt/kubernetes/pillar
+    - /srv/pillar
```
ereslibre: Will this approach work for our purposes?

```
caasp-admin:~ # mkdir -p /srv/pillar
mkdir: cannot create directory ‘/srv/pillar’: Read-only file system
```

ereslibre: If this override needs to be persistent, we'd need to mount this folder from somewhere on the host, why don't use the regular pillar override? We could provide a group of rake tasks that make it easy to configure feature gates, and they'd be saved on the database along with the rest of the pillars.

Note well: I'd love to kill the HTTP pillar API, but that's a different discussion :)

flavio:
I think I didn't explain myself properly, none of you understand how the pillar override is done. My fault :smile: 

The `customizations` file referenced inside of `top/pillars.sls` is not a leftover. This is a file that is **not** part of our git repository. The user should create a `/etc/salt/customizations.sls` file on the admin node.

If the file does not exist nothing bad will happen, everything will just work as expected. On the other hand, if the file exists, things will get interesting because this is the last pillar loaded inside of our `pillars/top.sls` file. Hence whatever is defined in here it will override our pillar values. Well... not all the pillars will be overridden, the ones exposed by our Velum API are immutable, because they override everything.

So in this case the proposal is to never expose the feature gates pillars through Velum. The only way to enable them is by creating this `/etc/salt/customizations.sls` file on the admin node with something like:

```yaml
feature_gates:
  DevicePlugins: true
```

Now, as to @ereslibre's question, this file is already mounted inside of our image because, due to changes introduced by the public cloud team, the `/etc/salt` directory on the admin node is already mounted to the right location inside of the salt-master container. If we decide to accept this I'll have to clean some little bits and pieces to avoid the duplication between the cloud changes and ours.

config/master.d/50-master.conf
kiall: We should probably avoid hijacking `/srv/pillar` from the public cloud guys - I think it would be better to use `/etc/caasp/pillar-customizations/` (both inside, and outside the container..)

pillar/top.sls
```diff
@@ -9,6 +9,7 @@ base:
     - docker
     - registries
     - schedule
+    - customizations
```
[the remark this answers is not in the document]
kiall: Ah. I see this now. The user creates that `customizations.sls`.

I **think**, if we just specify a new pillar root of `/etc/caasp/pillar-customizations/` - The operator can put another top.sls in there, and then whatever pillar files they want.. 

That said.. I'm not 100% certain the top.sls's will be additive within a single salt environment. If not, how does salt handle a missing pillar that's been referenced in top.sls?

pillar/args.sls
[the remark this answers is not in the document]
ereslibre: I still think namespacing this is a good idea.

config/master.d/50-master.conf
flavio: I think they are using that for the very same reason:
https://github.com/kubic-project/caasp-container-manifests/blob/d6b19f7b8bc6f8e57ac4099d35bb3382669afc99/manifests/public.yaml#L169-L171

Which is backed by:
https://github.com/kubic-project/caasp-container-manifests/blob/d6b19f7b8bc6f8e57ac4099d35bb3382669afc99/manifests/public.yaml#L558-L560

So we already have this directory mounted in. I don't think they have anything against renaming `public-cloud-config` to something more generic as `user-pillars`.

@bear454: what do you think about that?

flavio:
@kiall, I dunno why I cannot reply in the proper location...

>I think, if we just specify a new pillar root of /etc/caasp/pillar-customizations/ - The operator can put another top.sls in there, and then whatever pillar files they want..

Oh, I see. I can give it a try.

> That said.. I'm not 100% certain the top.sls's will be additive within a single salt environment. If not, how does salt handle a missing pillar that's been referenced in top.sls?

Right now salt doesn't complain about the missing `customizations.sls` pillar. Everything is working as expected also when there's so such a file.

config/master.d/50-master.conf
bear454: You can name the mount whatever you want, really, so long as you don't break import of the pillar file we put there.

That being said, I'm not sure `user-pillars` is the best choice; if someone thinks they have permission to modify our pillar it's going to be a support nightmare. _The public cloud pillar is definitely not intended for user modification._

bear454:
@flavio why not create the `customizations.sls`, but with just comments?

pillar/args.sls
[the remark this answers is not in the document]
inercia: +1

config/master.d/50-master.conf
kiall: Agree with @bear454  - I think we should keep the public cloud stuff separate to the unsupported "tweaks" an operator can make.

I think making the scope of this too wide is dangerous for supportability reasons, but we've found enough times already that customers are going to make tweaks. And we're going to actively undo those tweaks, making this better - while keeping it as unsupported is IMO the goal :)

kiall: As a side note - the "prod" env name and generic "/srv/pillar" folder used for public cloud stuff should ideally be renamed to something more specific to avoid confusion :)

salt/kube-apiserver/apiserver.jinja
```diff
@@ -34,6 +34,9 @@ KUBE_API_ARGS="--advertise-address={{ salt.caasp_net.get_primary_ip() }} \
                --cloud-config=/etc/kubernetes/openstack-config \
   {%- endif -%}
 {% endif %}
+{%- if pillar['feature_gates'] %}
+               --feature-gates={{ salt.caasp_pillar.get_feature_gates_flags() }} \
```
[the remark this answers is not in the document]
flavio: I can do that, but this would require to:

  * Introduce a new `FEATURE_GATES` environment variable inside of `/etc/kubernetes/config`
  * Change all the kubernetes package to have all the systemd units  reference the`FEATURE_GATES` environment variable on their `ExecStart`. 

Is everybody fine with that?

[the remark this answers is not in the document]
kiall: I think it would be better to create a `KUBE_COMMON_ARGS=` in /etc/kubernetes/config, and then update the services to use it - that way all common options can move there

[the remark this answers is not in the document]
flavio: Good idea. I'll take care of that

flavio:
So, it looks like there's no way to have multiple pillar roots merged together aside from using an external pillar. This is what I got so far as response on our internal salt mailing list.

I don't want to go that way for this feature...

I see two possible approaches.

## 1. Change the data model used to represent feature gates

Instead of having the `feature_gates` stored as a dictionary we would keep them as a flat string (basically the whole args of the cli flag). By doing that we can store this string into the generic pillars table we already have.

**Impact on us:** requires a small additional PR against Velum to make it aware of this new pillar.
**Impact on the users:**  feature gates can be toggled by altering a single row inside of the database. This is not the best user experience, but this is an unsupported feature plus it's the same approach used to manage other pillars.

## 2. Use overlayfs

Right now our pillars are stored inside of a  read-only directory on the admin node. We could create a new overlayfs mount specific for this directory. That would produce a writable copy of our pillar files on a specific location (eg: `/etc/caasp/pillars`).

**Impact on us:**
  * We have to create the overlayfs mount. That won't be hard.
  * We have to change the pod spec file to mount inside of the salt-master container the overlayfs mount instead of the read-only directory.
  * We should update our support-utils plugin to look for modifications done by the user.

**Impact on the customer:** best experience ever!

## My opinion

Let's be honest, this aspect (how to override pillars) is completely unrelated with this PR. We can discuss that in a broader scope with a separate PR.

I would just go with the first option.

flavio:
I updated the PR and went with the 1st approach as suggested by @kiall over irc.

salt/kubernetes-common/config.jinja
```diff
@@ -24,3 +24,16 @@ KUBE_ENABLE_DAEMONSETS=true
 # How the controller-manager, scheduler, and proxy find the apiserver
 # note: we will configure everything with a kubeconfig file
 # KUBE_MASTER=""
+
+# Kubernetes feature gates, by default no feature gate is enabled.
+# Customers can enable feature gates **at their risk** by executing thes
```
ereslibre: Typo `thes`

ereslibre:
I would like to revamp how we think about pillars and the whole API idea for fetching them, so that's as you said a complete different discussion.

My take on this is: solution 1) works well for us at the moment, we'll go through a big revamp of the architecture very soon and I think this is the best way to go at the moment, with no threats or doubts.

About:

> Right now our pillars are stored inside of a read-only directory on the admin node. We could create a new overlayfs mount specific for this directory. That would produce a writable copy of our pillar files on a specific location (eg: /etc/caasp/pillars)

While I think this can work, I think this can be an impediment if we in a near future work with a multi-master salt setup.

We have all customization possibilities on the database-based pillar override, for me: let's keep using that until we completely revamp the whole thing.

bear454:
@ereslibre said:
> I would like to revamp how we think about pillars and the whole API idea for fetching them, so that's as you said a complete different discussion.

Again? In 2.x it's database; in 3.x it's a REST API, and you want to change it _again_ for 4.x ?

kiall:
> The idea would be that Velum when launches an orchestration instead of e.g. just doing a salt-run state.orchestrate orch.kubernetes in the case of the bootstrap orchestration would do something like: salt-run state.orchestrate orch.kubernetes pillar='all_pillars_dumped_here_as_json'

This does have a whole different issue tho - manual salt commands just became so much harder, and per-machine pillars likely go away. The first can probably be solved easily enough, but the second will become important down the line. For example, if we allow more advanced network configurations - where we have multiple physical NICs with different classes of traffic assigned to each. With non-homogeneous hardware, you need to be able to configure each machine slightly differently - e.g. the PCI ID of each NIC might differ on different machines...

Anyway - We're way off topic :)

ereslibre:
> Anyway - We're way off topic :)

Yes...

> This does have a whole different issue tho - manual salt commands just became so much harder, and per-machine pillars likely go away. The first can probably be solved easily enough, but the second will become important down the line. For example, if we allow more advanced network configurations - where we have multiple physical NICs with different classes of traffic assigned to each. With non-homogeneous hardware, you need to be able to configure each machine slightly differently - e.g. the PCI ID of each NIC might differ on different machines...

Just a small comment. The pillar that gets dumped is only the overriden pillar (what gets served right now through API, that is not machine-aware either). If we want special configs for each machine then we can use the normal pillar [if it's known beforehand] or grains configured through the Velum interface on each machine with that special config (the case you presented, for example).

But yes, very offtopic at the moment.

PR #469, opened 2018-04-10
Include `ca-cert` and `cert` for the admin node explicitly

ereslibre:
It is included as a side effect because `velum/init.sls` includes them,
so make it explicit on the top file for the admin too.

feature#deployment-stability


ereslibre:
We could also completely get rid of `ca-cert` and `cert` here. In any case I think it's good to also be consistent on the admin here.

PR #470, opened 2018-04-10
Remove unneeded includes `ca-cert` and `cert` for `velum/init.sls`

ereslibre:
feature#deployment-stability


ereslibre:
This PR just fails to checkout using the remote ref of the PR, looks like a github glitch. Opening a new one with the same content!

PR #471, opened 2018-04-10
[WIP] Node addition orchestration

inercia:
# Description

New orchestration for adding nodes: instead of using the regular bootstrap orchestration for adding nodes, it would be better to use a more specific one that could:
    
* migrate etcd servers from minions to masters
* highstate only the nodes that are really affected by the new node
* resist nodes that are unresponsive
    
feature#node_addition


salt/orch/addition.sls
```diff
@@ -0,0 +1,120 @@
+# must provide the node (id) to be removed in the 'target' pillar
+{%- set target = salt['pillar.get']('target') %}
```
ereslibre: We allow several node addition at the same time, so this should be a list.

PR #472, opened 2018-04-10
Lowercase hostnames

FedericoCeratto:
Related to https://bugzilla.suse.com/show_bug.cgi?id=1087115

### DIFF
--- a/salt/_grains/caasp_nodename.py
+++ b/salt/_grains/caasp_nodename.py
@@ -1,0 +1,18 @@
+#!/usr/bin/python
+
+import salt.grains.core
+
+def caasp_nodename():
+    """
+    Override core.hostname grain to provide lowercased items
+    localhost, fqdn, host, domain, nodename
+    """
+    hn = salt.grains.core.hostname()
+    os_data = salt.grains.core.os_data()
+    return dict(
+        domain = hn['domain'].lower(),
+        fqdn = hn['fqdn'].lower(),
+        host = hn['host'].lower(),
+        localhost = hn['localhost'].lower(),
+        nodename = os_data['nodename'].lower()
+    )


salt/_grains/caasp_nodename.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/python
+
+import salt.grains.core
+
+def hostname():
```
inercia: Considering that the only function in this file is `hostname()`, wouldn't make more sense to name this file `caasp_hostname`?

salt/_grains/caasp_nodename.py
```diff
@@ -0,0 +1,15 @@
...
+    """
+    Override core.hostname grain to provide lowercased items
+    localhost, fqdn, host, domain
+    """
+    grains = salt.grains.core.hostname()
+    grains['domain'] = grains['domain'].lower()
+    grains['fqdn'] = grains['fqdn'].lower()
+    grains['host'] = grains['host'].lower()
+    grains['localhost'] = grains['localhost'].lower()
+    return grains
```
inercia: Instead of returning a dictionary with many items that are already in the global grains, I think this could be optimized and return a dictionary with only the overridden grains. Something like:

```Python
    grains = salt.grains.core.hostname()
    return {
        'domain': grains['domain'].lower(),
        'fqdn': grains['fqdn'].lower(),
        'host': grains['host'].lower(),
        'localhost': grains['localhost'].lower()
    }
```

kiall: We currently use `grains['nodename']` everywhere - so that's the one we most need to override!

salt/_grains/caasp_nodename.py
kiall: Actually, Does salt not require that the function name and file name matches?

salt/_grains/caasp_nodename.py
inercia: Right, `nodename` is the most important grain for us...

ereslibre: Aside from this, we need to double check, because without checking it looks to me that when the salt-minion starts (without having synced grains with the master), it will run its mine functions (https://github.com/kubic-project/salt/blob/143b3e663a4a8bc1f3fe3e2835156e82238168b2/pillar/mine.sls#L5-L7), and probably publish on the mine the `nodename` containing upper letters (we use a lot the pattern `mine.get` with `fun=nodename`).

salt/_grains/caasp_nodename.py
```diff
@@ -0,0 +1,17 @@
...
+    Override core.hostname grain to provide lowercased items
+    localhost, fqdn, host, domain, nodename
+    """
+    grains = salt.grains.core.hostname()
+    return dict(
+        domain = grains['domain'].lower(),
+        fqdn = grains['fqdn'].lower(),
+        host = grains['host'].lower(),
+        localhost = grains['localhost'].lower(),
+        nodename = grains['host'].lower()
```
ereslibre: Why not use the original `nodename` grain here?

ereslibre: Side note: salt in our current version gets this information by calling to `platform.uname()`: https://github.com/saltstack/salt/blob/3e844ed1df13fea1b8fbc4d5b9d5b4f130743364/salt/grains/core.py#L1187-L1188

FedericoCeratto: @ereslibre do we expect the hostname we are getting from uname to be different from the host grain? If so - this sounds like a bug.

kiall: @FedericoCeratto blame systemd and wicked :) Systemd has 3 different hostnames - static, transient and pretty. These no guarantee these line up, and no guarantee these match uname, /etc/hostname, etc. Wicked also updates some, but not all, of these when for example a DHCP hostname is provided.

ereslibre: TBH, I'd prefer not to do any assumptions here and just transform the very same grain we've been working with. Worst case, you cannot access to the original grain because it's still being populated I think the safest bet here would be us calling `platform.uname()` to retrieve the `nodename`, if we cannot read the original grain itself and transform it (that for me would be the ideal case -- since it's something we can forget about for future salt versions).

inercia: I agree with @ereslibre : we should just do a `nodename = grains['nodename'].lower()`

ereslibre:
@kiall As far as I can tell, it's documented: https://docs.saltstack.com/en/latest/topics/grains/#precedence, so we should be fine.

kiall:
> @kiall As far as I can tell, it's documented: https://docs.saltstack.com/en/latest/topics/grains/#precedence, so we should be fine.

kinda - there is details on how things will override each other once grains are loaded and your states start, but not necessarily much detail on using other lower precedence grains during a custom grain load.. Hence, reusing the logic rather than assuming the grain is ready to be consumed :)

PR #473, opened 2018-04-10
Fix race condition on update-etc-hosts

inercia:
fix#update-etc-hosts

### DIFF
--- a/salt/_modules/caasp_nodes.py
+++ b/salt/_modules/caasp_nodes.py
@@ -130,8 +130,8 @@ def get_with_expr(expr, **kwargs):
     if kwargs.get('exclude_in_progress', False):
         expr_items.append('not G@bootstrap_in_progress:true')
         expr_items.append('not G@update_in_progress:true')
-        expr_items.append('not G@removal_in_progress:true')
-        expr_items.append('not G@addition_in_progress:true')
+        expr_items.append('not G@node_removal_in_progress:true')
+        expr_items.append('not G@node_addition_in_progress:true')
 
     excluded = _sanitize_list(kwargs.get('excluded', []))
     if excluded:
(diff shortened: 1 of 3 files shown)


ereslibre:
Not strictly related to this change, but I see `addition_in_progress` grain (global) on salt/_modules/caasp_nodes.py and salt/_modules/tests/test_caasp_nodes.py, since we are not setting it anywhere, could we also clean those up on this PR?

salt/_modules/caasp_nodes.py
```diff
@@ -130,8 +130,8 @@ def get_with_expr(expr, **kwargs):
     if kwargs.get('exclude_in_progress', False):
         expr_items.append('not G@bootstrap_in_progress:true')
         expr_items.append('not G@update_in_progress:true')
-        expr_items.append('not G@removal_in_progress:true')
-        expr_items.append('not G@addition_in_progress:true')
+        expr_items.append('not G@node_removal_in_progress:true')
+        expr_items.append('not G@node_addition_in_progress:true')
```
ereslibre: These two look inconsistent with the other `expr_items` added.

salt/_modules/caasp_nodes.py
```diff
@@ -466,8 +466,8 @@ def get_expr_affected_by(target, **kwargs):
     if kwargs.get('exclude_in_progress', True):
         affected_items.append('not G@bootstrap_in_progress:true')
         affected_items.append('not G@update_in_progress:true')
-        affected_items.append('not G@removal_in_progress:true')
-        affected_items.append('not G@addition_in_progress:true')
+        affected_items.append('not G@node_removal_in_progress:true')
+        affected_items.append('not G@node_addition_in_progress:true')
```
ereslibre: Same here.

PR #474, opened 2018-04-11
Remove unneeded includes `ca-cert` and `cert` for `velum/init.sls` and `ldap/init.sls`

ereslibre:
feature#deployment-stability

### DIFF
--- a/salt/top.sls
+++ b/salt/top.sls
@@ -4,6 +4,7 @@ base:
     - ca
   'roles:(admin|kube-(master|minion))':
     - match: grain_pcre
+    - ca-cert
     - swap
     - etc-hosts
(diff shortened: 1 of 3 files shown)


salt/top.sls
```diff
@@ -4,6 +4,7 @@ base:
     - ca
   'roles:(admin|kube-(master|minion))':
     - match: grain_pcre
+    - ca-cert
```
ereslibre: The `ca-cert` is required in the admin because as part of the orchestration we try to connect to dex using the ca file present in `/etc/pki/trust/...` as in the rest of the machines. I won't harm explicitly making the admin include `ca-cert` too.

PR #482, opened 2018-04-16
Node removal: don't cleanup some things...

inercia:
This PR relaxes the `removal` orchestration by not cleaning up some _not-so-important_ things.

feature#node_remooval


ereslibre:
I'm not so sure about this, is this really a thing? It looks to me like it adds unnecessary clutter for very minor things that in a normal environment should not fail... no?

mjura:
I have this same question like @ereslibre. Is this really required to add?

salt/cleanup/remove-pre-reboot.sls
```diff
@@ -2,24 +2,10 @@
 # try to remove some dirs that could contain sensitive
 # information, even when they were not directly managed by us
 
-wipe-etc-kubernetes:
-  cmd.run:
-    - name: rm -rf /etc/kubernetes/*
```
mjura: I would keep it, in case if CaaSP will be deployed with CPI it will contains Public/OpenStack Cloud credentials

salt/docker/remove-pre-reboot.sls
```diff
@@ -11,11 +11,3 @@
 
 /etc/docker/certs.d:
   file.absent
-
-/var/lib/docker:
```
mjura: I'm sure that we don't want compromise all users containers and all data which they contains. This should definitely stay

salt/etcd/remove-pre-reboot.sls
```diff
@@ -1,23 +0,0 @@
...
-
-etcd-user-removal:
-  user.absent:
-    - name: etcd
-
-etcd-group-removal:
-  group.absent:
-    - name: etcd
-
-etcd-wipe-var-lib:
```
mjura: Etcd stores all data from K8s cluster including secrets. We have to keep also

salt/kube-apiserver/remove-pre-reboot.sls
```diff
@@ -8,6 +8,3 @@
 
 {{ pillar['ssl']['kube_apiserver_key'] }}:
   file.absent
-
-/etc/kubernetes/apiserver:
```
mjura: This will provide information about K8s cluster configuration including network addresses

salt/kubectl-config/remove-pre-reboot.sls
```diff
@@ -5,9 +5,3 @@
 
 {{ pillar['ssl']['kubectl_key'] }}:
   file.absent
-
-{{ pillar['paths']['kubeconfig'] }}:
-  file.absent
-
-/root/.kube:
-  file.absent
```
mjura: This contains kubeconfig with user and tokens

kiall:
Re @mjura

> From security reason we have to clean nodes properly

There was a lot of discussion around these cleanups, the risk of failure, and so on. We decided that - we're only going to remove the certs etc needed to rejoin the cluster (etcd/kubernetes) for the moment, and will expand later as needed.

The expectation is, once a node is removed, that machine is reimaged (for bare metal...) or the VM deleted (for clouds..) to remove any leftover secrets - including secrets we're not in control of - like docker images with hardcoded secrets, or hostpath volumes, etc etc.

salt/crio/remove-pre-reboot.sls
```diff
@@ -1,13 +1,4 @@
...
+{# See https://github.com/saltstack/salt/issues/14553 #}
+dummy_step:
   cmd.run:
-    - name: |-
-        for subvolume in {{pillar['cri']['crio']['dirs']['root']}}/btrfs/subvolumes/* ; do
-          btrfs subvolume delete $subvolume
-        done
-        rm -rf {{pillar['cri']['crio']['dirs']['root']}}*
-        rm -rf {{pillar['cri']['crio']['dirs']['runroot']}}*
+    - name: "echo saltstack bug 14553"
```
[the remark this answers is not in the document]
inercia: Yeah, you are right, this is a leftover... I'll remove it.

PR #485, opened 2018-04-17
WIP: Sync modules before refreshing grains

kiall:
Modules can provide pillar and grains. We should push those modules out before we
sync the pillar and grains data.

### DIFF
--- a/salt/orch/kubernetes.sls
+++ b/salt/orch/kubernetes.sls
@@ -35,14 +35,23 @@ set-etcd-roles:
       - set-bootstrap-in-progress-flag
 {% endif %}
 
-sync-pillar:
-  salt.runner:
-    - name: saltutil.sync_pillar
+update-modules:
+  salt.function:
+    - tgt: '*'
+    - name: saltutil.sync_all
+    - kwarg:
+        refresh: True
     - require:
       - set-bootstrap-in-progress-flag
 {%- if additional_etcd_members|length > 0 %}
       - set-etcd-roles
 {%- endif %}
+
+sync-pillar:
+  salt.runner:
+    - name: saltutil.sync_pillar
+    - require:
+      - update-modules
 
 update-pillar:
   salt.function:
@@ -66,15 +75,6 @@ update-mine:
       - update-pillar
       - update-grains
 
-update-modules:
-  salt.function:
-    - tgt: '*'
-    - name: saltutil.sync_all
-    - kwarg:
-        refresh: True
-    - require:
-      - update-mine
-
 disable-rebootmgr:
   salt.state:
     - tgt: 'roles:(admin|kube-master|minion|etcd)'
@@ -82,7 +82,7 @@ disable-rebootmgr:
     - sls:
       - rebootmgr
     - require:
-      - update-modules
+      - update-mine
 
 etc-hosts-setup:
   salt.state:
--- a/salt/orch/removal.sls
+++ b/salt/orch/removal.sls
@@ -73,10 +73,10 @@ sync-all:
   salt.function:
     - tgt: '*'
     - names:
+      - saltutil.sync_all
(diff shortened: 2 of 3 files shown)


kiall:
@inercia / @ereslibre  - I've not tested this yet - but - any reasons you can think of that this might cause issues?

ereslibre:
I think this order change is not very relevant, as `sync-pillar` is ran as a `salt.runner` only on the master to detect modules under `_pillar` (`velum.py`) while `update-modules` is just a synchronization of almost everything, but they are targeting different things (one is for the master, the other is for all the minions to sync custom modules and states).

kiall:
> I think this order change is not very relevant, as sync-pillar is ran as a salt.runner only on the master to detect modules under _pillar (velum.py) while update-modules is just a synchronization of almost everything, but they are targeting different things (one is for the master, the other is for all the minions to sync custom modules and states).

The diff might be hard to read.. It's moving the sync_modules above the sync_pillar (both of them), and above the sync_grains calls.. It may not be necessary to move it above one of the sync_pillars (the one that sync into the context of the salt-master process), but I believe the other two should be below the sync_modules call

ereslibre:
> The diff might be hard to read.. It's moving the sync_modules above the sync_pillar (both of them), and above the sync_grains calls.. It may not be necessary to move it above one of the sync_pillars (the one that sync into the context of the salt-master process), but I believe the other two should be below the sync_modules call

Yes, my point is that I don't think this order change is actually doing anything, as `sync-pillar` is a `salt.runner` executed on the master discovering modules under `_pillar`, while `update-modules` is sent to all minions, and will sync all modules, states... on the minions. That's why I think the order change here is not doing much.

kiall:
> Yes, my point is that I don't think this order change is actually doing anything, as sync-pillar is a salt.runner executed on the master discovering modules under _pillar, while update-modules is sent to all minions, and will sync all modules, states... on the minions. That's why I think the order change here is not doing much.

Before, we had this:

1. Sync pillar (on master)
1. Sync pillar (on all minions)
1. Sync grains (on all minions)
1. Sync modules (to all minions)

The problem is - `sync_modules` is what pushes out the _grains/* / _pillar/* etc code. We've already refreshed the grains on all minions, and now need to do it again for it to actually take effect:

1. Sync pillar (on master)
1. Sync pillar (on all minions)
1. Sync grains (on all minions)
1. Sync modules (to all minions)
1. Sync grains again (on all minions)

What I've changed it to is this:

1. Sync modules (to all minions)
1. Sync pillar (on master)
1. Sync pillar (on all minions)
1. Sync grains (on all minions)

And, it could also have been this:

1. Sync pillar (on master)
1. Sync modules (to all minions)
1. Sync pillar (on all minions)
1. Sync grains (on all minions)

The choice between the last two doesn't really matter, however, IMO, it's nicer to keep the pillar loads all together beside each other.

salt/orch/kubernetes.sls
```diff
@@ -35,15 +35,24 @@ set-etcd-roles:
...
       - set-etcd-roles
 {%- endif %}
 
+sync-pillar:
+  salt.runner:
+    - name: saltutil.sync_pillar
+    - require:
+      - update-modules
+
 update-pillar:
```
mjura: On logic this should be before synchronizing pillars. First we are updating value and then we are synchronizing it to all. Is it right?

kiall: I'm not sure I understand the question?

mjura: I asked about order 
```
sync-pillar:
update-pillar:
```
I have checked documentation https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.saltutil.html and looks OK for now.

flavio:
Is this still a WIP?

PR #491, opened 2018-04-19
cni: add cilium as alternate to flannel plugin

nirmoy:
*This still depends on cilium container images to be created in OBS

### DIFF
--- a/salt/cni/cilium-ds.yaml.jinja
+++ b/salt/cni/cilium-ds.yaml.jinja
@@ -1,0 +1,171 @@
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: cilium
+  namespace: kube-system
+spec:
+  updateStrategy:
+    type: "RollingUpdate"
+    rollingUpdate:
+      # Specifies the maximum number of Pods that can be unavailable during the update process.
+      # The current default value is 1 or 100% for daemonsets; Adding an explicit value here
+      # to avoid confusion, as the default value is specific to the type (daemonset/deployment).
+      maxUnavailable: "100%"
+  selector:
+    matchLabels:
+      k8s-app: cilium
+      kubernetes.io/cluster-service: "true"
+  template:
+    metadata:
+      labels:
+        k8s-app: cilium
+        kubernetes.io/cluster-service: "true"
+      annotations:
+        # This annotation plus the CriticalAddonsOnly toleration makes
+        # cilium to be a critical pod in the cluster, which ensures cilium
+        # gets priority scheduling.
+        # https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/
+        scheduler.alpha.kubernetes.io/critical-pod: ''
+        scheduler.alpha.kubernetes.io/tolerations: >-
+          [{"key":"dedicated","operator":"Equal","value":"master","effect":"NoSchedule"}]
+    spec:
+      serviceAccountName: cilium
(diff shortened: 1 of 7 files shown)


salt/cilium/init.sls
```diff
@@ -0,0 +1,16 @@
...
+{% if plugin == "cilium" %}
+
+{% from '_macros/certs.jinja' import certs with context %}
+{{ certs("cilium",
+         pillar['ssl']['cilium_crt'],
+         pillar['ssl']['cilium_key'],
+         cn = grains['nodename'],
+         o = 'system:nodes') }}
+
+{% endif %}
```
inercia: As a minor comment, I would move this folder from the top level to `cni/cilium` (and use `cni/cilium` instead of just `cilium` in `top.sls`).

salt/cilium/init.sls
```diff
@@ -0,0 +1,16 @@
+include:
```
inercia: I would move this `include` inside the `if` (so nothing is done if `cilium` is not enabled) and [add a dummy target ](https://github.com/kubic-project/salt/blob/crio-cleaups/salt/etc-hosts/init.sls#L11-L16) for the `else`

salt/cni/cilium-ds.yaml.jinja
```diff
@@ -0,0 +1,171 @@
...
+          - "--disable-ipv4=$(DISABLE_IPV4)"
+#        lifecycle:
+#          postStart:
+#            exec:
+#              command:
+#                - "/cni-install.sh"
+#          preStop:
+#            exec:
+#              command:
+#                - "/cni-uninstall.sh"
```
inercia: We should use the same `preStop` for `flannel`. Otherwise, it will leave some CNI config files that can mess things up...

nirmoy:
@inercia updated my branch according to your comments also now works with psp. Let me know what do you think.

salt/cni/init.sls
```diff
@@ -48,3 +48,64 @@ include:
...
+    - name: |
+        kubectl apply --namespace kube-system -f /etc/kubernetes/addons/cilium-rbac.yaml
+    - env:
+      - KUBECONFIG: {{ pillar['paths']['kubeconfig'] }}
+    - require:
+      - kube-apiserver
+      - file:      {{ pillar['paths']['kubeconfig'] }}
+    - watch:
+      - file:       /etc/kubernetes/addons/cilium-rbac.yaml
+/etc/kubernetes/addons/cilium-ds.yaml:
```
inercia: Minor: please separate from previous block with a `\n`

nirmoy: Thanks, updated the PR.

salt/cni/init.sls
```diff
@@ -48,3 +48,66 @@ include:
       - file:      /etc/kubernetes/addons/kube-flannel-rbac.yaml
 
 {% endif %}
+
+{% set plugin = salt['pillar.get']('cni:plugin', 'cilium').lower() %}
+{% if plugin == "cilium" %}
```
ereslibre: There's no need to fetch the pillar again, this can be reduced to:

```SaltStack
{% set plugin = salt['pillar.get']('cni:plugin', 'flannel').lower() %}
{% if plugin == "flannel" %}
# flannel things
{% elsif plugin == "cilium" %}
# cilium things (this whole block)
{% endif %}

nirmoy: Thanks for reviewing, Updated.

PR #492, opened 2018-04-19
[WIP] Reorganize the crio/docker code

inercia:
Reorganize the `crio`/`docker` code under a common `cri` dir

feature#node_cleanup


salt/cri/init.sls
```diff
@@ -26,3 +26,6 @@ podman:
     # flannel network, which is useful for debugging.
     - require:
       - pkg: podman
+
+include:
+  - cri/{{ pillar['cri'][salt.caasp_cri.cri_name()]['service'] }}
```
nanoscopic: My understanding is that this line includes cri/crio/init.sls when crio is selected, and cri/docker/init.sls when docker is selected, hence docker stuffs being moved under cri to make this possible, and docker being dropped from top.sls since it is now included by this line. Assuming this is what is going on these changes look sensible.

PR #493, opened 2018-04-20
Stop using __opts__ and os_data() 

FedericoCeratto:
bsc#1087115

### DIFF
--- a/salt/_grains/caasp_nodename.py
+++ b/salt/_grains/caasp_nodename.py
@@ -1,9 +1,8 @@
 #!/usr/bin/python
 
-import salt.grains.core
+import platform
 
-assert '__opts__' in globals()
-salt.grains.core.__opts__ = __opts__
+import salt.grains.core
 
 
 def caasp_nodename():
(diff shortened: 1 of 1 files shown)


salt/_grains/caasp_nodename.py
```diff
@@ -2,15 +2,13 @@
...
-salt.grains.core.__opts__ = __opts__
-
-
 def caasp_nodename():
     """
     Override core.hostname grain to provide lowercased items
     localhost, fqdn, host, domain, nodename
     """
+    assert '__opts__' in globals()
+    salt.grains.core.__opts__ = __opts__
```
ereslibre: I know this is intended more as a test than anything else, but I don't like the idea of setting `salt.grains.core.__opts__` here.

Given that `__opts__` should be available inside the grain function, and calling  `os_grains()` has proved to give us trouble regarding `__opts__`, what doesn't really match up to the idea claimed that `__opts__` should be available inside the function; I'd say the safest bet for us is to go straight the way salt goes: https://github.com/saltstack/salt/blob/3e844ed1df13fea1b8fbc4d5b9d5b4f130743364/salt/grains/core.py#L1187-L1188.

```
(_, grains['nodename'], _, version, _, _) = platform.uname()
```

and then `lower` that.

PR #488, opened 2018-04-18
Node removal: move forced-removal actions to SLS file

inercia:
Instead of running things on the forced-removal orchestration, move actions to SLS files (so they can be shared with the regular removal orchestration).

PR #494, opened 2018-04-20
Always try to remove the etcd member

ereslibre:
Don't try to be smart here: always try to unregister the etcd member,
since `manage.up` and `manage.down` won't list the down node using
roles, as salt cannot access it.

Just try to always remove the etcd member, if it's not an etcd member
it's fine, nothing will happen.

Fixes: bsc#1090392


inercia:
I would solve this with #488, because it will do the same not only for etcd but also for the kubernetes API...

ereslibre:
Right, going to try that one and will approve.

PR #496, opened 2018-04-23
Use `expr_form` instead of `tgt_type` until we update salt

ereslibre:
This is producing an error on our current salt version:

  `Rendering SLS 'base:cleanup.remove-post-orchestration' failed: Jinja error: get()
   got an unexpected keyword argument 'tgt_type'`

Go back to using `expr_form` until we update.

feature#deployment-stability

### DIFF
--- a/salt/cleanup/remove-post-orchestration.sls
+++ b/salt/cleanup/remove-post-orchestration.sls
@@ -10,7 +10,7 @@ include:
 # k8s cluster
 ###############
 
-{%- set k8s_nodes = salt['mine.get']('roles:(kube-master|kube-minion)', 'nodename', tgt_type='grain_pcre').keys() %}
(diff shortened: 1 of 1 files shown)


inercia:
Yeah, I was seeing this error too (and I don't understand why)...

ereslibre:
> Yeah, I was seeing this error too (and I don't understand why)...

Because of our ancient salt version: https://github.com/saltstack/salt/blob/2016.11/salt/modules/mine.py#L206

inercia:
I've gone a bit deeper with #497. We can merge this and then #497 or we can go directly to #497.

ereslibre:
Yup, I see. I'd go with this one first so QA can test it while we review and merge #497.

PR #497, opened 2018-04-23
Use get_with_expr() instead of mine.get/saltutil.runner

inercia:
## Rationale

We use `mine.get` in many places just for getting etcd members, k8s masters and minions. However, this does not work in the Salt master, so we must use the `saltutil.runner` with different arguments for doing the same thing: getting some list of nodes...

So instead of using these two functions (and after some problems with argument in `mine.get` solved in #496), I think it woud be better to just use the (already present) `get_with_expr` everywhere (or some shortcuts) for getting nodes.

## Changes

* New `get_etcd_members()`, `get_masters()`, `get_minions()` for getting the etcd members, k8s masters and minions.
* Use these shortcuts as well as `get_with_expr()` instead of using mine.get().

feature#cleanup

### DIFF
--- a/salt/_modules/caasp_nodes.py
+++ b/salt/_modules/caasp_nodes.py
@@ -172,6 +172,21 @@ def get_with_expr(expr, **kwargs):
     return res
 
 
+def get_etcd_members(**kwargs):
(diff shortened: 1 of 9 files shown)


ereslibre:
What's the status of this, should we close this PR?

PR #501, opened 2018-04-25
[2.1] salt-master config: support public cloud

jgleissner:
Configure salt-master to read public cloud pillar
Configure mysql pillar to support public cloud specific datanae pillars
that have more nested levels and also require a list data set.

Depends-On: https://github.com/kubic-project/caasp-container-manifests/pull/172

### DIFF
--- a/config/master.d/50-returner.conf
+++ b/config/master.d/50-returner.conf
@@ -9,7 +9,8 @@ mysql:
 
 ext_pillar:
   - mysql:
-    # This queries allows us to override our pillar values up to a depth of 3.
+    # This queries allows us to override our pillar values up to a depth of 5
+    # and a cloud specific query with a depth of 6.
     #
     # Considering we have on our static pillar:
     #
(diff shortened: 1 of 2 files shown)


ereslibre:
[salt-master config: support public cloud](https://trello.com/c/09i9Fhxk/827-salt-master-config-support-public-cloud)

kiall:
@jgleissner I've added you to this GitHub org - you'll need to accept for PRs to get CI'd automatically :)

config/master.d/50-returner.conf
```diff
@@ -9,7 +9,8 @@ mysql:
 
 ext_pillar:
   - mysql:
-    # This queries allows us to override our pillar values up to a depth of 3.
+    # This queries allows us to override our pillar values up to a depth of 5
```
rjschwei: Should be either "These queries allow us to ..." or "This query allows us to..."

kiall:
I'm not sure why, but this change seems to cause issues for salt:

     [CRITICAL] Pillar render error: Failed to load ext_pillar mysql: 'list' object has no attribute 'split'

jgleissner:
That error is caused by a bug in salt, see https://bugzilla.suse.com/show_bug.cgi?id=1072218. The bug was fixed in SLE12, but it would still show when testing with the salt SLE12 GA package.

kiall:
> That error is caused by a bug in salt, see https://bugzilla.suse.com/show_bug.cgi?id=1072218. The bug was fixed in SLE12, but it would still show when testing with the salt SLE12 GA package.

Then, it seems like we need to get an updated salt package into the 2.1 IBS projects (though, I thought we already did - would need to double check). If we merge this, then all the builds will start failing - and when it eventually goes out to customers - will break their environments too...

jgleissner:
> Then, it seems like we need to get an updated salt package into the 2.1 IBS projects (though, I thought we already did - would need to double check). If we merge this, then all the builds will start failing - and when it eventually goes out to customers - will break their environments too...

As said, the salt package in SUSE:SLE-12:Update was fixed, and the salt-master container in SUSE:SLE-12-SP3:Update:Products:CASP20:Update was already with using it. Anything update would automatically have the fix, or am I missing something?

kiall:
> As said, the salt package in SUSE:SLE-12:Update was fixed, and the salt-master container in SUSE:SLE-12-SP3:Update:Products:CASP20:Update was already with using it. Anything update would automatically have the fix, or am I missing something?

No, not missing anything. But - The builds tested here should include that fix/updated package, which suggests either the IBS projects are setup wrong - or this is a different issue of some kind..

cc @jordimassaguerpla

jordimassaguerpla:
Sorry guys there was an old salt package in the Devel repos. A leftover when our package was newer than the one in the official SUSE repos.

jordimassaguerpla:
waiting for the builds to finish...

jordimassaguerpla:
tests are green.

kiall:
Thanks @jordimassaguerpla for finding the issue!

PR #502, opened 2018-04-26
Ensure swap is disabled before kubelet is started

flavio:
We have to ensure the swap state is executed before the kubelet service is started, otherwise kubelt won't run and this will lead to issues like the ones causing bsc#1090337

### DIFF
--- a/salt/kubelet/init.sls
+++ b/salt/kubelet/init.sls
@@ -4,6 +4,7 @@ include:
   - cert
   - kubernetes-common
   - kubectl-config
+  - swap
 
 /etc/kubernetes/kubelet-initial:
   file.managed:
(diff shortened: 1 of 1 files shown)


salt/kubelet/init.sls
```diff
@@ -82,6 +83,7 @@ kubelet:
       - file:   /etc/kubernetes/manifests
       - file:   /etc/kubernetes/kubelet-initial
       - kubelet-config
+      - cmd: unmount-swaps
```
ereslibre: Wouldn't it be better to move this requirement here? https://github.com/kubic-project/salt/blob/c3b81a6ab211e84209056a3f687fb818b2aa785e/salt/kubelet/init.sls#L71, on the `service.running` statement?

flavio: The swap has to be disabled before the service is started, is that the right location to express this requirement?

ereslibre: Sorry, this diff view confused me. Everything looks fine for me.

flavio:
Force merged, the CI failure was not caused by this issue at all. I had to merge it quickly because this is blocking staging A

PR #506, opened 2018-04-27
[2.0] When we are upgrading between releases (2.0 -> updates), we need to also highstate the admin node as we do on 3.0

ereslibre:
Without doing this, fixes like `fe3e5c8eccd9b48c` render completely
invalid as we won't regenerate the LDAP certificate when the upgrade
happens.

While `fe3e5c8eccd9b48c` fixed the issue for new bootstraps on 2.0
(because the `orch.kubernetes` orchestration highstates the admin),
this would never work with maintenance requests of an already
bootstrapped cluster.

Fixes: bsc#1069175

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -37,6 +37,14 @@ update_modules:
     - require:
       - salt: update_mine
 
+admin-setup:
+  salt.state:
+    - tgt: 'roles:admin'
+    - tgt_type: grain
+    - highstate: True
+    - require:
+      - update_modules
+
 # Get list of masters needing reboot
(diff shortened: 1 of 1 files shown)


ereslibre:
[When upgrading from 2.0 GM to 2.0 with updates, the LDAP certificate can still be invalid](https://trello.com/c/TwFSe9O8/830-when-upgrading-from-20-gm-to-20-with-updates-the-ldap-certificate-can-still-be-invalid)

jordimassaguerpla:
Please backport this to 2.1.

ereslibre:
@jordimassaguerpla it was backported in https://github.com/kubic-project/salt/pull/506

PR #510, opened 2018-04-30
No not require an external_fqdn: try to guess an IP if no name provided.

inercia:
We are currently requiring an external API fqdn, but it is not extrictly neccessary as we could just get the IP address of one of the masters.

feature#cleanup

### DIFF
--- a/salt/_macros/certs.jinja
+++ b/salt/_macros/certs.jinja
@@ -27,7 +27,7 @@
                        "kubernetes.default.svc",
                        "api",
                        "api." + pillar['internal_infra_domain'],
-                       salt.caasp_pillar.get('api:server:external_fqdn'),
+                       salt.caasp_net.get_external_api_fqdn(),
                        salt.caasp_pillar.get('api:cluster_ip')] +
                        salt.caasp_pillar.get('api:server:extra_names', []) +
                        salt.caasp_pillar.get('api:server:extra_ips', []) +
--- a/salt/_modules/caasp_net.py
+++ b/salt/_modules/caasp_net.py
@@ -5,6 +5,11 @@
 
 from __future__ import absolute_import
 
+# note: do not import caasp modules other than caasp_log
+from caasp_log import abort, debug, error, info, warn
+
+
+EXTERNAL_API_FQDN_PILLAR = 'api:server:external_fqdn'
 
 def __virtual__():
     return "caasp_net"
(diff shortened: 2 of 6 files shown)


inercia:
> There's no circumstances I can think of where a deployment will end up without this pillar being set - Velum always sets it? I would much rather we didn't potentially mask a real error by guessing in salt...

I don't think it is a good idea to relay on Velum for setting values on the Salt layer, as, for example, syscalls do not relay on libc to use the right parameters. In my view, the Salt layer must be independant of upper layers and should be treated as a black box.

The problem is that we can currently set `None` as the external fqdn and it will be replaced in `/etc/hosts` without any further check or failure, so we need at least some checks on the Salt side. But we can also find a reasonable value that would work in environments where there is no real external FQDN (most of them).

Besides that, there are many pillars we are currently setting but could have reasonable, static, default values (ie, [all the ldap pillars](https://github.com/kubic-project/salt/blob/do_not_require_external_fqdn/pillar/params.sls#L163-L175)). And for other pillars that must be set dinamically and could lead to some problems (like this external API fqdn) I think we should at least 1) create _"getters"_ for those values that verify their correctness 2) create unit/integration tests too.

kiall:
> I don't think it is a good idea to relay on Velum for setting values on the Salt layer, as, for example, syscalls do not relay on libc to use the right parameters.

There is no other way to deploy CaaSP, so if we can't rely on that - we need to fix Velum, not mask the issue in Salt.

> In my view, the Salt layer must be independant of upper layers and should be treated as a black box.

I partially agree, salt should certainly be more independent - but IMO, it should also be way "dumber". Most of our issues come from salt being "too smart" - i.e. discovery / mine and guessing etc.. We need to drastically simplify our salt states, rather than make them smarter. Salt is a great tool at taking a set of inputs, and applying those onto machines, it fails miserably at generating those inputs for itself - it's just not designed to do it, but we're bolting more and more of that on rather than really fixing the underlying reasons.

> The problem is that we can currently set None as the external fqdn

We need to fix that in Velum if setting "None" is possible - otherwise, when a user gives a bad input, all they get is 100k salt events to dig through and a bunch of red X's. Or - worse - salt takes a guess, the system goes into production, and down the line, everything fails - the IP we guessed wasn't statically assigned, and the lease expired.

> and it will be replaced in /etc/hosts without any further check or failure, so we need at least some checks on the Salt side.

Safety checks in salt are IMO just fine - i.e. something like this:

    check-external-fqdn-valid:
      caasp_validate:
        type: fqdn_or_ip
        pillar: external_fqdn

> But we can also find a reasonable value that would work in environments where there is no real external FQDN (most of them).

This is simply masking the failure in Velum. There is no reason Velum should ever have no set this value. If this is happening, then we must fix Velum.

> Besides that, there are many pillars we are currently setting but could have reasonable, static, default values (ie, all the ldap pillars). And for other pillars that must be set dinamically and could lead to some problems (like this external API fqdn) I think we should at least 1) create "getters" for those values that verify their correctness 2) create unit/integration tests too.

We've (ab)used pillars as "constants" - at a guess, 90% of our pillars are not user editable, and shouldn't be user editable. For all the ones that can be set dynamically, we should IMO have no default values at all, and a test to ensure it's been set before starting the orchestrations.

Anything else is just hiding issues, duplicating default values between Velum, Salt's `pillar/*.sls` files, and often also duplicating throughout the states themselves through `pillar.get(name, <duplicated default>)`

PR #511, opened 2018-05-02
Harden the waiting for CRI socket to become active

kiall:
* Allow more time for the CRI socket to become active - 60 seconds
* Explicitly fail if the socket does not become active within this
  time.

Related to bsc#1091419

### DIFF
--- a/pillar/cri.sls
+++ b/pillar/cri.sls
@@ -1,6 +1,6 @@
 cri:
   chosen: 'docker'
-  socket_timeout: 10
+  socket_timeout: 20
   docker:
     description: Docker open-source container engine
     package: docker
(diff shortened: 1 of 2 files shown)


flavio:
Now that I looked deeper into the cause of [bsc#1091419](https://bugzilla.suse.com/show_bug.cgi?id=1091419) I don't think this might be a good idea.

Basically on docker nodes the `dockershim` socket is created by the `kubelet` process. If something is wrong with kubelet this file won't be there and we will just end up having a longer time before we get a failure.

I think the proper fix could be along the way of requiring the kubelet service to be running whenever we are making assumptions around containers.

flavio:
As discussed with @kiall let's go with 20 seconds instead of 60 seconds. Then let's work on a better fix with another PR (I can take care of that if you want)

PR #512, opened 2018-05-03
Install system wide certificates from pillars.

bergmannf:
`cert`-state will install the certificates as trust anchors.

### DIFF
--- a/salt/cert/init.sls
+++ b/salt/cert/init.sls
@@ -7,3 +7,29 @@ include:
          pillar['ssl']['crt_file'],
          pillar['ssl']['key_file'],
          o = pillar['certificate_information']['subject_properties']['O']) }}
+
+#######################################
+# additional system wide certificates #
+#######################################
+
+# Install additional certificates that were setup in Velum by the user as
+# system-wide certificates
+
+{% set system_certs = salt.caasp_pillar.get('system_certificates', []) %}
+{% for cert in system_certs %}
+  {% set name, cert = cert['name'], cert['cert'] %}
+
+/etc/pki/trust/anchors/{{ name }}.crt:
+  file.managed:
+    - makedirs: True
+    - user: root
+    - group: root
+    - mode: 644
+    - contents: |
+        {{ cert | indent(8) }}
+  cmd.run:
+    - name: update-ca-certificates
+    - onchanges:
+        - file: /etc/pki/trust/anchors/{{ name }}.crt
+
+{% endfor %}


bergmannf:
[Providing self-signed CA for CaaSP](https://trello.com/c/nNv85jtT/808-providing-self-signed-ca-for-caasp)

salt/cert/init.sls
```diff
@@ -7,3 +7,22 @@ include:
...
+#######################################
+
+# Install additional certificates that were setup in Velum by the user as
+# system-wide certificates
+
+{% set certs = salt.caasp_pillar.get('system_certificates', []) %}
+{% for cert_tuple in certs %}
+  {% set name, cert = cert_tuple %}
+
+/etc/pki/trust/anchors/{{ name }}.crt:
```
kiall: Don't we need to issue an `update-ca-certificates` call whenever one of these certs changes?

salt/cert/init.sls
```diff
@@ -7,3 +7,22 @@ include:
...
+  file.managed:
+    - makedirs: True
+    - contents: |
```
kiall: We should explicitly set the user/group to root/root and mode to 644 or similar.

salt/cert/init.sls
bergmannf: That's true - I added an `onchanges` run, but that would run after every certificate.

If the user adds a many certificates that's really wasteful though, but I am not sure how to only run it after the loop if changes occured (like in `ansible` I would use register, but I didn't find an equivalent for salt).

inercia: Does the `update-ca-certificates` really work on CaaSP? I remember we had some problems in the past with this command beacuse of the _read-only_ filesystrem...

flavio: Yes it does nowadays

bergmannf: @inercia I just ran it on a openstack deployment and it worked, so seems it does.

salt/cert/init.sls
```diff
@@ -7,3 +7,29 @@ include:
...
+    - group: root
+    - mode: 644
+    - contents: |
+        {{ cert | indent(8) }}
+  cmd.run:
+    - name: update-ca-certificates
+    - onchanges:
+        - file: /etc/pki/trust/anchors/{{ name }}.crt
+
+{% endfor %}
```
inercia: I think that, if you want to run the `update-ca-certificates` only once after several changes, you could achieve it with something like:

```SaltStack
{% set system_certs = salt.caasp_pillar.get('system_certificates', []) %}
{% for cert in system_certs %}
  {% set name, cert = cert['name'], cert['cert'] %}

/etc/pki/trust/anchors/{{ name }}.crt:
  file.managed:
    - makedirs: True
    - user: root
    - group: root
    - mode: 644
    - contents: |
        {{ cert | indent(8) }}
    - onchanges_in:
        - update-ca-certificates

{% endfor %}

update-ca-certificates:
  cmd.run

```

(This is not tested). This should insert a `onchanges` in the `update-ca-certificates` blocks for each certificate we write...

MaximilianMeister: @ellisab has tested this scenario in https://github.com/kubic-project/salt/pull/537 and that didnt work. you probably need a new sls with what you suggested in https://github.com/kubic-project/salt/pull/537#discussion_r189196888

inercia: Might be related to saltstack/salt#19592. Maybe you need to use `watch` instead of `onchanges`.

kiall:
I'm OK with running it multiple times for now.. though I do agree we should work on improving with @inercia's suggestion if that works...

PR #514, opened 2018-05-04
Make crictl handling more robust

flavio:
Some of our states are now depending on `crictl` tool. All these states have to depend on the `kubelet service.running` one, otherwise the `crictl` socket won't be available and the state will fail.

Also, with these changes, the "blame" of a failure should point directly to the guilty (`kubelet` service not running for whatever reason) instead of falling on the `haproxy` one.

Finally, the check looking for `crictl` socket has been changed to ensure the socket file exists and the service is actually listening.

This will help with bugs like bsc#1091419

### DIFF
--- a/salt/_modules/caasp_cri.py
+++ b/salt/_modules/caasp_cri.py
@@ -197,14 +197,18 @@ def __wait_CRI_socket():
     at bootstrap time, where the CRI is not yet running
     but some state interacting with it is applied.
     '''
-
-    socket = cri_runtime_endpoint()
     timeout = int(__salt__['pillar.get']('cri:socket_timeout', '20'))
     expire = time.time() + timeout
 
     while time.time() < expire:
-        if os.path.exists(socket):
+        cmd = "crictl --runtime-endpoint {socket} info".format(
+                socket=cri_runtime_endpoint())
+        result = __salt__['cmd.run_all'](cmd,
+                                         output_loglevel='trace',
+                                         python_shell=False)
+        if result['retcode'] == 0:
             return
+
         time.sleep(0.3)
 
     raise CommandExecutionError(
--- a/salt/_modules/caasp_nodes.py
+++ b/salt/_modules/caasp_nodes.py
@@ -500,3 +500,15 @@ def get_super_master(**kwargs):
             return master
 
     return ''
+
+
+def is_admin_node():
+    '''
+    Returns true if the node has the 'admin' and/or the 'ca'
+    roles.
+
+    Returns false otherwise
+    '''
+
+    node_roles = __salt__['grains.get']('roles', [])
+    return any(role in ('admin', 'ca') for role in node_roles)
--- a/salt/container-feeder/init.sls
+++ b/salt/container-feeder/init.sls
@@ -19,7 +19,7 @@ container-feeder:
   service.running:
     - enable: True
     - require:
-      {% if not salt.caasp_cri.needs_docker() %}
+      {% if not salt.caasp_nodes.is_admin_node() %}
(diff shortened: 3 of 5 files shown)


flavio:
ping @jordimassaguerpla, this can be relevant for the beta release

flavio:
Don't merge, I've still to look into some corner cases :(

flavio:
It turns out that including the `kubelet` state on the admin node causes the node to be added to the kubernetes cluster. That causes some system workloads to be assigned to the admin node, but they will never succeed -> the orchestration fails after an endless time.

I've updated the PR adding comments about that. Everything seems to be working in a reliable fashion now.

flavio:
**DO NOT MERGE** - wait for beta1 to be tagged.

cc @jordimassaguerpla

salt/_modules/caasp_filters.py
```diff
@@ -40,3 +40,15 @@ def get_max(seq):
     # TODO: use the builtin filter (https://docs.saltstack.com/en/latest/topics/jinja/index.html#max)
     #       once we Salt>2017.7.0
     return max(seq)
+
+
+def is_admin_node():
```
inercia: Maybe it would make more sense to move this function to [`caasp_nodes.py`](https://github.com/kubic-project/salt/blob/master/salt/_modules/caasp_nodes.py), as there are many functions for getting nodes by their roles... The function couold then be renamed to `is_admin`, so we would call it with `salt.caasp_nodes.is_admin(...)`

salt/_modules/caasp_filters.py
```diff
@@ -40,3 +40,15 @@ def get_max(seq):
...
+    '''
+    Returns true if the node has the 'admin' and/or the 'ca'
+    roles.
+
+    Returns false otherwise
+    '''
+
+    node_roles = __salt__['grains.get']('roles', [])
```
inercia: It would be safer to use [this function](https://github.com/kubic-project/salt/blob/master/salt/_modules/caasp_grains.py#L12-L20) (so `node_roles = __salt__['caasp_grains.get']('roles', [])`. Otherwise this will crash is someone tries to use `is_admin_node()` in a `orch/*.sls` file.

jordimassaguerpla:
@flavio : merging is unblocked now. You can merge into master and then create a PR backporting it to release-3.0

vrothberg:
`crictl --timeout` may come in handy here. The `/etc/crictl.yaml` is already configured at that point (pointing to either `dockershim.sock` or `crio.sock`). We could do something like `crictl --timeout 60 info` > /dev/null` to wait at most 60 seconds for the runtime endpoint to show up.

salt/_modules/caasp_filters.py
flavio: Done

salt/_modules/caasp_filters.py
flavio: As discussed we are not going to do that, this is not needed because right now this function is only called on the nodes that are executing the states.

flavio:
> crictl --timeout may come in handy here. The /etc/crictl.yaml is already configured at that point (pointing to either dockershim.sock or crio.sock). We could do something like crictl --timeout 60 info > /dev/null to wait at most 60 seconds for the runtime endpoint to show up.

Unfortunately this is not going to help. `crictl` will fail immediately if the socket file is not yet created.

vrothberg:
> Unfortunately this is not going to help. crictl will fail immediately if the socket file is not yet created.

That is unfortunate. I didn't test that scenario, but I try to address that upstream as it's something I would expect it to do.

PR #516, opened 2018-05-04
[bsc#1092140] enable psp opt-out

MaximilianMeister:
they need to be globally unique

orch error happened when setting psp to false in params.sls

fix#conflicting-sls-ids

### DIFF
--- a/salt/kube-apiserver/apiserver.jinja
+++ b/salt/kube-apiserver/apiserver.jinja
@@ -5,6 +5,7 @@
 #
 
 {% set cloud_provider = salt.caasp_pillar.get('cloud:provider') %}
+{% set kube_admission_control = ['Initializers', 'NamespaceLifecycle', 'LimitRanger', 'ServiceAccount', 'NodeRestriction', 'PersistentVolumeLabel', 'DefaultStorageClass', 'ResourceQuota', 'DefaultTolerationSeconds'] %}
 
 # The address on the local server to listen to.
 KUBE_API_ADDRESS="--insecure-bind-address=127.0.0.1 --bind-address=0.0.0.0"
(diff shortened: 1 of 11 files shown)


MaximilianMeister:
> what does setting psp to false do?

it's supposed to do the opt-out of pod security policies

MaximilianMeister:
@jordimassaguerpla it seems that setting `psp` to false makes bootstrap fail (see https://github.com/kubic-project/salt/pull/517) - however we should test the same thing with this patch here to confirm that this is the only issue...

MaximilianMeister:
just for the record, setting `psp: false` was green http://jenkins.caasp.suse.net/blue/organizations/jenkins/salt/detail/PR-516/5/pipeline

i'll drop the commit again

salt/kube-apiserver/apiserver.jinja
```diff
@@ -23,7 +23,11 @@ KUBE_ETCD_SERVERS="--etcd-cafile={{ pillar['ssl']['ca_file'] }} \
 KUBE_SERVICE_ADDRESSES="--service-cluster-ip-range={{ pillar['services_cidr'] }}"
 
 # default admission control policies
+{% if salt.caasp_pillar.get('addons:psp', True) %}
 KUBE_ADMISSION_CONTROL="--admission-control=Initializers,NamespaceLifecycle,LimitRanger,ServiceAccount,NodeRestriction,PodSecurityPolicy,PersistentVolumeLabel,DefaultStorageClass,ResourceQuota,DefaultTolerationSeconds"
+{% else %}
+KUBE_ADMISSION_CONTROL="--admission-control=Initializers,NamespaceLifecycle,LimitRanger,ServiceAccount,NodeRestriction,PersistentVolumeLabel,DefaultStorageClass,ResourceQuota,DefaultTolerationSeconds"
```
ereslibre: I think we can better have a real array for this, something like:

```
{% set kube_admission_control = ['Initializers', 'NamespaceLifecycle', 'LimitRanger', 'ServiceAccount', 'NodeRestriction', 'PersistentVolumeLabel', 'DefaultStorageClass', 'ResourceQuota', 'DefaultTolerationSeconds'] %}
{% if salt.caasp_pillar.get('addons:psp', True) %}
  {% do kube_admission_control.append('PodSecurityPolicy') %}
{% endif %}
KUBE_ADMISSION_CONTROL="{{ kube_admission_control|join('.') }}"
```

MaximilianMeister: i was just pushing an update when you comemnted, tried your suggestion afterwards on https://github.com/qn7o/jinja2-live-parser but it didn't work, i have it slightly different

PR #517, opened 2018-05-07
[do not merge] test psp opt-out feature

MaximilianMeister:
test#ci

### DIFF
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -67,7 +67,7 @@ kube_log_level:   '2'
 
 # install the addons (ie, DNS)
 addons:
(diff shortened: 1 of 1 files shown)


ereslibre:
I think this problem is related to `PodSecurityPolicy` being passed unconditionally here: https://github.com/kubic-project/salt/blob/2674ee5e0509878d220e78b4e77f8a685806216c/salt/kube-apiserver/apiserver.jinja#L26.

If we disable `psp` we should not include `PodSecurityPolicy` in the `KUBE_ADMISSION_CONTROL` envvar.

PR #523, opened 2018-05-09
[3.0] Make HAProxy work as an http proxy instead of a tcp proxy.

ereslibre:
This allows us to add fine-grained timeouts depending on the endpoint being
accessed or with what parameters (e.g. /log?follow=true should have no
timeout as happens on the apiserver). /exec is another example, but in this
case the protocol is upgraded to spdy.

Fixes: bsc#1071994
(cherry picked from commit 442a76cad214f2308f6a1de0ddef8febca8074c8)

### DIFF
--- a/salt/_macros/kubectl.jinja
+++ b/salt/_macros/kubectl.jinja
@@ -11,7 +11,7 @@
 
 {% macro _kubectl_run(args) -%}
   caasp_cmd.run:
-    - name: kubectl --kubeconfig={{ pillar['paths']['kubeconfig'] }} {{ args }}
+    - name: kubectl --request-timeout=1m --kubeconfig={{ pillar['paths']['kubeconfig'] }} {{ args }}
     - retry:
         attempts: 10
         interval: 1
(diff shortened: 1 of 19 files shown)


ereslibre:
[Bug 1071994 - kubectl run --attach exits early](https://trello.com/c/nNNdxJAn/704-bug-1071994-kubectl-run-attach-exits-early)

ereslibre:
Please, do not merge: figuring out why the two PR's together aren't passing in Jenkins: https://github.com/kubic-project/jenkins-library/pull/170

salt/_macros/kubectl.jinja
```diff
@@ -91,10 +91,10 @@
 wait-for-{{ deployment }}-deployment:
```
inercia: Not for this PR, but at some moment we should refactor this and create a `wait_for_deployment` state in Python.

ereslibre:
@flavio @kiall: We need to force merge this one. Passing on https://github.com/kubic-project/jenkins-library/pull/174

PR #526, opened 2018-05-10
Add missing cri-o removal states

flavio:
This is required to fix node removal on clusters using CRI-O as CRI.

Fixes bsc#1092614


flavio:
Merging, the CI failures are not related with the changes.

PR #529, opened 2018-05-14
introduce haproxy errors (bsc#1080636)

MaximilianMeister:
when the velum database is initializing, show a static 503 page

## depends on

* https://github.com/kubic-project/caasp-container-manifests/pull/188
* https://github.com/kubic-project/velum-branding/pull/6

### DIFF
--- a/salt/haproxy/haproxy.cfg.jinja
+++ b/salt/haproxy/haproxy.cfg.jinja
@@ -86,6 +86,8 @@ listen velum
         http-request set-header X-Forwarded-Proto https
         redirect scheme https code 302 if !{ ssl_fc } !path_autoyast
         server velum unix@/var/run/puma/dashboard.sock
+        errorfile 503 /etc/caasp/haproxy/errors/503.html.http
+
 
 listen velum-api
         bind 127.0.0.1:444 ssl crt {{ pillar['ssl']['velum_bundle'] }} ca-file /etc/pki/ca.crt
--- a/salt/haproxy/haproxy.yaml.jinja
+++ b/salt/haproxy/haproxy.yaml.jinja
@@ -43,6 +43,8 @@ spec:
           readOnly: True
         - name: velum-unix-socket
           mountPath: /var/run/puma
+        - name: velum-static-pages
+          mountPath: /etc/caasp/haproxy/errors
 {% endif %}
   volumes:
     - name: haproxy-cfg
@@ -70,4 +72,7 @@ spec:
     - name: velum-unix-socket
       hostPath:
         path: /var/run/puma
+    - name: velum-static-pages
+      hostPath:
+        path: /usr/share/velum/static-pages
 {% endif %}


salt/haproxy/errors/503-velum_database_not_ready.http
```diff
@@ -0,0 +1,73 @@
...
+
+<body>
+  <div class="dialog">
+    <div>
+      <h1>SUSE CaaS Platform is initializing</h1>
+    </div>
+    <p>If you just finished the installation, it may take some seconds for the database to be available. Should the error persist check the logs for more information.</p>
+  </div>
+</body>
+</html>
```
ereslibre: We cannot create this file with salt. It needs to be installed with some package, otherwise on the very first run of the admin node this file won't exist.

salt/haproxy/haproxy.cfg.jinja
```diff
@@ -31,6 +31,8 @@ listen kubernetes-master
 {% endfor -%}
 
 {%- if "admin" in salt['grains.get']('roles', []) %}
+errorfile 503 /etc/caasp/haproxy/errors/503-velum_database_not_ready.http
```
ereslibre: We also need to change caasp-container-manifests for this to be loaded on the very first run of the admin node.

salt/haproxy/errors/503-velum_database_not_ready.http
```diff
@@ -0,0 +1,73 @@
...
+      color: #333333;
+      box-shadow: 0 3px 8px rgba(50, 50, 50, 0.17);
+    }
+  </style>
+</head>
+
+<body>
+  <div class="dialog">
+    <div>
+      <h1>SUSE CaaS Platform is initializing</h1>
```
ereslibre: If the rule on haproxy is to show this warning in the case of a 503 error we cannot be sure that it's always initializing. It could happen that the backend is down for some reason, and in that case we don't want to show that the platform is initializing. This message should be much more generic, something like "SUSE CaaS Platform is not ready, please wait" or something along those lines.

salt/haproxy/errors/503-velum_database_not_ready.http
MaximilianMeister: i wanted to use the one we have in `velum/public` but that doesnt contain the headers, but maybe it works as well?

MaximilianMeister: @ereslibre we could simply add the header in https://github.com/kubic-project/velum/blob/master/public/503_database_not_ready.html but then we also need to mount the velum source directory into the salt-master container. do you think that would be worth it?

flavio:
ping @MaximilianMeister: it would be nice to have that inside of RC1

MaximilianMeister:
> it would be nice to have that inside of RC1

@flavio i've created a branding package now, first of all to have kubic and caasp branded. once this is in the product i'll come back to this PR and fetch the static error page for haproxy from there.

salt/haproxy/haproxy.cfg.jinja
```diff
@@ -59,6 +59,8 @@ backend no-timeout-backend
 
 
 {% if "admin" in salt['grains.get']('roles', []) %}
+errorfile 503 /etc/caasp/haproxy/errors/503_database_not_ready.html
```
MaximilianMeister: @ereslibre it might be that the file needs to end with `.http` I am not really sure there

lcavajani: @MaximilianMeister  @ereslibre The doc says It is recommended to follow the common practice of appending ".http" to the filename so that people do not confuse the response with HTML error pages so it should work with .html

https://cbonte.github.io/haproxy-dconv/1.6/configuration.html#4-errorfile

ereslibre: It looks like we need to set `.http` as the response, but also include the headers of the response. What's going to be returned here is the full contents of this file as the HTTP response on the TCP socket, so we need something like:

```
HTTP/1.0 503 Service Unavailable
Cache-Control: no-cache
Connection: close
Content-Type: text/html

<contents here>
```

MaximilianMeister: i've submitted https://github.com/kubic-project/velum-branding/pull/6 to fix this. also from reading the docs, i had to minify the html files, otherwise haproxy would truncate them after reaching 16kb :laughing:

MaximilianMeister:
> Same comment as in kubic-project/caasp-container-manifests#188 (review)

@ereslibre updated

MaximilianMeister:
@ereslibre i still missed the haproxy manifest managed by salt ... please have another look

MaximilianMeister:
> LGTM, can we get this merged?

i'll rebase it once more to make sure it still applies. last CI run has been a while

MalloZup:
@MaximilianMeister  can we close this?

MalloZup:
closing as outdated feel free to reopen

PR #535, opened 2018-05-15
Queue the /etc/hosts update when triggered from a reactor.

inercia:
It seems that the `update-etc-hosts` orchestration can conflict with... itself. Apparently this can happen when a new orchestration has been started and the previous one has not finished yet. This tried to solve that problem by queueing the only problematic part of the orchestration, the `salt.state` clause.

Fixes part of bsc#1093123


flavio:
@inercia please backport that to the 3.0 branch

PR #537, opened 2018-05-16
also reload docker when certificates change

MaximilianMeister:
fix#reload-certs

### DIFF
--- a/salt/docker/init.sls
+++ b/salt/docker/init.sls
@@ -98,3 +98,7 @@ docker-reload-config:
     - reload: True
     - onchanges:
       - file: /etc/docker/daemon.json
+  {% for cert_tuple in certs.items() %}
+    {% set name, _cert = cert_tuple %}
+      - file: /etc/docker/certs.d/{{ name }}/ca.crt
+  {% endfor %}


ellisab:
Function: cmd.run\n            Name: systemctl daemon-reload && systemctl restart docker.service\n          Result: True\n         Comment: State was not run because none of the onchanges reqs changed\n         Changes:   \n    ----------\n              ID: docker-reload-config-and-certs\n        Function: service.running\n            Name: docker\n          Result: False\n         Comment: The following requisites were not found:\n                                     onchanges:\n                                         file: /etc/docker/certs.d/*\n         Changes:   \n    ----------\n              ID: /etc/crictl.yaml\n

inercia:
Is this necessary? As far as I can remember the Docker daemon doesn't need to be reloaded after changing certificates, as they are loaded on demand...

MaximilianMeister:
> Is this necessary? As far as I can remember the Docker daemon doesn't need to be reloaded after changing certificates, as they are loaded on demand...

@ellisab can you please confirm again that this is necessary?

MaximilianMeister:
> Function: cmd.run\n Name: systemctl daemon-reload && systemctl restart docker.service\n Result: True\n Comment: State was not run because none of the onchanges reqs changed\n Changes: \n ----------\n ID: docker-reload-config-and-certs\n Function: service.running\n Name: docker\n Result: False\n Comment: The following requisites were not found:\n onchanges:\n file: /etc/docker/certs.d/*\n Changes: \n ----------\n ID: /etc/crictl.yaml\n

i think the glob doesnt work with the watcher, i'll try to make it differently

MaximilianMeister:
> i think the glob doesnt work with the watcher, i'll try to make it differently

i've done it now a bit differently, by watching each cert explicitly. however it could also be done with a `file.recurse` definition, and then watching on that resource id...

salt/docker/init.sls
```diff
@@ -24,6 +24,10 @@ include:
         {{ cert | indent(8) }}
     - require_in:
       - docker
+  module.run:
+    - name: service.systemctl_reload
+    - onchanges:
+      - file: /etc/docker/certs.d/{{ name }}/ca.crt
```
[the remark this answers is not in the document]
ellisab: Inercia it does not work without a reload, that's why this pull request was started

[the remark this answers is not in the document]
MaximilianMeister: what about

```salt
certs-d-files:
  file.recurse:
    - name: /etc/docker/certs.d
    - file_mode: 655
    - user: root
    - group: root
    - require:
      - pkg: docker

docker-reload-config:
  service.running:
    - name: docker
    - reload: True
    - onchanges:
      - file: /etc/docker/daemon.json
      - file: certs-d-files
```

MaximilianMeister:
> View changes 
> According to my own comment here, we do not need to restart docker after adding/removing certificates.

it says `You do not need to restart Docker.` nothing regarding reload

ellisab:
Bootstrapping fails with:

Function: file.recurse\n            Name: /etc/docker/certs.d\n          Result: False\n         Comment: Missing parameter source for state file.recurse\n         Changes:   \n    ----------\n              ID: docker-reload-config-and-certs\n        Function: service.running\n            Name: docker\n          Result: False\n         Comment: One or more requisite failed: docker.certs-d-files\n         Changes:   \n    ----------\n              ID: /etc/crictl.yaml\n

MaximilianMeister:
> Bootstrapping fails with:
> 
> Function: file.recurse\n Name: /etc/docker/certs.d\n Result: False\n Comment: Missing parameter source for state file.recurse\n Changes: \n ----------\n ID: docker-reload-config-and-certs\n Function: service.running\n Name: docker\n Result: False\n Comment: One or more requisite failed: docker.certs-d-files\n Changes: \n ----------\n ID: /etc/crictl.yaml\n

ok, this works only if the certs folder is managed by salt :angry: 

@inercia do you have any better suggestion besides the explicit reload on every cert like it was before?

MaximilianMeister:
> This would add a onchanges: X in docker-reload-config for each certificate X we are adding, so I think this would effectively reload docker (only once) at the end of the certificates generation.

i think that's not enough, because the user can update only a certificate, and so the `daemon.json` won't be touched/changed, which means that the `docker-reload-config` is not a reliable trigger ... :thinking:

inercia:
> i think that's not enough, because the user can update only a certificate, and so the daemon.json won't be touched/changed, which means that the docker-reload-config is not a reliable trigger ...


This would not only depend on `daemon.json` but on all the certificates, as:

```SaltStack
X:
  - something:
    - onchanges_in:
      - docker-reload-config
```

is equivalent to:

```SaltStack
docker-reload-config
  - something:
    - onchanges:
      - X
```

So you would be effectively adding `onchanges` to the `docker-reload-config` for all the certificates added by users.

ellisab:
Still got:
Failed to pull image "docker-testing-registry.suse.de:443/busybox": rpc error: code = Unknown desc = Error response from daemon: Get https://docker-testing-registry.suse.de:443/v2/: x509: certificate signed by unknown authority

Solved by reloading docker service manually on all nodes:
May 17 10:12:04 caasp-master0 systemd[1]: Reloading Docker Application Container Engine.
May 17 10:12:04 caasp-master0 systemd[1]: Reloaded Docker Application Container Engine.

MaximilianMeister:
> Still got:
> Failed to pull image "docker-testing-registry.suse.de:443/busybox": rpc error: code = Unknown desc = Error response from daemon: Get https://docker-testing-registry.suse.de:443/v2/: x509: certificate signed by unknown authority
> 
> Solved by reloading docker service manually on all nodes:
> May 17 10:12:04 caasp-master0 systemd[1]: Reloading Docker Application Container Engine.
> May 17 10:12:04 caasp-master0 systemd[1]: Reloaded Docker Application Container Engine.
> 
> Other then my manual reload i don't see a salt reload in the log

@inercia after all the logic you explained doesn't seem to work. as i was suspecting i think that this will only work when you update `daemon.json` otherwise this won't trigger ...

salt/docker/init.sls
```diff
@@ -24,6 +24,8 @@ include:
         {{ cert | indent(8) }}
     - require_in:
       - docker
+    - onchanges_in:
+      - docker-reload-config
```
inercia: Where is `docker-reload-config` defined?

MaximilianMeister: in the same file at the bottom

salt/docker/init.sls
```diff
@@ -24,6 +24,10 @@ include:
         {{ cert | indent(8) }}
     - require_in:
       - docker
+    module.run:
+      - name: service.systemctl_reload
+      - onchanges:
+        - file: /etc/docker/certs.d/{{ name }}/ca.crt
```
inercia: This will run a `systemctl reload` for each cert. Not a good idea...

salt/docker/init.sls
inercia: Ok, found it. So if my previous sugestion doesn't work for you, you could try the equivalent: add the `onchanges` explicitly in `docker-reload-config`:

```SaltStack
docker-reload-config:
  service.running:
    - name: docker
    - reload: True
    - onchanges:
      - file: /etc/docker/daemon.json
 {% for cert_tuple in certs.items() %}
   {% set name, cert = cert_tuple %}
      - file: /etc/docker/certs.d/{{ name }}/ca.crt
 {% endfor %}
```

If this doesn't work, your Salt master is broken ;-)

salt/docker/init.sls
MaximilianMeister: > This will run a systemctl reload for each cert. Not a good idea...

we know that already from your previous comment, i just changed it to something that works until we had your feedback

salt/docker/init.sls
MaximilianMeister: that should work, thanks. will update it

salt/docker/init.sls
```diff
@@ -98,3 +98,7 @@ docker-reload-config:
     - reload: True
     - onchanges:
       - file: /etc/docker/daemon.json
+  {% for cert_tuple in certs.items() %}
+    {% set name, _cert = cert_tuple %}
+      - file: /etc/docker/certs.d/{{ name }}/ca.crt
```
MaximilianMeister: @inercia btw i highly suspect that this is also reloading on each cert change...

inercia: Please make sure if that is the case and we will report it upstream to the Salt developers.

MaximilianMeister: i'll have a look upstream. can you give your last review on this one?

flavio:
I agree with @inercia, please go ahead and merge it

ellisab:
@inercia still does not work.

Works only afterrunning manually systemctl reload docker.service

PR #538, opened 2018-05-16
Only remove the master grains if there are any masters to be updated.

ereslibre:
The `salt.function` call will be marked as failed if there were no minions
to target. Make sure that we only run this step if we know that we'll have
some targets available.

Fixes: bsc#1093491

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -28,12 +28,14 @@
 {%- set is_worker_tgt            = is_responsive_node_tgt + ' and G@roles:kube-minion' %}
 {%- set is_updateable_master_tgt = is_updateable_tgt + ' and ' + is_master_tgt %}
 {%- set is_updateable_worker_tgt = is_updateable_tgt + ' and ' + is_worker_tgt %}
+{%- set all_masters = salt.saltutil.runner('mine.get', tgt=is_master_tgt, fun='network.interfaces', tgt_type='compound').keys() %}
+{%- set super_master = all_masters|first %}
 
 # Ensure all nodes with updates are marked as upgrading. This will reduce the time window in which
 # the update-etc-hosts orchestration can run in between machine restarts.
 set-update-grain:
   salt.function:
-    - tgt: '{{ is_regular_node_tgt }} and {{ is_updateable_tgt }}'
+    - tgt: '( {{ is_regular_node_tgt }} and {{ is_updateable_tgt }} ) or {{ super_master }}'
     - tgt_type: compound
     - name: grains.setval
     - arg:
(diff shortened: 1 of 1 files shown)


ereslibre:
[Bug 1093491 - When updating the cluster, if no masters are for update, salt fails on `remove-cluster-wide-removal-grains` step](https://trello.com/c/PQaQgWZm/893-bug-1093491-when-updating-the-cluster-if-no-masters-are-for-update-salt-fails-on-remove-cluster-wide-removal-grains-step)

salt/orch/update.sls
```diff
@@ -434,6 +434,7 @@ remove-caasp-fqdn-grain:
     - require:
       - admin-wait-for-services
 
+{%- if salt.saltutil.runner('mine.get', tgt=is_master_tgt + ' and G@update_in_progress:true', fun='network.interfaces', tgt_type='compound')|length > 0 %}
```
inercia: It would be better to use the `is_updateable_master_tgt` variable previously defined, so

```Jinja
{%- if salt.saltutil.runner('mine.get', tgt=is_updateable_master_tgt, fun='network.interfaces', tgt_type='compound')|length > 0 %}
```

I would even add a `updateable_master` masters affter the `all_masters` definition.

```Jinja
{%- set updateable_masters = salt.saltutil.runner('mine.get', tgt=is_updateablee_master_tgt, fun='network.interfaces', tgt_type='compound').keys() %}
...
{%- if updateable_masters|length > 0 %}
...
{% endif %}
```

inercia: I've just realized the `masters` variable holds that info. This would be enough:

```Jinja
{%- if masters|length > 0 %}
...
{% endif %}
```

ereslibre: No, it's not the same. `is_updateable_master_tgt` has `G@tx_update_reboot_needed:true and G@roles:kube-master` whereas we want `G@update_in_progress:true and G@roles:kube-master`.

ereslibre: We cannot use `masters` because that list is fetched using `is_updateable_master_tgt` that doesn't really match what we want on this step.

inercia: You are right, but then I think we will have a proble, because the Jinja code is evaluated once at beggining, so I think you cannot use a grain that is set in the orchestration.

inercia: Intelligently solved with a grain target. I think we are done.

flavio:
@ereslibre please backport that into the 3.0 branch

salt/orch/update.sls
```diff
@@ -434,10 +433,10 @@ remove-caasp-fqdn-grain:
     - require:
       - admin-wait-for-services
 
-masters-remove-update-grain:
+remove-update-grain:
```
MaximilianMeister: is this similar to what we do in the `{{ worker_id }}-remove-update-grain` sls? now that it's not tied to `is_master_tgt` anymore

PR #539, opened 2018-05-16
[3.0] Only remove the master grains if there are any masters to be updated.

ereslibre:
The `salt.function` call will be marked as failed if there were no minions
to target. Make sure that we only run this step if we know that we'll have
some targets available.

Fixes: bsc#1093491
(cherry picked from commit 28fccad520948a149ff0c0fb409b0981c6e1bcd5)


ereslibre:
[Bug 1093491 - When updating the cluster, if no masters are for update, salt fails on `remove-cluster-wide-removal-grains` step](https://trello.com/c/PQaQgWZm/893-bug-1093491-when-updating-the-cluster-if-no-masters-are-for-update-salt-fails-on-remove-cluster-wide-removal-grains-step)

flavio:
I don't think the error on jenkins is related. @ereslibre I can forcefully merge it if you want

ereslibre:
@flavio I'll retry now that the longer wait is merged (https://github.com/kubic-project/automation/pull/389)

PR #541, opened 2018-05-16
Add support for kube API auditing

flavio:
Allow users to enable kubernetes API server auditing feature.

The auditing will produce an audit log file locally that can then be pushed to a central logging solution (eg: by using a fluentd daemonset running on the master nodes).

By default there's no auditing in place. This is enabled only when the user provides a value for each one of the new pillars introduced by this commit.

feature#kube-api-audit
fate#325337

For more information: https://v1-9.docs.kubernetes.io/docs/tasks/debug-application-cluster/audit/#log-backend

### DIFF
--- a/salt/kube-apiserver/apiserver.jinja
+++ b/salt/kube-apiserver/apiserver.jinja
@@ -6,7 +6,6 @@
 
 {% set cloud_provider = salt.caasp_pillar.get('cloud:provider') %}
 {% set kube_admission_control = ['Initializers', 'NamespaceLifecycle', 'LimitRanger', 'ServiceAccount', 'NodeRestriction', 'PersistentVolumeLabel', 'DefaultStorageClass', 'ResourceQuota', 'DefaultTolerationSeconds'] %}
-
 # The address on the local server to listen to.
 KUBE_API_ADDRESS="--insecure-bind-address=127.0.0.1 --bind-address=0.0.0.0"
 
@@ -39,6 +38,14 @@ KUBE_API_ARGS="--advertise-address={{ salt.caasp_net.get_primary_ip() }} \
                --cloud-config=/etc/kubernetes/openstack-config \
   {%- endif -%}
 {% endif %}
+{#- An empty audit policy file would cause kubernetes apiserver to not start at all -#}
+{%- if salt.caasp_pillar.get('api:audit:log:enabled', False) and salt.caasp_pillar.get('api:audit:log:policy') %}
+               --audit-policy-file=/etc/kubernetes/audit-policy.yaml \
+               --audit-log-path=/var/log/kube-apiserver/audit.log \
+               --audit-log-maxage={{ salt.caasp_pillar.get('api:audit:log:maxage') }} \
+               --audit-log-maxsize={{ salt.caasp_pillar.get('api:audit:log:maxsize') }} \
+               --audit-log-maxbackup={{ salt.caasp_pillar.get('api:audit:log:maxbackup') }} \
+{%- endif -%}
                --tls-cert-file={{ pillar['ssl']['kube_apiserver_crt'] }} \
                --tls-private-key-file={{ pillar['ssl']['kube_apiserver_key'] }} \
                --tls-ca-file={{ pillar['ssl']['ca_file'] }} \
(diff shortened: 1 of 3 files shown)


salt/kube-apiserver/apiserver.jinja
```diff
@@ -6,7 +6,7 @@
 
 {% set cloud_provider = salt.caasp_pillar.get('cloud:provider') %}
 {% set kube_admission_control = ['Initializers', 'NamespaceLifecycle', 'LimitRanger', 'ServiceAccount', 'NodeRestriction', 'PersistentVolumeLabel', 'DefaultStorageClass', 'ResourceQuota', 'DefaultTolerationSeconds'] %}
-
+{% set audit_log = salt.caasp_pillar.get('api:audit:log:maxage') and salt.caasp_pillar.get('api:audit:log:maxsize') and salt.caasp_pillar.get('api:audit:log:maxbackup') and salt.caasp_pillar.get('api:audit:log:policy') %}
```
inercia: Wouldn't it be better to add a `api:audit:enabled` flag? It would make things easier and, at the same time, we could provide some safe defaults for these attributes while the audit feature could be disabled by default.

salt/kube-apiserver/init.sls
```diff
@@ -47,13 +47,30 @@ kube-apiserver:
       - sls:             ca-cert
       - caasp_retriable: {{ pillar['ssl']['kube_apiserver_crt'] }}
       - x509:            {{ pillar['paths']['service_account_key'] }}
+      - file:            /etc/kubernetes/audit-policy.yaml
```
inercia: You are forcing the need for this file, while it could be

```
{%- if salt.caasp_pillar.get('api:audit:enabled', False) %}
    - file:            /etc/kubernetes/audit-policy.yaml
{%- endif %}
```

salt/kube-apiserver/init.sls
```diff
@@ -47,13 +47,30 @@ kube-apiserver:
...
+/var/log/kube-apiserver:
+  file.directory:
+    - user: kube
+    - group: kube
+    - dirmode: 755
+    - filemode: 644
+    - require:
+      - pkg: kube-apiserver
+
+/etc/kubernetes/audit-policy.yaml:
```
inercia: Same here. I would skip this file if `api:audit:enable` is not `True`.

salt/kube-apiserver/apiserver.jinja
flavio: I will change that

salt/kube-apiserver/init.sls
flavio: There's no harm in having this file around, even if the feature is disabled. I would prefer to leave the salt states as slim as possible.

salt/kube-apiserver/init.sls
flavio: Same here.

flavio:
I updated the PR in the way @inercia recommended. Now it has also some sane defaults. I would not provide a default policy because that's a bit of a can of worms...

bergmannf:
I bootstrapped a cluster with the changes and when I fill out all the field it works fine.

Only problem would be if the the policy file is not filled in correctly (with a base64 encoded policy).

For an *empty* policy file the `apiserver` can't start:

```
May 17 12:06:32 master-0 hyperkube[11271]: Error: loading audit policy file: unknown group version field audit.k8s.io/__internal, Kind=Policy in policy file /etc/kubernetes/audit-policy.yaml
```

For a *non-unicode* policy file (just a long string with `\n`) it can't decode (obviously) in salt:

```
UnicodeDecodeError: 'ascii' codec can't decode byte 0x88 in position 1: ordinal not in range(128)
```

So maybe it would make sense to put the requirement of `base64` encoding in a comment in the `kube-master.sls` file, so this is documented at the location where changes should be made.

inercia:
> So maybe it would make sense to put the requirement of base64 encoding in a comment in the kube-master.sls file, so this is documented at the location where changes should be made.

@flavio where is the policy comming from? is it going to be filled in Velum? is Velum checking the policy is properly encoded as base64? can we skip the base64 thing at all?

flavio:
@ereslibre I went with a base64 encoded policy to avoid issues inside of Velum's database and with the "transmission" of the pillar value. Maybe we can skip that and store the policy in simple text format?

ereslibre:
Yes, @flavio, I'll do that.

flavio:
@ereslibre, great I'll take care of changing this PR to avoid usage of base64

PR #542, opened 2018-05-16
Lower the per-request timeout when we are checking for successful query

ereslibre:
When we are waiting for some service to be up, if the request hangs for
some reason, we want to retry at least several times. Without setting
this value explicitly, it takes the default (`http_request_timeout` as 3600),
what is way over our `wait_for` argument set at 300 seconds.

By setting the default `http_request_timeout` to a more reasonable default
when doing this kind of checks we can ensure that the request itself will
timeout several times before we call it done.

Fixes: bsc#1093540
Fixes: bsc#1093685

### DIFF
--- a/salt/_modules/caasp_http.py
+++ b/salt/_modules/caasp_http.py
@@ -1,0 +1,77 @@
+# -*- coding: utf-8 -*-
+'''
+Module for making various web calls. Primarily designed for webhooks and the
+like, but also useful for basic http testing.
+
+CaaS Platform specific:
+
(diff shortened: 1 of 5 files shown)


ereslibre:
[Bug 1093540 - Timeout can be reached during deployment when waiting for successful query](https://trello.com/c/rI6a4bYQ/895-bug-1093540-timeout-can-be-reached-during-deployment-when-waiting-for-successful-query)

ereslibre:
Ready!

PR #551, opened 2018-05-18
[WIP] Allow salt tests to be ran via tox and Jenkins

kiall:
Example to run them locally:

    tox -e tests-salt-2016.11.4-py27

or:

    tox -e tests-salt-2016.11.4-py34


kiall:
This won't pass, as the tests (legitimately) fail on Python 3.4 - a separate issue to fix. Merging.

PR #555, opened 2018-05-21
Enable the ValidatingAdmissionWebhook

flavio:
Enable by default this admission controller. Users can later define
custom validators leveraging external webhooks.

The validators definition happens via kubernetes native resources, hence
without any intervention from our side.

Note well: this commit introduces a new set of certificate. This pair of
certificate is going to be used by the API server to perform Mutual TLS
authentication against the external webhook.

In this scenario the API server is the client, while the server is the
webhook service specified by the user.

feature#ValidatingAdmissionWebhook

This is based on the research I did sometimes ago on image admission controllers: https://github.com/flavio/kube-image-bouncer

cc @ajaeger: this is going to make some of our customers happy, the ones who requested the possibility to use this generic admission controller.

### DIFF
--- a/salt/kube-apiserver/apiserver.jinja
+++ b/salt/kube-apiserver/apiserver.jinja
@@ -61,8 +61,10 @@ KUBE_API_ARGS="--advertise-address={{ salt.caasp_net.get_primary_ip() }} \
                --tls-cert-file={{ pillar['ssl']['kube_apiserver_crt'] }} \
                --tls-private-key-file={{ pillar['ssl']['kube_apiserver_key'] }} \
                --tls-ca-file={{ pillar['ssl']['ca_file'] }} \
-               --proxy-client-cert-file={{ pillar['ssl']['kube_apiserver_proxy_client_crt'] }} \
-               --proxy-client-key-file={{ pillar['ssl']['kube_apiserver_proxy_client_key'] }} \
+{%- if salt.caasp_pillar.get('api:admission_webhook:enabled', False) %}
+               --proxy-client-cert-file=/etc/pki/kube-apiserver-proxy-client.crt \
+               --proxy-client-key-file=/etc/pki/kube-apiserver-proxy-client.key \
+{%- endif -%}
                --cert-dir=/etc/pki \
                {{ salt.caasp_pillar.get('components:apiserver:args') }} \
                --requestheader-username-headers=X-Remote-User \
--- a/salt/kube-apiserver/apiserver.jinja
+++ b/salt/kube-apiserver/apiserver.jinja
@@ -5,7 +5,19 @@
 #
 
 {% set cloud_provider = salt.caasp_pillar.get('cloud:provider') %}
-{% set kube_admission_control = ['Initializers', 'NamespaceLifecycle', 'LimitRanger', 'ServiceAccount', 'NodeRestriction', 'PersistentVolumeLabel', 'DefaultStorageClass', 'ResourceQuota', 'DefaultTolerationSeconds'] %}
+{# Note well: order matters #}
+{% set kube_admission_control = [
+  'Initializers',
+  'NamespaceLifecycle',
+  'LimitRanger',
+  'ServiceAccount',
+  'NodeRestriction',
+  'PersistentVolumeLabel',
+  'DefaultStorageClass',
+  'DefaultTolerationSeconds',
+  'ValidatingAdmissionWebhook',
+  'ResourceQuota'] %}
+
 # The address on the local server to listen to.
 KUBE_API_ADDRESS="--insecure-bind-address=127.0.0.1 --bind-address=0.0.0.0"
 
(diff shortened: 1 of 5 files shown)


flavio:
I'm currently putting together some instructions both for QA and for a blog post.

kiall:
Should we really be using our CA to generate this client cert? That would require we allow operators to issue server certs from our CA - which is something I really don't think is a good idea to encourage. IMO, the operator should be supplying us the client key and certificate to use for this.

Our internal CA should never IMO be used outside of the cluster - and we already use it too much, as the public endpoints for Velum / Kube use it when they shouldn't!

flavio:
When using this admission controller the api-server (which is where the admission controller code runs) will make a https request against the external webhook endpoint provided by the user.

The external webhook must use a certificate and the user has to provide that when creating the `ValidationWebhookConfiguration` (see for example: [this](https://github.com/flavio/kube-image-bouncer/blob/master/kubernetes/validating-webhook-configuration.yaml) configuration).

The api-server will perform a mutual TLS authentication against the external webhook endpoint. As usual there are two actors during this phase:
  * server: this is the external service operated by the user. The one using the certificate generated by the user and specified into the `ValidationWebhookConfiguration`
  * client: this is the api server. It must be using a dedicated certificate. This is the one generated by this PR.

From the api-server documentation about `--proxy-client-cert-file`:

> Client certificate used to prove the identity of the aggregator or kube-apiserver when it must call out during a request. This includes proxying requests to a user api-server and calling out to webhook admission plugins. It is expected that this cert includes a signature from the CA in the --requestheader-client-ca-file flag. That CA is published in the 'extension-apiserver-authentication' configmap in the kube-system namespace. Components recieving calls from kube-aggregator should use that CA to perform their half of the mutual TLS verification.

So the certificate we generate is relevant to the user only if the webhook endpoint is doing a full mutual TLS verification.

I think that for most customers having us generate this certificate would be enough, for the others we should definitely allow them to provide a custom certificate. I would do that with another PR, probably as part of https://trello.com/c/nNv85jtT/808-p1-bug-1090067-providing-self-signed-ca-for-caasp

salt/kube-apiserver/apiserver.jinja
```diff
@@ -49,6 +61,8 @@ KUBE_API_ARGS="--advertise-address={{ salt.caasp_net.get_primary_ip() }} \
                --tls-cert-file={{ pillar['ssl']['kube_apiserver_crt'] }} \
                --tls-private-key-file={{ pillar['ssl']['kube_apiserver_key'] }} \
                --tls-ca-file={{ pillar['ssl']['ca_file'] }} \
+               --proxy-client-cert-file={{ pillar['ssl']['kube_apiserver_proxy_client_crt'] }} \
```
inercia: Shouldn't we enable/disable this feature with a pillar?

flavio: It can be turned on all the time. It's being used only if the user defines a new validation service by creating the right kubernetes objects via `kubectl`

ereslibre:
Based on the information provided by @flavio (I didn't look further into external admission controls), I'd say:

- We should ask the user to provide:
    - Client certificate key
    - Client certificate

If we can avoid having to set `requestheader-client-ca-file`, and ask for the client certificate and key to the user, set those with salt, the connections made from the apiserver to the external admission service will use these certificate, that will prove the external service that the apiserver is the one really requesting access.

This allows the customer to use CRL's on their service side, while they keep the control over their CA (or even they can use external well-known CA's to create the certificates if they wish).

flavio:
I chatted with Rafa about this PR.

We both agree the best workflow would consist of the user uploading a new certificate and key for this specific api-server proxy client feature.

However we will still have users who don't have an own PKI infrastructure or don't want to mess around with certificates (buying certificates to well-known CA's). For those users we should create a certificate and a key like this PR is currently doing.
We don't know if this certificate bundle should be created from our current internal CA or if we should create a new top-level CA to be used only by "external services". That would allow to be more flexible if something goes south and we have to revoke a CA authority.

Moreover when we are the ones generating the CA and key we should have a page on Velum from which the user can download the CA file.

I think the should go by steps:

1. Decide if we want to create a new internal CA just for services.
2. Generate the client certificate bundle automatically
3. Merge the PRs related with ["allow user to provide self-signed CA for caasp"](https://trello.com/c/nNv85jtT/808-p1-bug-1090067-providing-self-signed-ca-for-caasp)
4. Extend Velum UI to allow users to upload their client certificate bundle for api-server proxy client
5. Extend Velum to allow users to download the client proxy certificate bundle and CA (until this is done we have to provide docs about how to get it - a simple "ssh into the admin node, download file /etc/pki/..." should be enough)

salt/kube-apiserver/init.sls
```diff
@@ -12,6 +12,24 @@ include:
...
+    - mode: 644
+    - require:
+      - pkg: kube-apiserver
+
+/etc/pki/kube-apiserver-proxy-client.key:
+  file.managed:
+    - contents_pillar: api:admission_webhook:key
+    - user: kube
+    - group: kube
+    - mode: 644
```
MaximilianMeister: i think the private key should have `444` as we have for all other keys in https://github.com/kubic-project/salt/blob/b00d7641640c0394e1ffdd875728a60d44ba2d41/salt/_macros/certs.jinja#L60

inercia:
Putting this on hold until ready for merge...

flavio:
Please, let's leave everything on hold. I found some issues that have to be sorted out before merge.

salt/kube-apiserver/init.sls
```diff
@@ -12,6 +12,24 @@ include:
          cn = grains['nodename'],
          o = pillar['certificate_information']['subject_properties']['O']) }}
 
+/etc/pki/kube-apiserver-proxy-client.crt:
+  file.managed:
+    - contents_pillar: api:admission_webhook:cert
```
kiall: This cert/keypair is used for more than just the admission webhook. We shouldn't call the pillar this :)

MalloZup:
@flavio i'm closing this. feel free to reopen with up2date code tia :heart:

PR #562, opened 2018-05-22
Fix kubelet arguments

ereslibre:
Remove unsupported `--require-kubeconfig` argument deprecated in Kubernetes (and removed in 1.10)


ereslibre:
[Bug 1094217 - The kubelet on Kubernetes 1.10 refuses to start if `--require-kubeconfig` is provided](https://trello.com/c/ZoKHeYtn/916-bug-1094217-the-kubelet-on-kubernetes-110-refuses-to-start-if-require-kubeconfig-is-provided)

flavio:
Are you sure this is going to work? I tried to fix the warning sometimes ago, but that caused crictl to explode. Maybe we updated to a more recent version of it that handles that new format...

flavio:
Overall the code lgtm

PR #568, opened 2018-05-23
[WIP] fix docker reload again

MaximilianMeister:
it apparently doesnt work to use service.running to
do the reload. using cmd.run is reliable

fix#reload-cert

follow up of https://github.com/kubic-project/salt/pull/537


ellisab:
works fine @MaximilianMeister

PR #513, opened 2018-05-03
explicitly pass unix_socket

MaximilianMeister:
this affects only kubic for now where we use PyMySQL

we cant use the MYSQL_UNIX_PORT workaround anymore as we could
do with MySQLdb

salt#mysql-unix-socket

Signed-off-by: Maximilian Meister <mmeister@suse.de>

PR #572, opened 2018-05-24
50-returner.conf: specfify unix_socket for salt-2018

thkukuk:
With salt-2018 and the new PyMySQL, the MYSQL_UNIX_PORT environment variable will no longer work. But we can specify the UNIX socket directly.


MaximilianMeister:
https://github.com/kubic-project/salt/pull/513 already handles this, but it's set as WIP as we dont have salt-2018 yet in caasp

PR #578, opened 2018-05-29
override volume plugin dir (bsc#1084766)

MaximilianMeister:
kubernetes 1.10 uses /usr/libexec by default which doesnt exist,
and we want to stick with /usr/lib

kubelet config fix in the package is part of https://build.opensuse.org/request/show/612897


MaximilianMeister:
i think we should wait for https://build.opensuse.org/request/show/612897 to be in before merging this. to have the correct config from the start (before bootstrap)

MaximilianMeister:
new k8s is in devel now. so we can merge this

PR #583, opened 2018-06-04
[3.0] fix get node targets

inercia:
Backport of #581.

* Fix targets when calculating a replacement. For example, when getting a replacement for a `kube-master`, we were looking for `G@roles:etcd`, when we should look for `G@roles:etcd and not G@roles:kube-master`.
* Do not include removed nodes when generating `haproxy.cfg`
* Refactorings in the `get_with_expr()` code
* Do not use unassigned nodes when looking for a replacement.
* Minor cleanups

See https://trello.com/c/eKXnJbzz and https://trello.com/c/mKb0nykK

### DIFF
--- a/salt/_modules/caasp_etcd.py
+++ b/salt/_modules/caasp_etcd.py
@@ -137,6 +137,7 @@ def get_additional_etcd_members(num_wanted=None, **kwargs):
     #
     new_etcd_members = __salt__['caasp_nodes.get_with_prio_for_role'](
         num_additional_etcd_members, 'etcd',
+        unassigned=False,
         excluded=current_etcd_members + excluded)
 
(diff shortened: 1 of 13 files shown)


ereslibre:
[Bug 1095330 – haproxy configuration is not updpated after a node removal](https://trello.com/c/eKXnJbzz/989-bug-1095330-haproxy-configuration-is-not-updpated-after-a-node-removal)

ereslibre:
[Bug 1095336 – node removal for etcd can choose an unsuitable replacement](https://trello.com/c/mKb0nykK/990-bug-1095336-node-removal-for-etcd-can-choose-an-unsuitable-replacement)

PR #584, opened 2018-06-05
Uncordon node in a explicit sls action

ereslibre:
This way we don't try to uncordon the node in the `kubelet/init.sls` file,
required for example by `haproxy`, that will end up in the machine trying
to early uncordon itself (when `haproxy` configuration hasn't been written
yet, and leading to early failure).

Splitting this action and called only when required (this is: the update
process) is safer.

Fixes: bsc#1080978

PR #585, opened 2018-06-06
[3.0] Uncordon node in a explicit sls action

ereslibre:
This way we don't try to uncordon the node in the `kubelet/init.sls` file,
required for example by `haproxy`, that will end up in the machine trying
to early uncordon itself (when `haproxy` configuration hasn't been written
yet, and leading to early failure).

Splitting this action and called only when required (this is: the update
process) is safer.

Fixes: bsc#1080978

(cherry picked from commit 02f063385e3a8cd435a76280ca246a87099a01d5)

Backport of https://github.com/kubic-project/salt/pull/584

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -100,7 +100,7 @@ admin-setup:
     - require:
       - admin-apply-haproxy
 
-# Perform any migrations necessary before starting the update orchestration. All services and
+# Perform any necessary migrations before starting the update orchestration. All services and
 # machines should be running and we can migrate some data on the whole cluster and then proceed
 # with the real update.
 pre-orchestration-migration:
@@ -159,7 +159,7 @@ etcd-setup:
     - require:
         - etcd-setup
 
-# Perform any migratrions necessary before services are shutdown
+# Perform any necessary migrations before services are shutdown
 {{ master_id }}-pre-reboot:
   salt.state:
     - tgt: '{{ master_id }}'
@@ -192,7 +192,7 @@ etcd-setup:
     - require:
       - {{ master_id }}-reboot
 
-# Perform any migratrions necessary before salt starts doing
+# Perform any necessary migrations before salt starts doing
 # "real work" again
 {{ master_id }}-post-reboot:
   salt.state:
@@ -219,16 +219,6 @@ etcd-setup:
     - require:
       - {{ master_id }}-apply-haproxy
 
-# Perform any migratrions after services are started
-{{ master_id }}-post-start-services:
-  salt.state:
-    - tgt: '{{ master_id }}'
-    - sls:
-      - cni.update-post-start-services
-      - kubelet.update-post-start-services
-    - require:
-      - {{ master_id }}-start-services
-
 {{ master_id }}-reboot-needed-grain:
   salt.function:
     - tgt: '{{ master_id }}'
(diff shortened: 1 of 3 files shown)


ereslibre:
Upgrade is working.

salt/orch/update.sls
```diff
@@ -321,11 +326,12 @@ etcd-setup:
     - require:
       - salt: {{ worker_id }}-apply-haproxy
 
-# Perform any migratrions after services are started
+# Perform any migrations after services are started
 {{ worker_id }}-update-post-start-services:
   salt.state:
     - tgt: '{{ worker_id }}'
     - sls:
+      - kubelet.uncordon
```
inercia: All the states here are called by their _stage_ (`update-post-start-services`) instead of _what they do_ (to `uncordon`). I would try to follow the convention...

salt/orch/update.sls
```diff
@@ -238,10 +228,25 @@ etcd-setup:
...
+    - tgt_type: compound
+    - batch: 3
+    - sls:
+      - cni.update-post-start-services
+      - kubelet.update-post-start-services
+    - require:
+      - etcd-setup
+{%- for master_id in masters.keys() %}
+      - {{ master_id }}-reboot-needed-grain
+{%- endfor %}
```
inercia: What is the reason for doing the `{cni|kubelet}-post-start-services` in the master in parallel instead of one-by-one (as we have been doing so far)?

ereslibre: In this case, the upgrade from 2.0 to 3.0 includes an [auth change](https://github.com/kubic-project/salt/pull/424/files#diff-7504d75a43d2c611b6f390b9c99a27d6). This means that if we have M1, M2 and M3. We update M1 first, when it reaches the `post-start-services`, it will try to run `kubectl delete node old_node_name`, but that request can land in M2 or M3, that doesn't yet understand the new auth method (didn't get updated yet, we are updating M1), so the removal of the old nodename does not work.

This makes sure we perform this task after all master machines have been upgraded. The workers keep the same logic one by one, as at that point in time all masters have been updated and they all understand the authentication method.

salt/orch/update.sls
ereslibre: Right, fixed.

salt/orch/update.sls
inercia: I see, thanks for the clarification...

ereslibre:
Confirmed again with this exact PR: Upgraded a 3+2 cluster directly from 2.0 GM to 3.0:

```
Summary for admin.infra.caasp.local_master
-------------
Succeeded: 62 (changed=44)
Failed:     0
-------------
Total states run:     62
Total run time: 1109.206 s
```

and:

```
admin:~ # kubectl get nodes -o wide
NAME       STATUS    ROLES     AGE       VERSION   EXTERNAL-IP   OS-IMAGE                 KERNEL-VERSION          CONTAINER-RUNTIME
master-0   Ready     master    15m       v1.9.7    <none>        SUSE CaaS Platform 3.0   4.4.126-94.22-default   docker://17.9.1
master-1   Ready     master    15m       v1.9.7    <none>        SUSE CaaS Platform 3.0   4.4.126-94.22-default   docker://17.9.1
master-2   Ready     master    15m       v1.9.7    <none>        SUSE CaaS Platform 3.0   4.4.126-94.22-default   docker://17.9.1
worker-0   Ready     <none>    15m       v1.9.7    <none>        SUSE CaaS Platform 3.0   4.4.126-94.22-default   docker://17.9.1
worker-1   Ready     <none>    15m       v1.9.7    <none>        SUSE CaaS Platform 3.0   4.4.126-94.22-default   docker://17.9.1
```

PR #586, opened 2018-06-06
Use the cache whenever something bad happens when refreshing the Pillar.

inercia:
When working on bsc#1093123 I've seen that the orchestration fails becasue 1) haproxy was temporarily down 2) the pillar cannot be accessed 3) `velum.py` crashes with an exception. We can avoid this failure for this kind of transient errors by using the Pillar cache (when present).

Related to [bsc#1093123](https://bugzilla.suse.com/show_bug.cgi?id=1093123).


ereslibre:
I think we want to backport this one.

PR #588, opened 2018-06-07
Fix some orchestration flags

inercia:
* Do not set the `bootstrap_complete` flag in all the nodes: do it only in the nodes that had some role assigned.
  This is important because, when some nodes are _unassigned_ and we bootstrap the cluster, those nodes will have the `bootstrap_complete` set anyway.
  The problem appears when running the _node removal_ orchestration, as it uses that flag for prioritizing nodes when looking for a replacement (it tries to skip nodes with `bootstrap_complete` as it assumes that some role is already assigned and running in those machines).
* Remove the `bootstrap_in_progress` despite the result of the orchestration (as [we currently do in the node removal orchestration](https://github.com/kubic-project/salt/blob/master/salt/orch/removal.sls#L342-L351)... and we should do in the [upddate orchestration too](https://github.com/kubic-project/salt/blob/master/salt/orch/update.sls#L441-L451)).
* Fixed typo in target for disabling the `rebootmngr`

### DIFF
--- a/salt/orch/kubernetes.sls
+++ b/salt/orch/kubernetes.sls
@@ -12,10 +12,11 @@
 {%- set additional_etcd_members = salt.caasp_etcd.get_additional_etcd_members(num_wanted=num_etcd_members,
                                                                               etcd_members=etcd_members) %}
 
-# Ensure the node is marked as bootstrapping
+# Ensure all the nodes are marked with a 'bootstrap_in_progress' flag
 set-bootstrap-in-progress-flag:
   salt.function:
-    - tgt: '*'
+    - tgt: 'roles:(ca|admin|kube-master|kube-minion|etcd)'
+    - tgt_type: grain_pcre
     - name: grains.setval
     - arg:
       - bootstrap_in_progress
@@ -77,7 +78,7 @@ update-modules:
 
 disable-rebootmgr:
   salt.state:
-    - tgt: 'roles:(admin|kube-master|minion|etcd)'
+    - tgt: 'roles:(admin|kube-master|kube-minion|etcd)'
     - tgt_type: grain_pcre
     - sls:
       - rebootmgr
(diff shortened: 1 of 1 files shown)


salt/orch/kubernetes.sls
```diff
@@ -12,10 +12,11 @@
 {%- set additional_etcd_members = salt.caasp_etcd.get_additional_etcd_members(num_wanted=num_etcd_members,
                                                                               etcd_members=etcd_members) %}
 
-# Ensure the node is marked as bootstrapping
+# Ensure all the nodes are marked with a 'bootstrap_in_progress' flag
 set-bootstrap-in-progress-flag:
   salt.function:
-    - tgt: '*'
+    - tgt: 'roles:(ca|admin|kube-master|minion|etcd)'
```
ereslibre: Typo here: `|minion|`

inercia: I did the same typo I fixed some lines below... (faceplam)

Fixing.

salt/orch/kubernetes.sls
```diff
@@ -12,10 +12,11 @@
 {%- set additional_etcd_members = salt.caasp_etcd.get_additional_etcd_members(num_wanted=num_etcd_members,
                                                                               etcd_members=etcd_members) %}
 
-# Ensure the node is marked as bootstrapping
+# Ensure all the nodes are marked with a 'bootstrap_in_progress' flag
 set-bootstrap-in-progress-flag:
   salt.function:
-    - tgt: '*'
+    - tgt: 'roles:(ca|admin|kube-master|kube-minion|etcd)'
```
ereslibre: I'm not sure if we should include the `ca` in this set... It won't harm though I guess.

inercia: Yeah, I thought the same...

PR #593, opened 2018-06-12
Remove mine information when removing a node

ereslibre:
This will avoid to render stale information about critical components,
like `etcd` endpoints in the `etcd` configuration.

`etcd` is very sensitive to this kind of misleading (stale) information,
if more endpoints are provided in `ETCD_INITIAL_CLUSTER` than the ones
that actually exist in the cluster, a new instance of etcd will refuse
to start.

Fixes: bsc#1097001
Fixes: bsc#1097147

### DIFF
--- a/salt/orch/force-removal.sls
+++ b/salt/orch/force-removal.sls
@@ -39,6 +39,12 @@ remove-cluster-wide-removal-grain:
     - kwarg:
         destructive: True
 
+remove-target-mine:
+  salt.function:
+    - tgt: {{ target }}
+    - name: mine.flush
+    - fail_minions: {{ target }}
+
 remove-target-salt-key:
   salt.wheel:
     - name: key.reject
--- a/salt/orch/removal.sls
+++ b/salt/orch/removal.sls
@@ -271,6 +271,14 @@ remove-from-cluster-in-super-master:
     - require:
       - shutdown-target
 
+# remove target information from the mine
+remove-target-mine:
+  salt.function:
+    - tgt: '{{ target }}'
+    - name: mine.flush
+    - require:
+        - remove-from-cluster-in-super-master
+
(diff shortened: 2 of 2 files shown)


ereslibre:
After the second master removal, from a 3+2 to a 1+2 cluster:

```
Succeeded: 26 (changed=16)
Failed:     0
```

```
admin:~ # kubectl get nodes
NAME       STATUS    ROLES     AGE       VERSION
master-0   Ready     master    22m       v1.9.8
worker-0   Ready     <none>    22m       v1.9.8
worker-1   Ready     <none>    22m       v1.9.8
```

```
worker-1:~ # etcdctl cluster-health
member 2fca4ebcbf7277b4 is healthy: got healthy result from https://worker-0:2379
member 8bd2103089e626ae is healthy: got healthy result from https://master-0:2379
member a2d83e07587acbf5 is healthy: got healthy result from https://worker-1:2379
cluster is healthy
```

salt/orch/force-removal.sls
```diff
@@ -39,6 +39,11 @@ remove-cluster-wide-removal-grain:
     - kwarg:
         destructive: True
 
+remove-target-mine:
+  salt.function:
+    - tgt: '{{ target }}'
```
inercia: Is this run in the `target` or in the master? This could be problematic if it runs in the minions and it is not responding....

ereslibre: This is ran on the minion. Nothing depends on this state, so it shouldn't be a problem, but will update with `fail_minions` so we are sure that failure on `target` is an option. Thanks!

jordimassaguerpla:
I merged the backport to 3.0 because tests were passing for that branch.
https://github.com/kubic-project/salt/pull/594

PR #597, opened 2018-06-14
[WIP] early setup some services on updates

inercia:
We cannot wait until all the machines are updated for loading some manifests (like the CNI or the PSP). It is not a good idea to load them before doing the update because these manifests could be incompatible with the old cluster.

So the "best" solution would be to load them once we have updated the first master (and, if no masters are being updated now, we have no other option but to assume they were updated in a previous update orchestration)

bsc#1096992


salt/addons/psp/manifests/10-podsecuritypolicy-privileged.yaml
```diff
@@ -38,7 +38,7 @@ spec:
     - azureDisk
     - azureFile
     - vsphereVolume
-  allowedFlexVolumes: []
+  #allowedFlexVolumes: []
```
kiall: This is also set in 10-podsecuritypolicy-unprivileged.yaml - as far as I understand it, removing this has no effect on the policy (other than it being less explicit)

inercia: Ok, I will just remove it in both files...

inercia: Done.

PR #603, opened 2018-06-17
[3.0] Move the early services setup before updating the masters

inercia:
Move the early services setup even before updating the masters (we can do this by removing some
unnecessary dependencies).

bsc#1096992

### DIFF
--- a/salt/addons/init.sls
+++ b/salt/addons/init.sls
@@ -1,7 +1,5 @@
 include:
-  - kube-apiserver
   - kubectl-config
-  - cri
 
 /etc/kubernetes/addons:
   file.directory:
@@ -9,8 +7,6 @@ include:
     - group:    root
     - dir_mode: 755
     - makedirs: True
-    - require:
-      - file: /etc/cni/net.d/87-podman-bridge.conflist
 
 {% from '_macros/kubectl.jinja' import kubectl_apply_template with context %}
 
--- a/salt/addons/psp/init.sls
+++ b/salt/addons/psp/init.sls
@@ -1,10 +1,9 @@
 {% if salt.caasp_pillar.get('addons:psp', True) %}
 
 include:
-  - kube-apiserver
   - kubectl-config
 
-{% from '_macros/kubectl.jinja' import kubectl, kubectl_apply_template, kubectl_apply_dir_template with context %}
+{% from '_macros/kubectl.jinja' import kubectl_apply_dir_template with context %}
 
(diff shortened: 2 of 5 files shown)


inercia:
> I'm not sure about this overall change TBH. Removing includes that will allow us to properly run individual highstates with their requirements well-defined can have side effects.
> I think it's harder to maintain this logic mid-term because any new includes that we add that change a bit the order in which the orchestration expects the super_master can re-introduce errors again.
> I personally prefer to wait for one master to have updated, and apply those early-services on that one (possibily by generating a special kubeconfig file that can be used sometimes and that points always to the local :6444 internal instance).

Those `includes` were always wrong: we don't need the `kube-apiserver` for running `kubectl` against the API. The only thing you need is a valid `kubeconfig`. The way you generate the kubeconfig is the key.

Waiting for one of the new masters to be up is not working, because it goes via haproxy and the request can reach maybe a new master, maybe an old one. So we must make requests compatible with old masters anyway, so it is just better to load the manifests before the new-old masters mix thing...

ereslibre:
It's not that wrong I think. If we depend on `kube-apiserver` salt will try to start the apiserver locally, and then at the end of `kube-apiserver/init.sls` it will first try to wait for the local instance (`int_ssl_port`) and then (`ssl_port`), so we know that there's something behind haproxy to actually process requests.

flavio:
I'm fine with the changes that involve `/etc/cni/net.d/87-podman-bridge.conflist`, however I think we could even remove this code because we made some changes at the packaging level.

Now (including the current 3.0 project) this broken configuration is provided by a sub-package of podman named `podman-cni-config`. We made this package conflicts with the `kubernetes-kubelet` one, which makes impossible to have it installed on any kubic/caasp node (see [line 74 of the spec file](https://build.suse.de/package/view_file/Devel:CASP:3.0:ControllerNode/podman/podman.spec?expand=1)).

inercia:
> It's not that wrong I think. If we depend on kube-apiserver salt will try to start the apiserver locally, and then at the end of kube-apiserver/init.sls it will first try to wait for the local instance (int_ssl_port) and then (ssl_port), so we know that there's something behind haproxy to actually process requests.

The api server must be properly started (including waiting for the api server) at a higher level, at the orchestration level. For running `kubectl`, the only things we really need are the `kubeconfig` and, if you want to be more explicit, the local `haproxy`.

inercia:
This is the situation with this patch:

medium size update

The three masters updates seem to overlap (not good), while minions seem to be updated in the right order. Service is not unvailable during the update, but it could be improved...

PR #604, opened 2018-06-18
add default network policy for kube-dns and dex

nirmoy:


nirmoy:
@inercia @flavio Please review

salt/cni/policy/default-kube-dex-policy.yaml.jinja
```diff
@@ -0,0 +1,21 @@
...
+  podSelector:
+    matchLabels:
+      "app": "dex"
+  egress:
+  - ports:
+# port for LDAP tcp 10.17.1.0:389
+    - port: 389
+      protocol: TCP
+    - port: 6444
+      protocol: TCP
```
kiall: Humm - we're introducing support for integrating Dex with pre-existing LDAP, SAML or OpenID services. So, Dex will need to ability to connect out to any of those services - which being unmanaged by Kubic/CaaSP, could be on any IP+port combinations.

I wonder how we can handle these scenarios. I suspect we'll need to render specific IPs/ports etc here based on the config data for that. There is probably nothing to do in this PR - just a FYI

Also cc @nanoscopic

salt/cni/policy/default-kube-dns-policy.yaml.jinja
```diff
@@ -0,0 +1,17 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: kube-dns-policy
+  namespace: kube-system
+spec:
+  podSelector:
+    matchLabels:
+      "k8s-app": "kube-dns"
+  ingress:
```
kiall: Don't we need some egress rules here too?

salt/cni/policy/default-kube-dex-policy.yaml.jinja
```diff
@@ -0,0 +1,21 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: kube-dex-policy
```
kiall: Can we set the names to something like this:

    name: suse:caasp:dex-networkpolicy

Lots of our resources now use this pattern, so it's blindingly obvious these are created+managed by CaaSP etc.

salt/cni/policy/default-kube-dex-policy.yaml.jinja
nirmoy: Yes I would need more input here, what ports dex uses.

salt/cni/policy/default-kube-dns-policy.yaml.jinja
nirmoy: Ingress should be enough given that we trust our application/container.

salt/cni/policy/default-kube-dex-policy.yaml.jinja
nirmoy: yes I will modify that

nirmoy: worker-0:~ # kubectl apply -f  dns.yaml
The NetworkPolicy "suse:caasp:dns-networkpolicy" is invalid: metadata.name: Invalid value: "suse:caasp:dns-networkpolicy": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')

kiall: Arg. Some resources get validated with a DNS name pattern. I'm not sure I understand why K8S would apply that to NetworkPolicy resources though. The same happened with PodSecurityPolicy resources.

For those, we used: 

    metadata:
      name: suse.caasp.psp.privileged

Can we do the same here and use periods instead of colons.

nirmoy:
@inercia please review again when you have time

PR #606, opened 2018-06-18
Add configmap from pillar data to dex ldap connectors

nanoscopic:
Goes together with changes from https://github.com/kubic-project/velum/pull/585

### DIFF
--- a/salt/addons/dex/manifests/15-configmap.yaml
+++ b/salt/addons/dex/manifests/15-configmap.yaml
@@ -43,6 +43,39 @@ data:
           groupAttr: uniqueMember
 
           nameAttr: cn
+    {% for con in salt['pillar.get']('dex:connectors', []) %}
+    {% if con['type'] == 'ldap' %}
+    - type: ldap
+      id: {{ con['id'] | yaml_dquote }}
+      name: {{ con['name'] | yaml_dquote }}
+      config:
+        host: {{ con['server'] | yaml_dquote }}
+        startTLS: {{ 'true' if con.get('start_tls',false) else 'false' }}
+        {% if con['bind']['anonymous'] %}
+        # bindDN and bindPW not present; anonymous bind will be used
+        {% else %}
+        bindDN: {{ con['bind']['dn'] | yaml_dquote }}
+        bindPW: {{ con['bind']['pw'] | yaml_dquote }}
+        {% endif %}
+        usernamePrompt: {{ con['username_prompt'] | yaml_dquote }}
+        rootCAData: {{ con['root_ca_data'] | replace("\n","") | yaml_dquote }}
+        userSearch:
+          baseDN: {{ con['user']['base_dn'] | yaml_dquote }}
+          filter: {{ con['user']['filter'] | yaml_dquote }}
+          username: {{ con['user']['attr_map']['username'] | yaml_dquote }}
+          idAttr: {{ con['user']['attr_map']['id'] | yaml_dquote }}
+          emailAttr: {{ con['user']['attr_map']['email'] | yaml_dquote }}
+          nameAttr: {{ con['user']['attr_map']['name'] | yaml_dquote }}
+        groupSearch:
+          baseDN: {{ con['group']['base_dn'] | yaml_dquote }}
+          filter: {{ con['group']['filter'] | yaml_dquote }}
+
+          userAttr: {{ con['group']['attr_map']['user'] | yaml_dquote }}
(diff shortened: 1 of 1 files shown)


salt/addons/dex/manifests/15-configmap.yaml
```diff
@@ -43,6 +43,36 @@ data:
           groupAttr: uniqueMember
 
           nameAttr: cn
+    {% for con in pillar['dex_connectors'] %}
```
kiall: If we swap this out for something like this:

    {{ salt['pillar.get']('dex_connectors', []) }}

I believe this will then pass CI, and avoid blocking this PR on the Velum side of this change.

As a side note - can we "namespace" the keys? We've been slowly migrating towards a pattern that looks like `dex:connectors` in salt, and `{"dex": "connectors": [...]}` in JSON.

salt/addons/dex/manifests/15-configmap.yaml
```diff
@@ -43,6 +43,36 @@ data:
...
+    - type: ldap
+      id: "{{ con['id'] }}"
+      name: "{{ con['name'] }}"
+      config:
+        host: "{{ con['server'] }}"
```
kiall: For most of these variables, replacing this:

    host: "{{ con['server'] }}"

with this:

    host: {{ con['server'] | yaml_dquote }}

would be safer - e.g.if any of the input values contains a ", it would be handled correctly.

salt/addons/dex/manifests/15-configmap.yaml
nanoscopic: Agreed that adding the default value of empty array would be better and allow the change to go in without breaking prior to the output of the connectors. I will make that change and update this PR.

As for pulling connectors under a dex key, that change could certainly be made, but I am not sure what benefit it would have as there is no other dex configuration output besides the connectors coming out of the pillars.

salt/addons/dex/manifests/15-configmap.yaml
nanoscopic: Agreed that adding dquote filter to the values is a good change. Will do so and update this PR.

salt/addons/dex/manifests/15-configmap.yaml
kiall: > As for pulling connectors under a dex key, that change could certainly be made, but I am not sure what benefit it would have as there is no other dex configuration output besides the connectors coming out of the pillars.

We only have connector configuration, for now... Down the line, we will likely end up with more options - having them all nested together under a `dex` key will make this cleaner.

salt/addons/dex/manifests/15-configmap.yaml
```diff
@@ -43,6 +43,38 @@ data:
...
+    {% for con in salt['pillar.get']('dex:connectors', []) %}
+    {% if con['type'] == 'ldap' %}
+    - type: ldap
+      id: "{{ con['id'] }}"
+      name: "{{ con['name'] }}"
+      config:
+        host: "{{ con['server'] }}"
+        startTLS: {{ con['start_tls'] }}
+        {% if con['bind']['anonymous'] %}
+        {% else %}
```
inercia: Couldn't we replace this

```jinja
        {% if con['bind']['anonymous'] %}
        {% else %}
        ...
```

by something like

```jinja
        {% if not con['bind']['anonymous'] %}
        ...
```

?

salt/addons/dex/manifests/15-configmap.yaml
nanoscopic: Understood; I have updated both PRs to use a dex key to group configuration. Offhand it also does make sense as we will need information related to groups there in the future.

salt/addons/dex/manifests/15-configmap.yaml
nanoscopic: I agree that it would be equivalent and be one less line. Now that you point it out though it would really be best I think to output a comment into the configuration file instead of nothing when anonymous is used. That would make it clear that the lack of the bind information is telling dex to do an anonymous bind, which is unclear looking at the configuration file without a comment being present.

inercia: Yeah, I think you should at least add a comment clarifying why you are leaving an empty `if` statement...

nanoscopic:
Changed to using yaml_dquote to escape attributes in all cases. Added a comment in empty if block to make it clear why it exists and also to show the comment in the output configuration file itself for clarity to anyone reviewing that file.

PR #607, opened 2018-06-20
Salt 2018 compatibility

bergmannf:
Changes required to make our salt states run with `salt 2018.3`.

### DIFF
--- a/config/master.d/50-reactor.conf
+++ b/config/master.d/50-reactor.conf
@@ -6,3 +6,5 @@ reactor:
     - /usr/share/salt/kubernetes/reactor/etc-hosts.sls
   - 'salt/minion/*/start':
     - /usr/share/salt/kubernetes/reactor/etc-hosts.sls
+    - /usr/share/salt/kubernetes/reactor/sync-modules.sls
+    - /usr/share/salt/kubernetes/reactor/update-mine.sls
--- a/reactor/sync-modules.sls
+++ b/reactor/sync-modules.sls
@@ -1,0 +1,3 @@
+sync_modules:
+  local.saltutil.sync_modules:
+    - tgt: {{ data['id'] }}
--- a/reactor/update-mine.sls
+++ b/reactor/update-mine.sls
@@ -1,0 +1,3 @@
+update_mine:
+  local.mine.update:
+    - tgt: {{ data['id'] }}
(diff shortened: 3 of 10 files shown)


salt/addons/dex/legacy.sls
```diff
@@ -0,0 +1,24 @@
+include:
+  - .main
```
inercia: Is this `legacy.sls` invoked from anywhere? I cannot find any reference...

bergmannf: Ah - I forgot to add it back into `init.sls`.

I am still hoping to find a good way to not having to show the dependencies in the state files *and* the orchestration (`orch.kubernetes`) as well.

salt/addons/dex/pre.sls
```diff
@@ -0,0 +1,27 @@
...
+         cn = 'Dex',
+         extra_alt_names = alt_master_names(dex_alt_names)) }}
+
+# This file needs to be explicitly copied over *before* ./main.sls runs
+# kubectl_apply_dir_template(). Otherwise one of the templates attempts to
+# access the file, and it does not exist yet, breaking the orchestration.
+# Unfortunately using the default include mechanism of salt, to force order
+# doesn't seem to work when running via orchestrate runner, so this is
+# split in different files that the orch runner calls with explicit
+# dependencies.
```
inercia: Isn't enough with an `include` **and** a `require` in the state that requires it? I find this split in different files a bit confusing...

bergmannf:
[Test salt update](https://trello.com/c/KbKz6kJY/305-test-salt-update)

salt/addons/dex/main.sls
```diff
@@ -0,0 +1,13 @@
+include:
+  - .pre
+
+explicit-pre-dependency:
+  test.nop:
+  - require:
+      - sls: addons.dex.pre
```
inercia: Thiss is not very elegant.. do we really need this? We are already setting the order in the orchestration...

bergmannf: @inercia I also didn't like having to artificially splitting the states.

The problem I had with everything being in one file - the `copying` of the `15-secret.yaml` did not happen in that case - I *assumed* that is because the macro-expansion have to run first, which already run other states.

However I just tried moving everything into one file again and this time it worked ...

bergmannf:
@inercia Apparently the `dex` changes were unneeded after all - retrying it again it works while keeping it in one file. Guess I just was unlucky, when trying it the first time and something else prevented it from working (maybe the missing reactors).

PR #608, opened 2018-06-20
During upgrade, ensure masters always have the correct taints

kiall:
When migrating from the "old" to "new" names for the kubelets, we pre-create
the new node so that we can clone the network config. This means the kubelet
is NOT self-registering, and the "single use options" like --register-with-taints
are ignored. This means the kubelet is connected from the period of time where it
starts, to where salt later forcefully adds the taint. Any pods created during
this window could end up scheduled to the master.

bsc#1098383

### DIFF
--- a/salt/kubelet/update-pre-orchestration.sls
+++ b/salt/kubelet/update-pre-orchestration.sls
@@ -11,7 +11,11 @@ include:
     - source: salt://kubelet/update-pre-orchestration.sh
     - mode: 0755
   cmd.run:
-    - name: /tmp/kubelet-update-pre-orchestration.sh {{ grains['machine_id'] + "." + pillar['internal_infra_domain'] }} {{ grains['nodename'] }}
+{% if "kube-master" in salt['grains.get']('roles', []) %}
+    - name: /tmp/kubelet-update-pre-orchestration.sh {{ grains['machine_id'] + "." + pillar['internal_infra_domain'] }} {{ grains['nodename'] }} master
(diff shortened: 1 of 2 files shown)


salt/kubelet/update-pre-orchestration.sls
```diff
@@ -11,7 +11,11 @@ include:
     - source: salt://kubelet/update-pre-orchestration.sh
     - mode: 0755
   cmd.run:
-    - name: /tmp/kubelet-update-pre-orchestration.sh {{ grains['machine_id'] + "." + pillar['internal_infra_domain'] }} {{ grains['nodename'] }}
+{% if "kube-master" in salt['grains.get']('roles', []) %}
+    - name: /tmp/kubelet-update-pre-orchestration.sh {{ grains['machine_id'] + "." + pillar['internal_infra_domain'] }} {{ grains['nodename'] }} master
+{% else %}
+    - name: /tmp/kubelet-update-pre-orchestration.sh {{ grains['machine_id'] + "." + pillar['internal_infra_domain'] }} {{ grains['nodename'] }} worker
+{% endif %}
```
inercia: Minor suggestion:

```jinja
{% set role = "master" if "kube-master" in salt['grains.get']('roles', []) else "worker" %}
- name: /tmp/kubelet-update-pre-orchestration.sh {{ grains['machine_id'] + "." + pillar['internal_infra_domain'] }} {{ grains['nodename'] }} {{ role }}
```

PR #610, opened 2018-06-20
Do not install recommends

flavio:
As discussed yesterday on irc, we should instruct salt to not install recommended packages.


kiall:
Q - Should we instead just remove all the pkg.installed entries? Given how CaaSP/Kubic works, they are unnecessary steps anyway?

PR #614, opened 2018-06-21
Allow `etcd` to grow as required and shrink to optimal etcd cluster sizes on corner cases

ereslibre:
- [x] Fix documentation everywhere
- [x] Add test cases

Improve `etcd` configuration handling to allow it to grow as needed. This
change includes:

* Adding several masters at the same time
  * `etcd` will grow instance by instance still, as recommended by the `etcd` administration best practices.

* Try to use the current endpoints reported by `etcd`. This makes much
  easier to grow several instances one by one without having to relay
  on internal hacks to properly set up `ETCD_INITIAL_CLUSTER` environment
  variable.

* Add helper methods that allow us to list current members (active and
  unstarted)

* Differentiate between the first bootstrap (`ETCD_INITIAL_CLUSTER_STATE`
  defaults to `new`) and *any* other run, where `ETCD_INITIAL_CLUSTER_STATE`
  will be `existing`, as the `etcd` cluster is already running.

When we grow, we take into account the golden ratio; however, when shrinking
the cluster we don't. It might happen that a cluster ends up with not
recommended etcd number of nodes (2, 4, 6...) depending on how it grew before
and how it shrank.

This logic makes sure that we are always on an etcd golden ratio, also
on corner cases when removing nodes.

Fixes: bsc#1098433
Fixes: bsc#1098064
Fixes: bsc#1098161

### DIFF
--- a/salt/_modules/caasp_etcd.py
+++ b/salt/_modules/caasp_etcd.py
@@ -1,14 +1,21 @@
 from __future__ import absolute_import
 
+import re
 import subprocess
 
+from urlparse import urlparse
 
-# minimum number of etcd masters we recommend
+# note: do not import caasp modules other than caasp_log
+from caasp_log import debug, error, warn
+
+# minimum number of etcd members we recommend
 MIN_RECOMMENDED_MEMBER_COUNT = 3
 
 # port where etcd listens for clients
 ETCD_CLIENT_PORT = 2379
 
+# default etcd peer port
+ETCD_PEER_PORT = 2380
 
 def __virtual__():
     return "caasp_etcd"
@@ -63,7 +70,7 @@ def get_cluster_size(**kwargs):
 
         member_count = _optimal_etcd_number(num_masters)
         if member_count < MIN_RECOMMENDED_MEMBER_COUNT:
-            # Attempt to increase the number of etcd master to 3,
+            # Attempt to increase the number of etcd members to 3,
             # however, if we don't have 3 nodes in total,
             # then match the number of nodes we have.
             increased_member_count = len(masters) + len(minions)
@@ -93,6 +100,46 @@ def get_cluster_size(**kwargs):
     return member_count
 
 
+def get_surplus_etcd_members(num_wanted=None, targets=[], **kwargs):
+    '''
+    Taking into account
+
+      1) the current number of etcd members, and
+      2) the number of etcd nodes we should be running in the
+         cluster (obtained with `get_cluster_size()`)
+      3) the `targets` to be removed overall
+
+    get a list of surplus nodes (IDs) that should not run `etcd`, not included
+    in `targets`
+    '''
+    excluded = kwargs.get('excluded', [])
+
+    current_etcd_members = __salt__['caasp_nodes.get_from_args_or_with_expr'](
+        'etcd_members', kwargs, 'G@roles:etcd')
+    num_current_etcd_members = len(current_etcd_members)
+
+    targets_and_etcd_members = set(targets).intersection(set(current_etcd_members))
+
+    # the number of etcd members that should be in the cluster
+    num_wanted_etcd_members = num_wanted or get_cluster_size(**kwargs)
+    # ... and the number we are passing
+    num_surplus_etcd_members = num_current_etcd_members - num_wanted_etcd_members
+
+    if num_surplus_etcd_members <= 0:
+        debug('get_surplus_etcd_members: we dont need to remove etcd members')
+        return []
+
+    debug('get_surplus_etcd_members: curr:%d wanted:%d -> %d surplus',
+          num_current_etcd_members, num_wanted_etcd_members, num_surplus_etcd_members)
+
+    result = __salt__['caasp_nodes.get_with_prio_for_role'](
+        num_surplus_etcd_members, 'etcd-removal',
+        unassigned=False,
+        excluded=targets + excluded)
+
+    return result[:-len(targets_and_etcd_members)]
+
+
 def get_additional_etcd_members(num_wanted=None, **kwargs):
     '''
     Taking into account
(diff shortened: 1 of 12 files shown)


ereslibre:
I need to also check how this plays with machines that had etcd but died, and how that interacts with node addition/removal... Still WIP.

ereslibre:
> I'm concerned with using the cluster information reported by etcd as the source of truth. I'm thinking on a 1-member etcd cluster that is updated (so no etcd would be available for a while when restarting the only master), or a etcd member that is temporarily overloaded and doesn't respond, or a etcd cluster that is just down. Would this code resist those situations? Would this consider that a new cluster must be created? Note well that the cluster membership is also a persisted information that could conflict with the etcd-reported information...

I will try to answer all those questions.  At the moment, if `etcd` cannot provide this information (for whatever reason), we fallback to the information we have based on grains. However, it's very important to understand that for handling the `etcd` cluster properly we need to find out what's the current `etcd` status.

> Anyway I don't think it is a good idea to do this kind of mega-refactoring of something so sensitive as the etcd cluster management at this moment. I definitely think that the etcd code should be cleaned up at some moment, but at this time we need to stabilize the code inistead of turning some fundamental parts of our code upside down.

The current code is basically broken, it's trivial to find any way to break `etcd` at the moment, and instead of writing more hacks I found easier and more convenient to rethink a couple of things. I'm still working on how this behaves when some `etcd` instances are down for example.

IMHO, if QA tests this patch and confirm that they don't have a way to break it (I couldn't break it with zillions of tests), I don't see a reason why not to go ahead with this.

ereslibre:
> I'm concerned with using the cluster information reported by etcd as the source of truth. I'm thinking on a 1-member etcd cluster that is updated (so no etcd would be available for a while when restarting the only master)

This is working fine, just double checked. When a 1-member etcd cluster is updated, if we ask for the endpoints with self function, the list from the salt grains will be used as a fallback (what we were doing all the time previously before this patch).

> or a etcd member that is temporarily overloaded and doesn't respond, or a etcd cluster that is just down. Would this code resist those situations?

I think we should specify "resist" here, resist to what? For example, if you have a degraded cluster, you cannot do a `member add`, you first need to `member remove` and then move that faulty one somewhere else. We are really close to do that in an automated way with this code now, as we can compare what `etcd` has, what we need to do in order to migrate from what `etcd` has to what we want it to be.

In one sentence: we can do the difference between "current status" and "desired status" and act in consequence.

> Would this consider that a new cluster must be created? Note well that the cluster membership is also a persisted information that could conflict with the etcd-reported information...

Not in this code, but when your cluster is `unavailable` (worse than `degraded`) this code **could** also do that, and it should be fairly simple. I wanted to fix the `etcd` most dire issues we have at the moment (to not include even more changes to this PR), but I'll continue the improvements in further work.

inercia:
>> Would this consider that a new cluster must be created? Note well that the cluster membership is also a persisted information that could conflict with the etcd-reported information...

> Not in this code, but when your cluster is unavailable (worse than degraded) this code could also do that, and it should be fairly simple. I wanted to fix the etcd most dire issues we have at the moment (to not include even more changes to this PR), but I'll continue the improvements in further work.

My concern is what would happen in a cluster formed by `m1`, `m2` and `m3`, if the master `m1` is down (for example, because I just do `systemctl stop etcd`). I'm concerned if these states would be fooled by the status reported by `etcdctl`, while we should just try to start the `etcd` service in `m1` (as with `service.running`) (in a more general case, all three masters could be down and I think we wouldn't need any `add` and starting all the servers would be enough, as their cluster membership has already been persisted)

ereslibre:
> My concern is what would happen in a cluster formed by m1, m2 and m3, if the master m1 is down (for example, because I just do systemctl stop etcd). I'm concerned if these states would be fooled by the status reported by etcdctl, while we should just try to start the etcd service in m1 (as with service.running) (in a more general case, all three masters could be down and I think we wouldn't need any add and starting all the servers would be enough, as their cluster membership has already been persisted)

Please, note that we are running `etcdctl member list` to detect the active an unstarted members. This is not affected by the status of any etcd member (it's not a health check). As long as the cluster can answer to us at all we have the member list; if not, we fallback to what we were doing previously (relying on machines having the `etcd` role). However, the problem with only relying on the `etcd` role is that you cannot produce a correct list of `ETCD_INITIAL_CLUSTER` when new members join, because even if machines have this grain it doesn't mean they actually joined the cluster; and etcd is really really picky for the number of endpoints having to match the number of endpoints `member list` has.

Besides this, this whole logic of comparing what should be and what there is at the moment in the cluster can allow us to know what to do (what machines should have `etcd` and what machines actually have `etcd`). In this process, we can remove extra members and add missing members one by one, until the real state of the `etcd` cluster matches what salt says it should be. All in a idempotent fashion.

I will add more tests for more cases and I will consider this patch done. I still want to improve some things, but really; I literally couldn't make the cluster break (there's some edge cases I need to fix but `etcd` is always in a healthy state no matter what you do to the cluster).

QA will be testing this soon, and since this only messes with `etcd` I think that if they give their OK with all the tests they can come up with I think it should be safe to backport if you don't have dire issues with the changes.

ereslibre:
Keeping the `WIP` until QA confirms that they cannot break the `etcd` cluster by adding or removing machines.

flavio:
I acknowledge this is a big PR and I think now is a better time to get it merged. It's important to handle etcd add/remove operations in a smoother way now that we have node add/removal released.

BTW, I'm really looking forward to drop all this code as soon as we can switch to use the etcd-operator.

ereslibre:
[Allow `etcd` to grow as required and shrink to optimal etcd cluster sizes on corner cases](https://trello.com/c/RgWLIma2/1161-allow-etcd-to-grow-as-required-and-shrink-to-optimal-etcd-cluster-sizes-on-corner-cases)

MaximilianMeister:
@ereslibre can you rebase? the backport already got merged https://github.com/kubic-project/salt/pull/660

ereslibre:
Rebased

PR #615, opened 2018-06-25
Move some kubectl Jinja code to a kubectl state.

inercia:
Move some kubectl Jinja code to a kubectl state.

feature#kubectl_state


ereslibre:
It seems it has conflicts now, this would need a rebase.

PR #616, opened 2018-06-26
Generate the /etc/hosts file with a custom state

inercia:
Generate the /etc/hosts file from a state, merging our entries with previously found entries.

* first time we runn this state, we copy `/etc/hosts` to `/etc/caasp/hosts`
* after that, `/etc/hosts` is written as we have done so far (with entries for _master_, _workers_, etc) and **merged** with all the entries found in `/etc/caasp/hosts`.

(Users are instructed with a comment in `/etc/hosts` to add their custom entries in `/etc/caasp/hosts`)

See https://bugzilla.suse.com/show_bug.cgi?id=1098334

bsc#1098334


salt/etc-hosts/init.sls
```diff
@@ -1,20 +1,19 @@
...
-    - append_if_not_found: True
 {% else %}
 {# See https://github.com/saltstack/salt/issues/14553 #}
 etc-hosts-dummy:
   cmd.run:
     - name: "echo saltstack bug 14553"
 {% endif %}
 
+# TODO: all these `docker cp` could be removed:
+#       we could just mount the /etc/hosts file
```
ereslibre: Should we do this in a follow-up PR?

bergmannf:
I think there might be a hiccup somewhere - when I run this on the `SLE15` cluster I get the following output in `/etc/hosts` on `admin`:

```bash
#
# This file is automatically generated/managed by CaaSP/Salt
# Please add any custom entries in /etc/caasp/hosts
# Any other modification will be lost...
#
10.17.1.0        admin admin.infra.caasp.local
10.17.2.0        c4a68b42abf84ab1a3f1e819b32c9f36 c4a68b42abf84ab1a3f1e819b32c9f36.infra.caasp.local master-0 master-0.infra.caasp.local
10.17.3.0        94d1f49319bd444ba44e2963a2b5b831 94d1f49319bd444ba44e2963a2b5b831.infra.caasp.local worker-0 worker-0.infra.caasp.local
10.17.3.1        38b7d32ccd9f4c8d90be65b5816d02e1 38b7d32ccd9f4c8d90be65b5816d02e1.infra.caasp.local worker-1 worker-1.infra.caasp.local
127.0.0.1        admin admin.infra.caasp.local api api.infra.caasp.local ldap.infra.caasp.local localhost master-0.devenv.caasp.suse.net
::1        admin admin.infra.caasp.local ipv6-localhost ipv6-loopback localhost
fe00::0        ipv6-localnet ipv6-localnet
ff00::0        ipv6-mcastprefix ipv6-mcastprefix
ff02::1        ipv6-allnodes ipv6-allnodes
ff02::2        ipv6-allrouters ipv6-allrouters
ff02::3        ipv6-allhosts ipv6-allhosts
```

In `127.0.0.1` there is also `master-0.devenv.caasp.suse.net` - I didn't see this on the worker nodes, there the aliases were fine.

salt/etc-hosts/init.sls
kiall: docker treats /etc/hosts as special, it can't be mounted in.

salt/etc-hosts/update-pre-reboot.sls
```diff
@@ -1,8 +1,16 @@
 # Remove our /etc/hosts entries, so that Systemd/Wicked hostname
 # logic doesn't pick up out /etc/hosts entry names.
```
kiall: So - while this isn't a new issue, it's worth pointing out this won't run when an operator just reboots a machine, e.g. with `shutdown -r now` or due to power loss etc. What exactly happens when this code doesn't run and the reboot happens?

inercia: I don't fully understand you question. It will happen the same that happened before when `/etc/hosts` was managed by a Salt state.

salt/etc-hosts/init.sls
ereslibre: Hmmm, I am pretty sure you can mount `/etc/hosts`? Our problem before was that it was an atomic write if it was modified, and once you mounted it you weren't able to update it, so we had to `docker cp` it.

inercia: Yeah, I'm not sure you can't mount either. I think the real problem was that the bind-mount really references an inode, so Salt was updating the file by creating a new copy of the file and overwritting the old copy, while the container kept the old copy... In contrast, this Python code does the replacement in place, so no file is replaced. So maybe we could try to mount it and see what happens...

salt/etc-hosts/update-pre-reboot.sls
bergmannf: @kiall I agree with @inercia here - this was a problem before, and I doubt that we can fix it for all cases. In the end there could always be a power outage that could trigger not running our code.

PR #622, opened 2018-07-05
switch to etcd3 as a storage back-end

MaximilianMeister:
upgrade#etcdctl

## note

~~still untested and WIP~~ bootstrap was green :+1: locally

@mjura will look at the upgrade migration/orch  from 2 to 3

### DIFF
--- a/salt/_modules/caasp_etcd.py
+++ b/salt/_modules/caasp_etcd.py
@@ -20,6 +20,10 @@ class NoEtcdServersException(Exception):
     pass
 
 
+def api_version():
+    return __salt__['pillar.get']('etcd_version', 'etcd3')
+
+
 def _optimal_etcd_number(num_nodes):
     if num_nodes >= 7:
         return 7
@@ -183,9 +187,14 @@ def get_etcdctl_args(skip_this=False):
     Build the list of args for 'etcdctl'
     '''
     etcdctl_args = []
-    etcdctl_args += ["--ca-file", __salt__['pillar.get']('ssl:ca_file')]
-    etcdctl_args += ["--key-file", __salt__['pillar.get']('ssl:key_file')]
-    etcdctl_args += ["--cert-file", __salt__['pillar.get']('ssl:crt_file')]
+    if api_version() == 'etcd2':
+        etcdctl_args += ["--ca-file", __salt__['pillar.get']('ssl:ca_file')]
+        etcdctl_args += ["--key-file", __salt__['pillar.get']('ssl:key_file')]
+        etcdctl_args += ["--cert-file", __salt__['pillar.get']('ssl:crt_file')]
+    else:
+        etcdctl_args += ["--cacert", __salt__['pillar.get']('ssl:ca_file')]
+        etcdctl_args += ["--key", __salt__['pillar.get']('ssl:key_file')]
+        etcdctl_args += ["--cert", __salt__['pillar.get']('ssl:crt_file')]
 
     etcd_members_lst = get_endpoints(skip_this=skip_this)
 
@@ -211,6 +220,10 @@ def get_member_id(nodename=None):
                   the local node will be used.
     '''
     command = ["etcdctl"] + get_etcdctl_args() + ["member", "list"]
+    if api_version() == 'etcd2':
+        command.insert(0, "ETCDCTL_API=2")
+    else:
+        command.insert(0, "ETCDCTL_API=3")
 
     target_nodename = nodename or __salt__['caasp_net.get_nodename']()
 
--- a/salt/_states/caasp_etcd.py
+++ b/salt/_states/caasp_etcd.py
@@ -14,6 +14,10 @@ DEFAULT_ATTEMPTS_INTERVAL = 2
 ETCD_PEER_PORT = 2380
 
 
+def api_version():
+    return __salt__['caasp_etcd.api_version']()
+
+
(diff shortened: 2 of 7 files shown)


salt/_states/caasp_etcd.py
```diff
@@ -56,7 +66,10 @@ def member_add(name, **kwargs):
     this_nodename = __salt__['caasp_net.get_nodename']()
     this_peer_url = 'https://{}:{}'.format(this_nodename, port)
 
-    name = 'member add {} {}'.format(this_id, this_peer_url)
+    if version() == 'etcdv2':
```
mjura: I think it should be `etcd2`

salt/_states/caasp_etcd.py
```diff
@@ -34,6 +38,9 @@ def etcdctl(name, retry={}, **kwargs):
 
     args = __salt__['caasp_etcd.get_etcdctl_args_str'](skip_this=skip_this)
     cmd = 'etcdctl {} {}'.format(args, name)
+    if version() == 'etcd3':
+        cmd = 'ETCDCTL_API=3 {}'.format(cmd)
```
mjura:
```
else 
    cmd = 'ETCDCTL_API=2 {}'.format(cmd)
```

salt/_modules/caasp_etcd.py
```diff
@@ -211,6 +220,8 @@ def get_member_id(nodename=None):
                   the local node will be used.
     '''
     command = ["etcdctl"] + get_etcdctl_args() + ["member", "list"]
+    if version() == 'etcd3':
+        command.insert(0, "ETCDCTL_API=3")
```
mjura:
```
else
      cmd = 'ETCDCTL_API=3 {}'.format(cmd)
```

salt/_modules/caasp_etcd.py
```diff
@@ -20,6 +20,10 @@ class NoEtcdServersException(Exception):
     pass
 
 
+def version():
```
kiall: This method name could be a little confusing.. It's not really the "etcd version", it's the "etcd API version" - maybe rename this to `def api_version()` to avoid confusion?

(And yes, the pillar key suffers from the same issue :()

salt/orch/etcd-migrate.sls
```diff
@@ -46,7 +48,7 @@ backup-etcd:
     - tgt_type: grain_pcre
     - name: cmd.run
     - arg:
-      - mkdir /tmp/backup; btrfs subvolume snapshot /var/lib/etcd /tmp/backup/etcd
+      - if [ -d /var/lib/etcd/member ] && ! [ -d /tmp/backup ]; then mkdir /tmp/backup; btrfs subvolume snapshot /var/lib/etcd /tmp/backup/etcd; fi
```
kiall: I wonder should we put a datestamp into the folder name? to avoid repeated attempts blowing away the original backup.

salt/_modules/caasp_etcd.py
MaximilianMeister: i wonder changing the pillar key would likely force us to do a migration, does it?

salt/orch/etcd-migrate.sls
MaximilianMeister: i guess the `! [ -d /tmp/backup ]` condition will prevent this

mjura: I've added `date` timestamp

salt/etcd/etcdctl.conf.jinja
```diff
@@ -5,15 +5,19 @@
...
 ETCDCTL_CERT_FILE={{ pillar['ssl']['crt_file'] }}
 ETCDCTL_KEY_FILE={{ pillar['ssl']['key_file'] }}
 
 # etcd v3 style flags
+ETCDCTL_ENDPOINTS="https://{{ grains['nodename'] }}:2379"
 ETCDCTL_CACERT={{ pillar['ssl']['ca_file'] }}
 ETCDCTL_CERT={{ pillar['ssl']['crt_file'] }}
 ETCDCTL_KEY={{ pillar['ssl']['key_file'] }}
+
+ETCDCTL_API=3
```
kiall: Shouldn't this be based on the `caasp_etcd.api_version` pillar?

MaximilianMeister: yeah, but we still need to parse it

MaximilianMeister: updated

mjura:
Re-run CI

flavio:
What's blocking us from merging that into master?

salt/_modules/caasp_etcd.py
```diff
@@ -211,6 +220,10 @@ def get_member_id(nodename=None):
                   the local node will be used.
     '''
     command = ["etcdctl"] + get_etcdctl_args() + ["member", "list"]
+    if api_version() == 'etcd2':
+        command.insert(0, "ETCDCTL_API=2")
+    else:
+        command.insert(0, "ETCDCTL_API=3")
```
ereslibre: I know it's too late for this, but given that this `command` gets ran directly by `subprocess.check_output` I don't think this is working fine, as the result of the `check_output` command will likely be `OSError: [Errno 2] No such file or directory` as it won't be able to find the `ETCDCTL_API=version` executable in the `PATH`.

We could move the envvar to the end of the command (it should still apply) or use `popen` directly instead.

PR #624, opened 2018-07-05
Fix update problems

ereslibre:
* Remove all dependencies on `container-feeder`.
  * This is a one-shot service that runs when the machine boots and dies,
we don't need to stop it or explicitly start it, since this will happen
off-band automatically. All we care is that this service is enabled,
so each time a machine gets rebooted, if a new snapshot applies it will
load the new images from the new snapshot, and those runtime dependencies
are managed by systemd.

  * On an upgrade process we are going to perform different migrations; only
perform these migrations on machines that are part of the current subset
of machines to be updated.

* Bypass haproxy and wait for the kubelet to be back before continuing with the upgrade process. (Fixes: bsc#1100212)


ereslibre:
This needs confirmation from QA.

flavio:
About the container-feeder dependency. These have been introduced with CRI-O because:
  * By default container-feeder is configured to load images into the docker engine
  * By the time a node is up and running and registered against the salt master, container-feeder has already completed the import of the docker images
  * Nodes deployed with docker as CRI will always have the images needed by the orchestration (eg: the HAProxy image)

All o fthat is not true when deploying a node with CRI-O as CRI. In that case the orchestration immediately needs some images (like the HAProxy one) but container feeder has to be reconfigured to load images into the CRI-O storage, has to be restarted and has to complete the import (which can take some time).

Is there any reason why we are getting rid of these requirements? Are they interfering with the upgrades?

ereslibre:
> All o fthat is not true when deploying a node with CRI-O as CRI. In that case the orchestration immediately needs some images (like the HAProxy one) but container feeder has to be reconfigured to load images into the CRI-O storage, has to be restarted and has to complete the import (which can take some time).

Absolutely, completely forgot about this. I'll remove the `container-feeder` part of the PR.

flavio:
Everything LGTM inside of "[Only perform migrations on machines that are going to be updated](ca580b36858c94a78886585f44604f8eee7d6fc6)" commit

flavio:
About the "[Bypass haproxy and wait for the kubelet to be back before continuing](https://github.com/kubic-project/salt/pull/624/commits/e3f7974edf8dc2e6a57c2e23cf9b53732682907c)":

* I would like to see what the others think about it (ping @inercia and @kiall)
* I've to admit I'm a bit scared to mess around with iptables like that, but I see the point.
* Question: don't we have to remove the iptables rules once everything is back to normality?

ereslibre:
> Question: don't we have to remove the iptables rules once everything is back to normality?

This should not be necessary as we aren't persisting the `iptables` rules. After a reboot all rules will be gone.

ereslibre:
I'm testing an alternate approach.

PR #625, opened 2018-07-05
[3.0] Fix update issues

ereslibre:
* Remove all dependencies on `container-feeder`.
  * This is a one-shot service that runs when the machine boots and dies,
we don't need to stop it or explicitly start it, since this will happen
off-band automatically. All we care is that this service is enabled,
so each time a machine gets rebooted, if a new snapshot applies it will
load the new images from the new snapshot, and those runtime dependencies
are managed by systemd.

  * On an upgrade process we are going to perform different migrations; only
perform these migrations on machines that are part of the current subset
of machines to be updated.

* Bypass haproxy and wait for the kubelet to be back before continuing with the upgrade process. (Fixes: bsc#1100212)

Backport of https://github.com/kubic-project/salt/pull/624


ereslibre:
I'm testing an alternate approach.

PR #626, opened 2018-07-05
Fix update issues

ereslibre:
* Migrate all labels when renaming a node (builtin and user-defined labels). (Fixes: bsc#1100891)
* Add haproxy migration sls to apply during upgrade (Fixes: bsc#1100212)


ereslibre:
@flavio, @inercia: Still testing the patch, but until now it seems to work as expected. It's a WIP, but I'm asking for your opinion on the `migrations` folder structure mainly.

ereslibre:
More context: I'm keeping the original `pre-` steps on their original folders that apply to all updates (e.g. `kubelet.update-post-start-services`) regardless of specific migrations (in this case we always want to uncordon the node when it comes back and we cordoned it).

What goes in the `migrations/v1-v2` folder are specific migration logic that can be removed on future versions. I found far easier to reason with this structure when checking the whole update logic.

ereslibre:
The upgrade went very smoothly. At all times I could see `Ready` nodes (the old ones -- able to contact the new masters, and the new ones when they were updated one by one):

```
NAME                                                 STATUS                        ROLES     AGE       VERSION
59a524dffa4d4189bdac5a5ce10ed640.infra.caasp.local   NotReady,SchedulingDisabled   <none>    54m       v1.8.9
90669b90cfd04db5826da66e3ef0dad7.infra.caasp.local   Ready                         <none>    53m       v1.8.9
b63f313481e74f92b8ff0aa66ed9f19b.infra.caasp.local   NotReady,SchedulingDisabled   <none>    53m       v1.8.9
d130787c801d4c4f8d9d851ac11a5f41.infra.caasp.local   Ready                         <none>    53m       v1.8.9
e58fd682b8a4471db237e0fc7992e2f4.infra.caasp.local   Ready                         <none>    53m       v1.8.9
master-0                                             Ready,SchedulingDisabled      <none>    8m        v1.9.8
worker-0                                             NotReady,SchedulingDisabled   <none>    8m        v1.9.8
worker-1                                             NotReady                      <none>    8m
worker-2                                             NotReady                      <none>    8m
worker-3                                             NotReady                      <none>    8m
worker-4                                             Ready                         <none>    8m        v1.9.8
```

```
NAME                                                 STATUS                        ROLES     AGE       VERSION
59a524dffa4d4189bdac5a5ce10ed640.infra.caasp.local   NotReady,SchedulingDisabled   <none>    57m       v1.8.9
d130787c801d4c4f8d9d851ac11a5f41.infra.caasp.local   Ready                         <none>    56m       v1.8.9
e58fd682b8a4471db237e0fc7992e2f4.infra.caasp.local   Ready,SchedulingDisabled      <none>    56m       v1.8.9
master-0                                             Ready,SchedulingDisabled      <none>    12m       v1.9.8
worker-0                                             Ready                         <none>    12m       v1.9.8
worker-1                                             NotReady                      <none>    12m
worker-2                                             Ready                         <none>    12m       v1.9.8
worker-3                                             NotReady                      <none>    12m
worker-4                                             Ready                         <none>    12m       v1.9.8
```

However, when all machines were done and we were waiting for deployments, the `master` never got removed from the list and never was uncordoned. I'm looking at this at the moment; when this gets fixed I think the upgrade process should be stable:

```
NAME                                                 STATUS                        ROLES     AGE       VERSION
59a524dffa4d4189bdac5a5ce10ed640.infra.caasp.local   NotReady,SchedulingDisabled   <none>    1h        v1.8.9
master-0                                             Ready,SchedulingDisabled      master    20m       v1.9.8
worker-0                                             Ready                         <none>    20m       v1.9.8
worker-1                                             Ready                         <none>    20m       v1.9.8
worker-2                                             Ready                         <none>    20m       v1.9.8
worker-3                                             Ready                         <none>    20m       v1.9.8
worker-4                                             Ready                         <none>    20m       v1.9.8
```

salt/migrations/2-3/haproxy/init.sls
```diff
@@ -0,0 +1,65 @@
...
+
+# Send a SIGTERM to haproxy when the config changes
+# TODO: There should be a better way to handle this, but currently, there is not. See
+# kubernetes/kubernetes#24957
+haproxy_restart:
+  cmd.run:
+    - name: |-
+        haproxy_id=$(docker ps | grep -E "k8s_haproxy.*\.{{ pillar['internal_infra_domain'] | replace(".", "\.") }}_kube-system_" | awk '{print $1}')
+        if [ -n "$haproxy_id" ]; then
+            docker kill $haproxy_id
```
flavio: This is not going to work on clusters deployed with CRI-O, we should use `crictl stop` instead. Also the code that gets the container ID should be extended to handle CRI-O. We should have all the needed pieces inside of our `caasp_cri` salt module. If it's not there we should take a look at the file history, I'm pretty sure I wrote that...

salt/migrations/2-3/etcd/pre-reboot.sls
```diff
@@ -0,0 +1,24 @@
...
+{% set has_etcd_role = ("etcd" in roles) %}
+
+{% if not has_etcd_role %}
+  # make sure there is nothing left in /var/lib/etcd
+
+cleanup-old-etcd-stuff:
+  cmd.run:
+    - name: rm -rf /var/lib/etcd/*
+
+uninstall-etcd:
```
flavio: I think this is misleading, can you rename it to something like `disable-etcd`?

salt/migrations/2-3/etcd/pre-reboot.sls
```diff
@@ -0,0 +1,24 @@
+{% set roles = salt['grains.get']('roles', []) %}
+{% set has_etcd_role = ("etcd" in roles) %}
+
+{% if not has_etcd_role %}
+  # make sure there is nothing left in /var/lib/etcd
+
+cleanup-old-etcd-stuff:
+  cmd.run:
+    - name: rm -rf /var/lib/etcd/*
```
flavio: Wouldn't it be more *"salty"* to use a `salt.states.file.absent`?

salt/migrations/2-3/haproxy/init.sls
```diff
@@ -0,0 +1,65 @@
...
+         pillar['ssl']['kube_apiserver_proxy_crt'],
+         pillar['ssl']['kube_apiserver_proxy_key'],
+         cn = grains['nodename'] + '-proxy',
+         o = pillar['certificate_information']['subject_properties']['O'],
+         extra_alt_names = alt_master_names()) }}
+
+haproxy:
+  file.managed:
+    - name: /etc/kubernetes/manifests/haproxy.yaml
+    - source: salt://haproxy/haproxy.yaml.jinja
```
flavio: Here we are using the `haproxy.yaml` file of the "main/upstream" HAProxy status on v2 nodes. Wouldn't it be safer to copy the current version into this  `salt/migrations/2-3/haproxy/` directory to avoid breakages in the future?

salt/migrations/2-3/haproxy/init.sls
ereslibre: This is not going to run on clusters with cri-o. We need this state so it can be applied on 2.0 versions:

- Masters are updated
- Deploy new haproxy configuration to all workers (all still running 2.0)
- Old workers see the masters
- We can continue with the upgrade process on the workers (draining them etc...)

salt/migrations/2-3/etcd/pre-reboot.sls
ereslibre: As far as I know there's no way to replicate this exact command, `file.absent` will do a `rm -rf /var/lib/etcd/`, removing the `etcd` folder too.

salt/migrations/2-3/etcd/pre-reboot.sls
ereslibre: Changing.

salt/migrations/2-3/haproxy/init.sls
flavio: I see, all good then!

salt/migrations/2-3/etcd/pre-reboot.sls
flavio: :+1:

ereslibre:
Updated the patch to also address the issue commented on https://github.com/kubic-project/salt/pull/626#issuecomment-403012348

flavio:
@inercia can you take a look please?

ereslibre:
I'm still testing. I think it's worth handing this over to QA once I have finished my tests so they can confirm that problems are gone and after that we can merge if you are fine with the change.

salt/orch/update.sls
```diff
@@ -265,25 +286,46 @@ early-services-setup:
...
+# will) be in `NotReady` state. This means that any operation that we perform
+# that go through the apiserver down to the kubelets won't work (e.g. draining
+# nodes).
+#
+# To fix this problem we'll apply the haproxy sls to all worker nodes, so they
+# can rejoin the cluster and we can operate on them normally.
+all-workers-2.0-pre-clean-shutdown:
   salt.state:
-    - tgt: '{{ is_master_tgt }}'
+    - tgt: '( {{ is_updateable_worker_tgt }} ) and G@osrelease:2.0'
```
inercia: In case we have this grain `G@osrelease:2.0`, maybe we could protect all the blocks were we run `migration-2-3-*`...

salt/orch/update.sls
```diff
@@ -265,25 +286,46 @@ early-services-setup:
...
+# Sanity check. If an operator manually rebooted a machine when it had the 3.0
+# snapshot ready, we are already in 3.0 but with an unapplied haproxy config.
+# Apply the main haproxy sls to 3.0 workers (if any).
+all-workers-3.0-pre-clean-shutdown:
+  salt.state:
+    - tgt: '( {{ is_updateable_worker_tgt }} ) and G@osrelease:3.0'
+    - tgt_type: compound
+    - sls:
+        - etc-hosts
+        - haproxy
```
inercia: Is this `tgt` matching anything? This `tgt` means _"all the 3.0 machines were we have installed updated but they still need to be rebooted"_, and I think the `tx_update_reboot_needed` is cleaned up on reboot by some `service`, right?. So I think it will not really match anything...

Anyway, I think in this kind of scenarios it is better to just highstate instead of running some key (and maybe conflicting) SLSs.

salt/orch/update.sls
```diff
@@ -116,13 +117,13 @@ admin-setup:
 # with the real update.
 pre-orchestration-migration:
   salt.state:
-    - tgt: '{{ is_regular_node_tgt }}'
+    - tgt: '{{ is_updateable_node_tgt }}'
```
inercia: Note that there is a change in the target for these migrations, from _"all the master and minion machines in the cluster"_ to _"all the **updateable** master and minion machines in the cluster"_. Not sure about the consequences of this change.

salt/orch/update.sls
```diff
@@ -294,9 +336,10 @@ all-masters-post-start-services:
   salt.state:
     - tgt: '{{ worker_id }}'
     - sls:
+      - migrations.2-3.kubelet.stop
```
inercia: I miss some more granularity in the versions numbers for migrations. I think it would be more future-proof to have something like `migrations.2.0-3.0.kubelet.stop`.

By doing that, maybe we could do in the future some more flexible migrations with something like:

```SaltStack
{%- set worker_id_osrelease = salt.caasp_grains.get(worker_id, 'osrelease')[worker_id] %}

{{ worker_id }}-early-clean-shutdown:
   salt.state:
     - tgt: '{{ worker_id }}'
     - sls:
      - migrations.{{ worker_id_osrelease }}-3.0.kubelet.stop
   ...
```

that would choose the right migration depending on the current CaaSP version.

salt/orch/update.sls
ereslibre: I only protected the one that would certainly not work, new haproxy on old system will break, as will do an old haproxy on a new system.

However, being on a 3.0 snapshot doesn't mean that we should not run a migration yet (e.g. if an update failed and we are retrying it, we still want to rename nodes, delete old node names, deploy cni...).

That is why I only explicitly set `G@osrelease:2.0` with the old `haproxy` and `G@osrelease:3.0` with the new haproxy.

salt/orch/update.sls
ereslibre: `tx_update_reboot_needed` is only cleaned by a script in the admin node, in the rest of the machines it's still there until we remove it with salt (that is, when the upgrade on that machine is known to have succeeded).

This is a special case: imagine a machine had downloaded the 3.0 new snapshot and didn't apply it yet. The operator for whatever reason reboots the machine, it will apply the new snapshot -- it's a corner case, weird to happen, but it was free to protect from. We just apply `haproxy` on all worker nodes (the old version on 2.0 and the new on 3.0 [if any]).

salt/orch/update.sls
ereslibre: Yes, this was a side effect discovered by @kiall. It's also a rather artificial case, but it boils down to: only rename nodes that are going to be part of the update process. Otherwise you'll end up with a `kubectl get nodes` listing that could have `NotReady` new entries that weren't going to be migrated in any moment.

The way to reproduce as I say is pretty artificial (e.g. you run transactional-update only on the masters, and apply the upgrade, all machines will get renamed without this change).

salt/orch/update.sls
ereslibre: Despite I like the idea we cannot do like that (as the dot `.` is the path separator) -- your example would go to look into `migrations/2/0-3/0/kubelet/stop.sls`.

I think the best we could achieve would be something like `migrations.2-0-3-0.kubelet.stop`, but it's a bit weird, what do you think?

flavio: Oh my... what @inercia proposed was really nice, too bad we cannot do it :man_facepalming: 

I don't know what feels less weird between the two proposals @ereslibre made, IMO I would go with `migrations.2-0-3-0.kubelet.stop`

ereslibre: Yes, with `major-minor` scheme we could transform the `osrelease` grain replacing dots by dashes and we could have the same feature.

I will update the PR.

ereslibre:
The migration procedure is working really great now. I didn't test what happens with an active ceph mount on the workers yet.

salt/orch/update.sls
inercia: Ok, I see.

salt/orch/update.sls
inercia: We can create a `get()` in `_module/caas_osrelease.py` with something like

```Python
from caasp_grains import get as get_grain

# the separator for versions, like 2_1 insstead of 2.1
VERSION_SEP = '_'

def get(host):
    return str(get_grain(host, 'osrelease')[host]).replace('.'. VERSION_SEP)
```

so we could just do:

```SaltStack
{{ worker_id }}-early-clean-shutdown:
   salt.state:
     - tgt: '{{ worker_id }}'
     - sls:
      - migrations.{{ salt.caasp_osrelease.get(worker_id) }}-3_0.kubelet.stop
```

then we should have a `migrations.2_1-3_0.kubelet.stop` file

salt/orch/update.sls
inercia: Ok, this looks like another undocumented, magic, corner case... (facepalm)

salt/orch/update.sls
inercia: The thing is that we are introducing many `migrations` and I don't like them to be unconditionally executed: they are just run when the update orchestration is run. In some of them we are not effectively checking if the migration must be applied or not (I'm thinking for example on `etcd`). And I don't think those checks should be done individually in the `migration-*` files either.

So my personal take on this would be to find a general solution at the orchestration level that would run `migration-x-y`  _iff_ the node is at `x` and it is being updated to `y`.

ereslibre: I think this logic is very probably out of the scope for this PR. I believe we should do something similar to what you suggest, but not right now IMHO (in this PR). What I did was to move all migration logic from code mixed with the normal states into a specific `migration` folder, nothing else.

In any case, I have been thinking a bit about this, so we have two kinds of migrations:

* Cluster-wide migrations
  * We cannot know if we ran a specific migration that affects the whole cluster (e.g. deploying PSP, or CNI new manifests).
    * We would need a mechanism where we could save this information (e.g. callbacks to a central piece like Velum that saves cluster-wide migrations), or
    * Always run the cluster-wide migrations; making them idempotent from the very beginning (as they are now).

* Node specific migrations
  * Migrations that only affect the current node.
  * Relying on `osrelease` is truly problematic and I'll detail why:
    * When `transactional-update` creates a new snapshot, it sets the default btrfs snapshot to the new one: https://github.com/openSUSE/transactional-update/blob/2b46aa83d6ec39e050a67876c5618ea11a4df27e/sbin/transactional-update.in#L956.
    * This means that an "innocent" reboot of a machine could upgrade it (loading a new snapshot) and we could skip migrations on that machine (because its `osrelease` is not what we expect to be, and we skip the migrations on that one).
  * Migrations could be identified with some string or hash and we could save the migrations that have been executed in the machine's grains as a list.
    * Ala-rails, we run the migrations that are not present in the machine list of migrations run. This is absolutely snapshot agnostic; either you have run a migration or not; if not: do it.
  * Aside from migrations, I think it would be good that we control when we `btrfs set-default` a new snapshot. If we could add a new flag to `transactional-update` that skips the call to `btrfs set-default` when it creates a new snapshot, then we could, when upgrading:
    * Get machines with `tx_update_reboot_needed` flag
    * For each machine, we run `btrfs subvolume set-default $BTRFS_ID ${SNAPSHOT_DIR}`
    * We reboot the machines as we do right now

The main difference is that an "innocent" reboot won't apply new snapshots randomly on the cluster, but we'll apply the new snapshots when the customer hits "Upgrade cluster" and they really want the cluster to be upgraded.

So, summing up: I see a combination of cluster-wide migrations and node-specific migrations. I think all of them should be idempotent just in case, as much as possible. We could do a "migrations-like" feature that only runs node migrations once per node. Also, we could be the ones applying the new btrfs snapshots instead of being `transactional-update` the one activating it automatically [making an "innocent" reboot issued by an operator load a new version without our control/knowledge].

salt/orch/update.sls
ereslibre: I wrote some thoughts about migrations here: https://github.com/kubic-project/salt/pull/626#discussion_r200917311

ellisab:
worked with @ereslibre and draining works (tested on v3).
For v2 to v3 upgrade we only drain using the new names so that's why draining is not working. We ned to drain for both old and new names

PR #630, opened 2018-07-16
Some configuration options in the API server

inercia:
Set things like the runtime config and admission control in the pillar. Minor cleanups.

feature#cleanups

### DIFF
--- a/salt/kube-apiserver/apiserver.jinja
+++ b/salt/kube-apiserver/apiserver.jinja
@@ -4,8 +4,12 @@
 # The following values are used to configure the kube-apiserver
 #
 
-{% set cloud_provider = salt.caasp_pillar.get('cloud:provider') %}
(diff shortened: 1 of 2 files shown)


salt/kube-apiserver/apiserver.jinja
```diff
@@ -59,7 +59,9 @@ KUBE_API_ARGS="--advertise-address={{ salt.caasp_net.get_primary_ip() }} \
                --storage-media-type=application/json \
                --service-account-key-file={{ pillar['paths']['service_account_key'] }} \
                --service-account-lookup=true \
-               --runtime-config=admissionregistration.k8s.io/v1alpha1 \
+{%- for cfg in pillar['kubernetes']['runtime_configs'] %}
+               --runtime-config={{ cfg }} \
```
MaximilianMeister: shouldn't there be a list in just one flag?

```jinja
--runtime-config=cfg[0]=true,cfg[1]=true,cfg[2]=true ...
```

inercia: I think both forms are acceptable. Looking at [some other usages in GitHub](https://github.com/search?q=--runtime-config&type=Code), I see both lists as well as repeated parameters...

PR #655, opened 2018-08-24
Fix bsc#1105910 CAdvisor is publicly exposed on the kubernetes nodes(:::4194)

lcavajani:
This PR disables CAdvisor in kubelet.


lcavajani:
CHANGELOG for Kubernetes 1.10: The kubelet's cadvisor port has been deprecated. The default will change to 0 (disabled) in 1.12, and the cadvisor port will be removed entirely in 1.13.

PR #654, opened 2018-08-23
channel migration

bergmannf:
This orchestration will run transactional-update salt migration to change
the cluster to a new channel.

PR #658, opened 2018-08-27
[3.0] channel migration

bergmannf:
Backport of #654

### DIFF
--- a/salt/orch/prepare-product-migration.sls
+++ b/salt/orch/prepare-product-migration.sls
@@ -1,0 +1,94 @@
+{#- Get a list of nodes seem to be down or unresponsive #}
+{#- This sends a "are you still there?" message to all #}
+{#- the nodes and wait for a response, so it takes some time. #}
+
+{%- set default_batch = salt['pillar.get']('default_batch', 5) %}
+{#- tx_update_migration_available needs to be last, as it is still used as a selector til reboot #}
+{%- set migration_grains = ['tx_update_migration_notes', 'tx_update_migration_newversion', 'migration_in_progress', 'tx_update_migration_available'] %}
+
+{%- set nodes_down = salt.saltutil.runner('manage.down') %}
+{%- if nodes_down|length >= 1 %}
+# {{ nodes_down|join(',') }} seem to be down: skipping
+  {%- do salt.caasp_log.debug('CaaS: nodes "%s" seem to be down: ignored', nodes_down|join(',')) %}
+  {%- set is_responsive_node_tgt = 'not L@' + nodes_down|join(',') %}
+{%- else %}
+# all nodes seem to be up
+  {%- do salt.caasp_log.debug('CaaS: all nodes seem to be up') %}
+  {#- we cannot leave this empty (it would produce many " and <empty>" targets) #}
+  {%- set is_responsive_node_tgt = '*' %}
+{%- endif %}
+
+{%- set is_migrateable = is_responsive_node_tgt + ' and G@tx_update_migration_available:true' %}
+
+{%- set is_migrateable_admin = 'G@tx_update_migration_available:true and G@roles:admin' %}
+
+set-progress-grain:
+  salt.function:
+    - tgt: '{{ is_migrateable_admin }}'
+    - tgt_type: grains
+    - tgt_type: compound
+    - name: grains.setval
+    - arg:
+      - migration_in_progress
+      - true
+
+# We have to disable the transactional-update.timer to prevent interference with
+# our update. The timer could create a newer snapshot otherwise and break the
+# update after a reboot by booting into the assumed newer snapshot that was
+# created after our migration.
+disable-transactional-update-timer:
+  salt.function:
+    - tgt: '{{ is_migrateable }}'
+    - tgt_type: compound
+    - batch: {{ default_batch }}
+    - name: service.disable
+    - arg:
+        - transactional-update.timer
+
+run-transactional-migration:
+  salt.state:
+    - tgt: '{{ is_migrateable }}'
+    - tgt_type: compound
+    - batch: {{ default_batch }}
+    - sls:
+      - transactional-update.migration
+    - require:
+        - disable-transactional-update-timer
+
+update-data:
+  salt.function:
+    - tgt: '{{ is_migrateable }}'
+    - tgt_type: compound
+    - names:
+      - saltutil.refresh_pillar
+      - saltutil.refresh_grains
+    - require:
+        - run-transactional-migration
+
+{%- for grain in migration_grains %}
+unset-admin-grain-{{ grain }}:
+  salt.function:
+    - name: grains.delval
+    - tgt: '{{ is_migrateable_admin }}'
+    - tgt_type: compound
+    - arg:
+        - {{ grain }}
+    - kwarg:
+        destructive: true
+    - require:
+        - update-data
+{%- endfor %}
+
+reboot-admin:
+  salt.function:
+    - tgt: 'G@roles:admin'
+    - tgt_type: compound
+    - name: cmd.run
+    - arg:
+        - sleep 15; systemctl reboot
+    - kwarg:
+        bg: true
+    - require:
+{%- for grain in migration_grains %}
+        - unset-admin-grain-{{ grain }}
+{%- endfor %}
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -29,19 +29,29 @@
 {%- set is_updateable_master_tgt = is_updateable_tgt + ' and ' + is_master_tgt %}
 {%- set is_updateable_worker_tgt = is_updateable_tgt + ' and ' + is_worker_tgt %}
 {%- set is_updateable_node_tgt   = '( ' + is_updateable_master_tgt + ' ) or ( ' + is_updateable_worker_tgt + ' )' %}
-
 {%- set all_masters = salt.saltutil.runner('mine.get', tgt=is_master_tgt, fun='network.interfaces', tgt_type='compound').keys() %}
 {%- set super_master = all_masters|first %}
 
+{%- set is_migration = salt['pillar.get']('migration', false) %}
+{%- if is_migration %}
+  {%- set progress_grain = "migration_in_progress" %}
+{%- else %}
+  {%- set progress_grain = "update_in_progress" %}
+{%- endif %}
+
+{% if is_migration %}
+{%- set migrated_nodes_tgt = is_responsive_node_tgt + ' and G@migration_in_progress:true' %}
+{% endif %}
+
 # Ensure all nodes with updates are marked as upgrading. This will reduce the time window in which
 # the update-etc-hosts orchestration can run in between machine restarts.
-set-update-grain:
+set-progress-grain:
   salt.function:
     - tgt: '( {{ is_regular_node_tgt }} and {{ is_updateable_tgt }} ) or {{ super_master }}'
     - tgt_type: compound
     - name: grains.setval
     - arg:
-      - update_in_progress
+      - {{ progress_grain }}
       - true
 
 # this will load the _pillars/velum.py on the master
@@ -49,7 +59,7 @@ sync-pillar:
   salt.runner:
     - name: saltutil.sync_pillar
     - require:
-      - set-update-grain
+      - set-progress-grain
 
 update-data:
   salt.function:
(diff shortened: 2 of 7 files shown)


salt/orch/update.sls
```diff
@@ -29,10 +29,15 @@
...
 {%- set is_updateable_worker_tgt = is_updateable_tgt + ' and ' + is_worker_tgt %}
 {%- set is_updateable_node_tgt   = '( ' + is_updateable_master_tgt + ' ) or ( ' + is_updateable_worker_tgt + ' )' %}
-
 {%- set all_masters = salt.saltutil.runner('mine.get', tgt=is_master_tgt, fun='network.interfaces', tgt_type='compound').keys() %}
 {%- set super_master = all_masters|first %}
 
+{%- set is_migration = salt['pillar.get']('migration', false) %}
+
+{% if is_migration %}
+{%- set migrated_nodes = is_responsive_node_tgt + 'and G@migration_in_progress:true' %}
```
MaximilianMeister: i think there is a whitespace missing? ' and G@migration_in_progress:true'

error from salt master: `Invalid compound target: *and G@migration_in_progress:true`

salt/orch/migration.sls
```diff
@@ -0,0 +1,85 @@
+{#- Get a list of nodes seem to be down or unresponsive #}
```
inercia: `migration.sls` gives the impression this orchestration is  general cluster migration orochestration, while is it focused on migrating just the `transactional-update` stuff. I would rename this file, so it is more clear this only affects the `transactional-update`...

salt/orch/migration.sls
```diff
@@ -0,0 +1,85 @@
...
+  {%- do salt.caasp_log.debug('CaaS: all nodes seem to be up') %}
+  {#- we cannot leave this empty (it would produce many " and <empty>" targets) #}
+  {%- set is_responsive_node_tgt = '*' %}
+{%- endif %}
+
+{%- set is_migratable = is_responsive_node_tgt + ' and G@tx_update_migration_available:true' %}
+
+{%- set is_admin = is_migratable + ' and G@roles:admin' %}
+
+{%- set nodes_up = salt.saltutil.runner('manage.up')|reject("equalto", "ca") %}
```
inercia: Having a `nodes_up` and a `nodes_down` is a waste of resources, as each `manage.up` involves sending a `ping` to all the machines in the cluster, waiting for the response, etc. You can build `nodes_up` as the _inverse_ of `nodes_down`...

salt/orch/migration.sls
```diff
@@ -0,0 +1,85 @@
+{#- Get a list of nodes seem to be down or unresponsive #}
+{#- This sends a "are you still there?" message to all #}
+{#- the nodes and wait for a response, so it takes some time. #}
+
+{%- set parallelism = salt['pillar.get']('parallelism', 3) %}
```
inercia: I would use the same `default_batch` variable we have in `orch/kubernetes.sls`. Maybe it could be updated in both places and be obtained from the same pillar variable...

salt/orch/update.sls
```diff
@@ -555,3 +560,29 @@ remove-update-grain:
...
+unset-migration-grains:
+  salt.function:
+    - tgt: '{{ migrated_nodes }}'
+    - tgt_type: compound
+    - name: grains.delval
+    - arg:
+        - tx_update_migration_available
+        - tx_update_migration_notes
+        - tx_update_migration_mirror_synced
+        - migration_in_progress
```
inercia: I don't see where all these grains are set...

MaximilianMeister: they're eventually set by this service: https://github.com/thkukuk/update-checker

salt/orch/migration.sls
MaximilianMeister: hmm not sure what you mean with "migrating just the transactional-update stuff". this orch is the actual migration (installation) from version to next version. and the tool to do it is `transactional-update`

salt/orch/migration.sls
MaximilianMeister: updated

salt/orch/migration.sls
inercia: then I don't understand what is the difference between the `migration` and the `update` orchestrations...

bergmannf: I think I understand @inercia point, because the `orchestration` doesn't actually *do* the migration, but just prepares the node to be ready to take the correct steps to do the actual upgrade:

1. Admin - only needs a reboot

2. Cluster nodes -  need to use the update orchestration

Maybe we should rename it to something like `migration_bootstrap` or `migration_preparation` instead?

inercia: Yes, I wouldn't use a generic name for something that is not really generic (it seems more like a corner case or a previous step necessary for doing other things)...

MaximilianMeister: renamed to `prepare-product-migration`

salt/haproxy/init.sls
```diff
@@ -69,6 +69,7 @@ haproxy:
       - {{ pillar['api']['ssl_port'] }}
     - proto:      tcp
 
+{% if not salt['grains.get']('tx_update_migration_available', false) %}
```
MaximilianMeister: @inercia the node migration (through update orch) keeps failing here, so this exclusion is a bit of a shot in the dark. do you maybe have an idea why it could fail? for some more context, see the ML: "[caasp-devel] cri not reacting during update orchestration"

salt/orch/prepare-product-migration.sls
```diff
@@ -0,0 +1,104 @@
...
+{%- else %}
+# all nodes seem to be up
+  {%- do salt.caasp_log.debug('CaaS: all nodes seem to be up') %}
+  {#- we cannot leave this empty (it would produce many " and <empty>" targets) #}
+  {%- set is_responsive_node_tgt = '*' %}
+{%- endif %}
+
+{%- set is_migrateable = is_responsive_node_tgt + ' and G@tx_update_migration_available:true' %}
+
+{%- set is_admin = is_migrateable + ' and G@roles:admin' %}
```
inercia: Not sure about this. Imagine a node with id `XXXX` is down, so:

```Python
is_responsive_node_tgt = 'not L@' + nodes_down|join(',') = 'not L@XXXX'
is_migrateable = is_responsive_node_tgt + ' and G@tx_update_migration_available:true' =  \
  'not L@XXXX and G@tx_update_migration_available:true'
is_admin = is_migrateable + ' and G@roles:admin' = \
  'not L@XXXX and G@tx_update_migration_available:true and G@roles:admin'
```

So, at the end, the only important part is `G@tx_update_migration_available:true and G@roles:admin`, right? In conclusion:

* I would use `set is_admin = 'G@tx_update_migration_available:true and G@roles:admin'`
* I would fail with an exception if `ca in nodes_down` (because I guess we should not consider that case).

And, as a suggestion, I would use `is_migrateable_admin`, so it clearly states that this target is about a migrateable admin and not just the `admin`.

salt/orch/prepare-product-migration.sls
```diff
@@ -0,0 +1,104 @@
...
+        - {{ grain }}
+    - kwarg:
+        destructive: true
+    - require:
+        - update-data
+{%- endfor %}
+
+reboot-admin:
+  salt.function:
+    - tgt: 'G@roles:admin'
```
inercia: Why do you use `G@roles:admin` here instead of `{{ is_admin }}`?

MaximilianMeister: yeah, this is due to the bad naming you mentioned. at this point the target is not correct anymore as the minion doesnt have the `tx_update_migration_available` grain anymore. so we just want to target the admin role

salt/orch/prepare-product-migration.sls
MaximilianMeister: > So, at the end, the only important part is `G@tx_update_migration_available:true and G@roles:admin`, right?

i think so, because we're only handling the admin in this variable, and if it would be down we wouldn't even be able to talk to the salt master IIUC

MaximilianMeister:
@inercia updated to your suggestion, please have a look if that makes sense to you and i'll test it again in the meantime

salt/orch/prepare-product-migration.sls
```diff
@@ -0,0 +1,104 @@
+{#- Get a list of nodes seem to be down or unresponsive #}
+{#- This sends a "are you still there?" message to all #}
+{#- the nodes and wait for a response, so it takes some time. #}
+
+{%- set default_batch = salt['pillar.get']('default_batch', 5) %}
+{#- tx_update_migration_available needs to be last, as it is still used as a selector til reboot #}
+{%- set migration_grains = ['tx_update_migration_notes', 'tx_update_migration_newversion', 'migration_in_progress', 'tx_update_migration_available'] %}
+
+{%- set nodes_down = salt.saltutil.runner('manage.down') %}
+{%- if nodes_down|length >= 1 %}
```
inercia: Minor comment: I think you can replace `if nodes_down|length >= 1` by `if nodes_down`...

salt/orch/prepare-product-migration.sls
```diff
@@ -0,0 +1,104 @@
...
+{%- set is_migrateable_admin = 'G@tx_update_migration_available:true and G@roles:admin' %}
+
+{%- set nodes_up = salt.saltutil.runner('manage.up')|reject("equalto", "ca") %}
+
+{% for node in nodes_up %}
+  # Maybe using the cache here might lead to problems, if it is out of date.
+  {%- set migration_grain = salt['saltutil.runner']('cache.grains', tgt=node)[node].get('tx_update_migration_available', false) %}
+  {%- do salt.caasp_log.debug('Checking {0} for migration: {1}'.format(node, migration_grain)) %}
+  {% if not migration_grain %}
+    {% do salt.test.exception('Not all nodes are migratable: {0} is missing'.format(node)) %}
```
inercia: So all the nodes that are up must be _migrateable_, or otherwise the migration will be aborted, right?. Imagine the following scenario:

1. `nodes_up=[1,2,3,4]` and `nodes_down=[666]`]
2. the user runs this orchestration
3. `nodes_up=[1,2,3,4]` are not migrateable anymore
4. `666` comes back to life
5. the user wants to run the migration again, for migrating `666`
6. the mix of migrateable (`666`) and non-mirateable (`[1,2,3,4]`) nodes abort the migration

I guess the same would happen with new nodes, right?

bergmannf: @inercia True - I think bigger users often have nodes offline, so that will very likely happen.

In that case we could just switch `salt.test.exception` with `salt.caasp_log.info` to have some information in the logs about nodes missing the `migration` grain, and just proceed with the others.

MaximilianMeister: agree, right now in the update handling it's similar i think. if a single node hasn't been updated yet, you can still trigger that later

salt/orch/prepare-product-migration.sls
MaximilianMeister: are you sure that this will work? havben't tested it

inercia: I guess it should work, but don't worry...

salt/orch/prepare-product-migration.sls
```diff
@@ -0,0 +1,104 @@
...
+{%- set nodes_up = salt.saltutil.runner('manage.up')|reject("equalto", "ca") %}
+
+{% for node in nodes_up %}
+  # Maybe using the cache here might lead to problems, if it is out of date.
+  {%- set migration_grain = salt['saltutil.runner']('cache.grains', tgt=node)[node].get('tx_update_migration_available', false) %}
+  {%- do salt.caasp_log.debug('Checking {0} for migration: {1}'.format(node, migration_grain)) %}
+  {% if not migration_grain %}
+    {% do salt.caasp_log.warn('Not all nodes are migratable: {0} is missing'.format(node)) %}
+  {% endif %}
+{% endfor %}
```
inercia: I would remove this block. Doing all of this just for maybe a `salt.caasp_log.warn` is an overkill...

MaximilianMeister: dropped ;-)

MaximilianMeister:
I've added 2 small enhancements:

* let orch finish before rebooting the admin
* make transactional-update really fail salt, because piping the output into `tee` will always create a 0 returncode

MaximilianMeister:
> re-running CI

@ereslibre all `release-3.0` runs are failing ATM because of `kubernetes 1.10.8` issues

MaximilianMeister:
rebased

jordimassaguerpla:
Can we have an approval on this one?

jordimassaguerpla:
Hi. Sorry I have to close this one. We are removing this branch and since we are not going to need this feature in 3.0, I am not merging it. Feel free to reopen against the maintenance branch if this is needed.

PR #660, opened 2018-09-06
[3.0] Allow `etcd` to grow as required and shrink to optimal etcd cluster sizes on corner cases

ereslibre:
- [x] Fix documentation everywhere
- [x] Add test cases

Improve `etcd` configuration handling to allow it to grow as needed. This
change includes:

* Adding several masters at the same time
  * `etcd` will grow instance by instance still, as recommended by the `etcd` administration best practices.

* Try to use the current endpoints reported by `etcd`. This makes much
  easier to grow several instances one by one without having to relay
  on internal hacks to properly set up `ETCD_INITIAL_CLUSTER` environment
  variable.

* Add helper methods that allow us to list current members (active and
  unstarted)

* Differentiate between the first bootstrap (`ETCD_INITIAL_CLUSTER_STATE`
  defaults to `new`) and *any* other run, where `ETCD_INITIAL_CLUSTER_STATE`
  will be `existing`, as the `etcd` cluster is already running.

When we grow, we take into account the golden ratio; however, when shrinking
the cluster we don't. It might happen that a cluster ends up with not
recommended etcd number of nodes (2, 4, 6...) depending on how it grew before
and how it shrank.

This logic makes sure that we are always on an etcd golden ratio, also
on corner cases when removing nodes.

Fixes: bsc#1098433
Fixes: bsc#1098064
Fixes: bsc#1098161

Backport of https://github.com/kubic-project/salt/pull/614

### DIFF
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -40,7 +40,7 @@ api:
   # port for listening for SSL connections (kube-api)
   int_ssl_port:   '6444'
   # etcd storage backend version for cluster
-  etcd_version:   'etcd3'
+  etcd_version:   'etcd2'
 
   server:
     # should be set from the UI
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -40,7 +40,7 @@ api:
   # port for listening for SSL connections (kube-api)
   int_ssl_port:   '6444'
   # etcd storage backend version for cluster
-  etcd_version:   'etcd2'
+  etcd_version:   'etcd3'
 
   server:
     # should be set from the UI
--- a/salt/_modules/caasp_etcd.py
+++ b/salt/_modules/caasp_etcd.py
@@ -21,7 +21,7 @@ class NoEtcdServersException(Exception):
 
 
 def api_version():
-    return __salt__['pillar.get']('etcd_version', 'etcd3')
+    return __salt__['pillar.get']('etcd_version', 'etcd2')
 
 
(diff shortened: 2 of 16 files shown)


ereslibre:
[Allow `etcd` to grow as required and shrink to optimal etcd cluster sizes on corner cases](https://trello.com/c/RgWLIma2/1161-allow-etcd-to-grow-as-required-and-shrink-to-optimal-etcd-cluster-sizes-on-corner-cases)

ellisab:
Tested on caaspv3 with and without maintenance updates. Patch looks good. Testing plan available here:
 [Allow `etcd` to grow as required and shrink to optimal etcd cluster sizes on corner cases](https://trello.com/c/RgWLIma2/1161-allow-etcd-to-grow-as-required-and-shrink-to-optimal-etcd-cluster-sizes-on-corner-cases)

jordimassaguerpla:
In the queue to be released

MaximilianMeister:
This patch might have introduced a bug, when running a normal update and when running a migration from 2 to 3 it seems that a duplicate member are added to etcd, which makes the cluster unhealthy. we will open a bug

bergmannf:
It seems that the duplicate member bug that @MaximilianMeister found is also that happens in this bugzilla: https://bugzilla.novell.com/show_bug.cgi?id=1115236

PR #663, opened 2018-09-25
Python 3/SLE-15 compatibility for salt states

bergmannf:
These changes make the the salt-states run on `sle15` using `python3` and keep working on `sle12` using `python2`.

Right now for `sle15` they hard-code images from registry.suse.de that have to be replaced with more stable ones, once the tagging strategy is finalized.

The PR also includes the changes from #616, as without those changes the cluster will break, once a node is rebooted.

### DIFF
--- a/pillar/cni.sls
+++ b/pillar/cni.sls
@@ -1,6 +1,6 @@
 # the flannel backend ('udp', 'vxlan', 'host-gw', etc)
 flannel:
-  image:          'sles12/flannel:0.9.1'
+  image:          '{{ salt.caasp_registry.base_image_url() }}/flannel:0.9.1'
   backend:        'vxlan'
   port:           '8472'    # UDP port to use for sending encapsulated packets. Defaults to kernel default, currently 8472.
   healthz_port:   '8471'    # TCP port used for flannel healthchecks
--- a/pillar/params.sls
+++ b/pillar/params.sls
@@ -198,4 +198,4 @@ ldap:
   mail_attribute: ''
 
 # infra container to use instead of downloading gcr.io/google_containers/pause
-pod_infra_container_image: sles12/pause:1.0.0
+pod_infra_container_image: {{ salt.caasp_registry.base_image_url() }}/pause:1.0.0
--- a/salt/_grains/caasp_nodename.py
+++ b/salt/_grains/caasp_nodename.py
@@ -1,5 +1,3 @@
-#!/usr/bin/python
-
 import platform
 
 import salt.grains.core
--- a/salt/_modules/caasp_etcd.py
+++ b/salt/_modules/caasp_etcd.py
@@ -75,7 +75,7 @@ def get_cluster_size(**kwargs):
             member_count = _optimal_etcd_number(increased_member_count)
 
             __utils__['caasp_log.warn']("etcd member count too low (%d), increasing to %d",
-                           num_masters, increased_member_count)
+                                        num_masters, increased_member_count)
 
             # TODO: go deeper and look for candidates in nodes with
             #       no role (as get_replacement_for_member() does)
@@ -86,7 +86,7 @@ def get_cluster_size(**kwargs):
 
         if member_count < MIN_RECOMMENDED_MEMBER_COUNT:
             __utils__['caasp_log.warn']("etcd member count too low (%d), consider increasing "
-                           "to %d", member_count, MIN_RECOMMENDED_MEMBER_COUNT)
+                                        "to %d", member_count, MIN_RECOMMENDED_MEMBER_COUNT)
 
     member_count = max(1, member_count)
     __utils__['caasp_log.debug']("using member count = %d", member_count)
@@ -124,7 +124,7 @@ def get_additional_etcd_members(num_wanted=None, **kwargs):
         return []
 
     __utils__['caasp_log.debug']('get_additional_etcd_members: curr:%d wanted:%d -> %d missing',
-                    num_current_etcd_members, num_wanted_etcd_members, num_additional_etcd_members)
+                                 num_current_etcd_members, num_wanted_etcd_members, num_additional_etcd_members)
 
     # Get a list of `num_additional_etcd_members` nodes that could be used
     # for running etcd. A valid node is a node that:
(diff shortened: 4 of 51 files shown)


salt/_grains/caasp_use_registry_images.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python
+
+import sys
+
+
+def caasp_use_registry_images():
+    """
+    Override core.hostname grain to provide lowercased items
```
inercia: I guess this description should be updated...

salt/_modules/caasp_etcd.py
```diff
@@ -172,10 +170,12 @@ def get_endpoints(with_id=False, skip_this=False, skip_removed=False, port=ETCD_
         member_endpoint = 'https://{}:{}'.format(name, port)
         if with_id:
             member_endpoint = "{}={}".format(node_id, member_endpoint)
-        etcd_members_lst.append(member_endpoint)
+            etcd_members_lst.append(member_endpoint)
+        else:
+            etcd_members_lst.append(member_endpoint)
```
inercia: I think this should be left as it was, as `etcd_members_lst.append(member_endpoint)` will be done anyway.

bergmannf: I just realized that it was already `dedented` one level - I must have run into some spurious error, because you are obviously right ...

salt/_grains/caasp_nodename.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
```
bergmannf: This one concerns myself a bit - I am not sure if it would make sense to symlink python3 to python if no python2 is installed.

Otherwise this will always flip-flop between working on `python2` or `3`, *if* it ever gets run as a script (which i doubt will happen), so we could probably also remove the whole shebang line.

pillar/cni.sls
```diff
@@ -1,6 +1,6 @@
 # the flannel backend ('udp', 'vxlan', 'host-gw', etc)
 flannel:
-  image:          'sles12/flannel:0.9.1'
+  image:          '{{ grains["base_image_url"] }}/flannel:0.9.1'
```
jordimassaguerpla: thanks

salt/haproxy/init.sls
```diff
@@ -83,7 +85,9 @@ haproxy-restart:
 {% if not salt.caasp_nodes.is_admin_node() %}
     - require:
       - service: kubelet
+      {%- if not salt.caasp_version.use_registry_images() %}
       - service: container-feeder
+      {%- endif %}
 {% endif %}
```
jordimassaguerpla: this looks good to me

jordimassaguerpla:
The changes regarding the container-feeder lgtm, as well as using a base_image variable.

salt/_grains/caasp_nodename.py
jordimassaguerpla: Do this script require python3? Or could it run with python2 as well?

inercia: I would just require `python3`, we don't have to be so flexible, right? So `env python3` seems right to me...

jordimassaguerpla: We are still building images with sles12sp3 with python2, so if possible we should be compatible so we don't break the current images.

pillar/cni.sls
ereslibre: Isn't a pillar more suitable for this than a grain? (regarding `base_image_url`). A grain is machine-oriented, but I think this setting is more cluster-wide, right?

pillar/params.sls
```diff
@@ -198,4 +198,4 @@ ldap:
   mail_attribute: ''
 
 # infra container to use instead of downloading gcr.io/google_containers/pause
-pod_infra_container_image: sles12/pause:1.0.0
+pod_infra_container_image: {{ grains['base_image_url'] }}/pause:1.0.0
```
ereslibre: ditto

salt/_modules/caasp_version.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def caasp_version():
+    # Python 3 does not allow comparing list & tuples, so force a tuple:
+    version = tuple(__salt__['grains.get']('osrelease_info', UNKNOWN_VERSION))
+    return version
+
+
+def use_registry_images():
+    return __salt__['grains.get']('use_registry_images', False)
```
ereslibre: Same comment on this one, if some machines will allow to `use_registry_images`, then it makes sense to have it as a grain, otherwise (if it's cluster-wide) I'd say having this information inside the pillar would be better.

pillar/cni.sls
bergmannf: I 100% agree with you that it *should* be a pillar.

The problem is that I couldn't get it to work, as I would have to access the pillar in other pillars, which seems not possible in salt.
I tried setting a variable as well `{% set some_pillar = 'some_value' %}` and use that instead, but that also doesn't work, so I resorted to using a grain as it was the only solution that worked.

If there is a way to use a pillar instead I would love to use it.

salt/_grains/caasp_use_registry_images.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python
+
+import sys
+
+
+def use_registry_images():
```
inercia: If these functions are not exported, I think it would be better to hide them with a `_`, like `_use_registry_images`. That way it would be clear that the only function that returns something useful is the `caasp_use_registry_images()`...

pillar/cni.sls
inercia: Yeah, that is a known problem with pillars: you cannot call a pillaar from a pillar. Maybe you could try to call a `_module/*` function... I don't know.

salt/_grains/caasp_nodename.py
bergmannf: I believe the same as @inercia - or they even just `import` it, which ignores the interpreter line completely. I'll just remove the line completely to avoid confusion.

bergmannf:
@ereslibre @inercia I moved the `registry_image` grain into a `module` now, given that a pillar is not possible.

PR #665, opened 2018-09-25
[3.0] Always wait for haproxy to be serving requests before continuing

ereslibre:
We could do the wait on the different places to avoid a generic piece
like haproxy having to wait for a specific component like the apiserver,
but we are already writing specific components in its configuration, and
a future reordering of states could trigger this error again.

So, when we kill haproxy, wait for it to be serving requests again before
continuing with the next state.

On the 2 to 3 upgrade this was causing a failure because right after
restarting haproxy we were trying to drain the node. Since we run this
operation on the very same machine that is being targeted, this `kubectl`
command cannot reach the apiserver (because haproxy is still initializing),
causing the whole update orchestration to fail.

Fixes: bsc#1109661
(cherry picked from commit 95c1980e99e0e2a9787caab02b86056db8e199c0)

Backport of https://github.com/kubic-project/salt/pull/664

### DIFF
--- a/salt/haproxy/init.sls
+++ b/salt/haproxy/init.sls
@@ -91,11 +91,31 @@ haproxy-restart:
 # The admin node is serving the internal API with the pillars. Wait for it to come back
 # before going on with the orchestration/highstates.
 wait-for-haproxy:
-  caasp_http.wait_for_successful_query:
+  caasp_retriable.retry:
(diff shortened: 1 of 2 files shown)


jordimassaguerpla:
@inercia : Is this still on hold?

inercia:
> @inercia : Is this still on hold?

@jordimassaguerpla no :-)

PR #666, opened 2018-09-25
Configure aggregation layer (bsc#1108195)

lcavajani:
This PR configures the aggregation by adding a specific certificate and the required flags

Signed-off-by: Ludovic Cavajani <lcavajani@suse.com>

### DIFF
--- a/salt/kube-apiserver/apiserver.jinja
+++ b/salt/kube-apiserver/apiserver.jinja
@@ -54,8 +54,8 @@ KUBE_API_ARGS="--advertise-address={{ salt.caasp_net.get_primary_ip() }} \
                --requestheader-group-headers=X-Remote-Group \
                --requestheader-extra-headers-prefix=X-Remote-Extra \
                --requestheader-client-ca-file={{ pillar['ssl']['ca_file'] }} \
-               --proxy-client-cert-file={{ pillar['ssl']['kube_apiserver_aggregator_crt'] }} \
-               --proxy-client-key-file={{ pillar['ssl']['kube_apiserver_aggregator_key'] }} \
+               --proxy-client-cert-file={{ pillar['ssl']['kube_apiserver_proxy_client_crt'] }} \
+               --proxy-client-key-file={{ pillar['ssl']['kube_apiserver_proxy_client_key'] }} \
                --storage-backend={{ pillar['api']['etcd_version'] }} \
                --storage-media-type=application/json \
                --service-account-key-file={{ pillar['paths']['service_account_key'] }} \
--- a/salt/kube-apiserver/apiserver.jinja
+++ b/salt/kube-apiserver/apiserver.jinja
@@ -54,6 +54,8 @@ KUBE_API_ARGS="--advertise-address={{ salt.caasp_net.get_primary_ip() }} \
                --requestheader-group-headers=X-Remote-Group \
                --requestheader-extra-headers-prefix=X-Remote-Extra \
                --requestheader-client-ca-file={{ pillar['ssl']['ca_file'] }} \
+               --proxy-client-cert-file={{ pillar['ssl']['kube_apiserver_aggregator_crt'] }} \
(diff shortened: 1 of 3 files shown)


salt/kube-apiserver/apiserver.jinja
```diff
@@ -54,6 +54,8 @@ KUBE_API_ARGS="--advertise-address={{ salt.caasp_net.get_primary_ip() }} \
                --requestheader-group-headers=X-Remote-Group \
                --requestheader-extra-headers-prefix=X-Remote-Extra \
                --requestheader-client-ca-file={{ pillar['ssl']['ca_file'] }} \
+               --proxy-client-cert-file={{ pillar['ssl']['kube_apiserver_aggregator_crt'] }} \
+               --proxy-client-key-file={{ pillar['ssl']['kube_apiserver_aggregator_key'] }} \
```
kiall: These flags, and the certs pointed to by them, are used for more than just the apiserver aggregator. Naming them like this will be confusing at best!

On top of that, theres some discussion in https://github.com/kubic-project/salt/pull/555 - which also sets these flags about if we should really be generating these certs vs asking the customer to provide them, given they are used when making requests to external non-caasp systems, and the CaaSP CA shouldn't have to be trusted by external systems.

At the very least, the pillar names, filenames, etc should remove the "apiserver-aggregator" references since they are inaccurate and will be v.confusing down the line.

lcavajani: Could we use a generated one in case the customer does not provide one?

mjura:
[Bug 1108195 - Aggregation layer needs configuration for kube-metrics-server](https://trello.com/c/jrg8P69h/1170-bug-1108195-aggregation-layer-needs-configuration-for-kube-metrics-server)

salt/kube-apiserver/init.sls
```diff
@@ -12,6 +12,13 @@ include:
          cn = grains['nodename'],
          o = pillar['certificate_information']['subject_properties']['O']) }}
```
mjura: You could add here and generate certs only if doesn't exist yet or wasn't provided by user:
```
{% if not ( pillar['ssl']['kube_apiserver_aggregator_crt'] or pillar['ssl']['kube_apiserver_aggregator_key'] ) %}
```

salt/kube-apiserver/apiserver.jinja
mjura: yes, I think that this is good idea +1

lcavajani: So I propose this:

1. Rename kube_apiserver_aggregator to kube_apiserver_proxy_client so we fit with the flag name
2. As we speak, it is not possible to provide the certificates in Velum so in a first time, we could generate self-signed certificates.
3. Once the customer can provide his own certificates, maybe this can be introduced in [555](https://github.com/kubic-project/salt/pull/555)

Starting with self-signed certs allows us to enable pod/nodes metrics, hence pod auto-scaling feature which is the main reason why customers need aggregation layer.

kiall: > So I propose this:
> 
> 1. Rename kube_apiserver_aggregator to kube_apiserver_proxy_client so we fit with the flag name
> 2. As we speak, it is not possible to provide the certificates in Velum so in a first time, we could generate self-signed certificates.
> 3. Once the customer can provide his own certificates we can switch the certificates, maybe this can be introduced in [555](https://github.com/kubic-project/salt/pull/555)
> 
> Starting with self-signed certs allows us to enable pod/nodes metrics, hence pod auto-scaling feature which is the main reason why customers need aggregation layer.

+1

lcavajani:
Does it need to be backported in release-3.0 branch ?

MaximilianMeister:
> Does it need to be backported in release-3.0 branch ?

yes please, so we can close the bug

PR #667, opened 2018-09-25
Add support for OIDC connectors to dex configmap

dannysauer:
This complements the Velum PR at kubic-project/velum#665 by adding OIDC connector listing support to salt.

NOTE: Do not merge until kubic-project/velum#665 is merged, though, as it will cause orchestration to fail if the new database table (created in that PR) does not exist.  Marking as Work In Progress, since cross-project PR dependencies isn't a thing in GitHub yet. :)


dannysauer:
Rebased on current master yesterday; I guess that means it'll need re-reviewed.

PR #669, opened 2018-10-09
Do not expect masters to always need to be updated

bergmannf:
If the masters already updated, but workers failed to update this state will
not have any minions to run on and fail if 'execpt_minions: false' is not set.


bergmannf:
@jordimassaguerpla bsc# is now in the commit-message for this PR as well.

PR #672, opened 2018-10-12
Fix bsc#1111168: Do not expect masters to always need to be updated

MaximilianMeister:
If the masters already updated, but workers failed to update this state will
not have any minions to run on and fail if 'execpt_minions: false' is not set.

Follow up of https://github.com/kubic-project/salt/pull/669

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -119,6 +119,7 @@ pre-orchestration-migration:
   salt.state:
     - tgt: '{{ is_updateable_node_tgt }}'
     - tgt_type: compound
+    - expect_minions: false
     - batch: 3
     - sls:
       - migrations.2-3.cni.pre-orchestration


salt/orch/update.sls
```diff
@@ -119,6 +119,7 @@ pre-orchestration-migration:
   salt.state:
     - tgt: '{{ is_updateable_node_tgt }}'
     - tgt_type: compound
+    - expect_minions: false
```
inercia: Looking at the final expression for the `target`:
``` 
is_updateable_node_tgt = '( ' + 'G@tx_update_reboot_needed:true' + ' and ' + \
  'not L@' + nodes_down|join(',') + \
  ' and G@roles:kube-master' + ' ) or ( ' + \
  'G@tx_update_reboot_needed:true' + ' and ' + \
  'not L@' + nodes_down|join(',') + \
  ' and G@roles:kube-minion' + ' )'
```
So, in theory, the `expect_minions` should not be necessary as the `target` is already excluding the `nodes_down`. So maybe the problem is in the `target` because, when computed, this expression is empty.

MalloZup:
any update on this PR? can we close it?

MalloZup:
closing ! outdated, feel free to reopen in case contrary

PR #678, opened 2018-10-30
Add a whitelist for returned events so we only save events that we care about

ereslibre:
- [x] Verify that everything is working as expected

### DIFF
--- a/config/master.d/50-master.conf
+++ b/config/master.d/50-master.conf
@@ -9,6 +9,13 @@ event_return: mysql
 presence_events: True
 timeout: 20
 
+event_return_whitelist:
+  - salt/auth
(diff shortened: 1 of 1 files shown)


nanoscopic:
This change breaks initial node acceptance. The message from salt that is logged of a node being acceptance is not whitelisted, so velum does not detect that a node has been accepted and you cannot continue the install process within Velum. The message type for that needs to be determined and added to this whitelist before further testing can occur.

ereslibre:
I'm trying a slightly different approach. Will report back.

ereslibre:
The other approach was more intrusive than this one. I have added `salt/auth` (we can only discriminate by tags in this approach) to the allow list.

PR #681, opened 2018-11-06
Enable `ExperimentalCriticalPodAnnotation` feature gate in the kubelet

ereslibre:
Fixes: bsc#1114812


ereslibre:
Removing WIP flag. It was there because I couldn't verify that it fixed the problem, but this should be safe to merge in any case.

PR #688, opened 2018-11-15
Fix bsc#1116005: Dex pods should run only on master nodes

emilianolangella:
Dex pods should run only on master nodes due to firewall/security policies that could be applied to workers nodes.

### DIFF
--- a/salt/addons/dex/manifests/20-deployment.yaml
+++ b/salt/addons/dex/manifests/20-deployment.yaml
@@ -49,6 +49,13 @@ spec:
                   values:
                   - dex
               topologyKey: "kubernetes.io/hostname"
+      # ensure dex pods are running only on master nodes
+        nodeAffinity:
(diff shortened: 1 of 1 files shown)


stefsuse:
please include bug ref 1116005

MaximilianMeister:
> please include bug ref 1116005

@emilianolangella oh yes, forgot to mention, for bugfixes we need to include a tag in the commit message, so it will land in the package's changelog. this will then be picked up by openbuildservice's issue tracker.

just mention this: `bsc#1116005` somewhere in the commit message

MaximilianMeister:
Just checked the [first CI run](http://jenkins.caasp.suse.net/blue/organizations/jenkins/salt.integration/detail/PR-688/1) and no dex was running on any worker. We might need to do some manual testing though, but this can be done by QA or QAM later

PR #693, opened 2018-11-29
Workaround bsc#1116933: remove /var/lib/container on crio installation

bergmannf:
This will remove the complete `/var/lib/containers` folder on every update as long as the `crio` version is `1.9.*` - once the crio version does not match `1.9.*` it will no longer remove the directory.


ereslibre:
Backport merged, merging.

PR #698, opened 2018-12-04
Add a dummy state to not have an empty state in an orchestration

bergmannf:
This is a workaround for https://github.com/saltstack/salt/issues/14553 when
upgrading crio 1.9 to 1.10.


MalloZup:
@ereslibre do we still need this? in that case merge it

PR #702, opened 2018-12-21
[WIP] Use the registries operator to handle custom certificates for registries

ereslibre:
Depends on https://github.com/kubic-project/velum/pull/706


ereslibre:
The point behind this work is to work on operators for the new stack, but at the same time, being able to "backport" these operators to the old CaaSP; this would mean treat operators as we treat other addons (such as `tiller` or `kubedns`).

This would help us in that the operator is the one actually managing the logic of handling these resources, while salt would just submit the resources (the YAML documents themselves). This would allow us for an easier integration with the new pieces on the old world, without having to introduce many changes on the way we manage everything. It would also allow us to incrementally reduce the salt code as soon as those operators are ready and we deploy the operators on the cluster (old caasp), so they can manage the resources created by salt (that come from the UI).

The neat part IMO is that salt would be a simple mapping from database -> YAML documents [external API] submitted to the cluster, and then operators handling these documents and performing changes to the cluster as we do in Kubic today. I see this is a good balance between the old and new worlds, and how we can start "backporting" operators to the old world, increasingly removing logic from salt.

MalloZup:
@ereslibre  thx for info :love_letter:  i get anyways inof today, thx :dragon:

salt/addons/registries-operator/apply.rb.jinja
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env ruby
```
inercia: Ruby???

salt/addons/registries-operator/init.sls
```diff
@@ -0,0 +1,28 @@
...
+    - mode:     0600
+    - template: jinja
+    - makedirs: true
+
+{{ kubectl_apply_dir_template("salt://addons/registries-operator/manifests/",
+                              "/etc/kubernetes/addons/registries-operator/") }}
+
+{% set certs = salt.caasp_docker.get_registries_certs(salt.caasp_pillar.get('registries', [])) %}
+
+apply registries operator resources:
```
inercia: Even when they are valid names in Salt, I think we jhave never used this kind of ids for Salt: we have always followed the `thihs-is-an-id` convention.

salt/addons/registries-operator/apply.rb.jinja
ereslibre: Not anymore :)

salt/_modules/caasp_kubernetes_secrets.py
```diff
@@ -0,0 +1,35 @@
...
+    return prefix + sha1(content).hexdigest()
+
+
+def present_with_contents(secret_key, secret_contents, secret_name_prefix=None, secret_name=None, namespace="kube-system"):
+    secret_name = name_by_content(secret_contents) if secret_name is None else secret_name
+    if secret_name_prefix is not None:
+        secret_name = secret_name_prefix + secret_name
+    secret_file_path = temp_file_with_contents(secret_contents)
+    kubectl('create secret generic %s --from-file=%s=%s' % (secret_name, secret_key, secret_file_path), namespace=namespace)
+    os.remove(secret_file_path)
```
inercia: Instead of using this pattern for creating a temporary file, I think it would be better to have a `kubectl_apply` function that could accept maybe two different parameters, `file=` and `string=`. Then, when using a `string`, we could use the better Python idiom:

```Python
with tempfile.TemporaryFile() as fp:
    fp.write("the manifest contents")
    # run the kubectl apply now
```

salt/_modules/caasp_kubernetes_crds.py
```diff
@@ -0,0 +1,62 @@
...
+    ret, _, _ = kubectl('delete -f -', stdin=serialize(custom_resource, default_flow_style=False), namespace=namespace)
+    return ret == 0
+
+
+def fetch_current_custom_resources(custom_resource_name, namespace='kube-system'):
+    _, current_resources, _ = kubectl('get %s' % custom_resource_name, namespace=namespace, output='yaml')
+    return deserialize(current_resources)
+
+
+def reconcile_desired_resources(custom_resource_name, namespace, desired_resources=[]):
```
inercia: I think this function tries to be too smart. I would mimic many other Salt modules and have a `caasp_kubernetes_crd.absent` (equivalent to [`file.missing`](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html#salt.states.file.absent)) and `caasp_kubernetes_crd.managed` (equivalent to [`file.managed`](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html#salt.states.file.absent)). Then we could skip all this code for detecting the current CRD and if we want to delete it or not...

salt/_states/caasp_kubernetes_crds.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import absolute_import
+
+
+def reconcile(name, namespace='kube-system', desired_resources=[]):
```
inercia: I think this tries to be too smart too. A `caasp_kubernetes_secret.absent` and `caasp_kubernetes_secret.managed` would be more explicit and clear in my view.

salt/_modules/caasp_kubernetes_secrets.py
ereslibre: Oh, do you suggest to use `kubectl apply` to create the secrets too? We can do it that way, yes. In that case I could reuse the `caasp_kubernetes_kubectl.kubectl` logic that doesn't use a temporary file but `stdin` instead.

salt/_modules/caasp_kubernetes_crds.py
ereslibre: I agree it tries to be too smart, however I don't see a way around it. When a user modifies the settings on the UI and we run an orchestration we know what settings we want, and for removing the extra settings we need to reconcile somehow with the real state of the cluster and that needs to happen somewhere. Am I missing an alternative approach?

inercia: I think we have two cases:

1) the _cleanup_ of registries not present in the pillar: for each registry obtained with `kubectl get registries`, if it is not managed by us, do a `caasp_kubernetes.crd.absent` (a `kubectl delete` under the hood). This could be done with a Jinja loop.
2) the `add/modify` registry. It would be handled by `kubectl apply`s (I think it doesn't really matter if we are creating the CR for the first time or if we are modifying the resource: the k8s API will deal with updating it if that is the case). We could iterate for all the certificates in `get_certs()` and  `caasp_kubernetes.crd.manage` each one. This could be done with a Jinja loop.

ereslibre: Okay, so moving a bit of the logic to the rendering. I'm not sure I'm very happy with that, because with the current solution the SLS look simpler and for future operators we will have less replicated logic in the templates and will be rendering less clutter.

If this was a state/module for others to consume I'd agree present/absent is the way to go, but since this is for us and for making it easier to integrate this pattern in future operators for this mixed world I think I prefer to have simpler SLS and the big chunk of the logic in the python module.

I don't oppose to what you propose either.

inercia: I agree it would mean more code on the SLSs, but (and this is a completely personal view) I think it would be better because

1) _explicit is better than implicit_, and it would be clearer what the code does.
2) I think it would follow more Salt's philosophy, where things are declared as _present_, _missing_ and so on instead of using _deus ex machina_ states that reconcile desired things. 

But, as I said, this is my personal take on this...

salt/_modules/caasp_kubernetes_crds.py
```diff
@@ -0,0 +1,62 @@
...
+
+from salt.serializers.yaml import serialize, deserialize
+
+
+def __virtual__():
+    return "caasp_kubernetes_crds"
+
+
+def kubectl(command, **kwargs):
+    return __salt__['caasp_kubernetes_kubectl.kubectl'](command, **kwargs)
```
flavio: I don't know if `caasp_kubernetes_kubectl.kubectl` is already doing that, but maybe it would be a good idea to log eventual errors

salt/_modules/caasp_docker.py
```diff
@@ -27,17 +27,18 @@ def _get_hostname_and_port(url, default_port=None):
 
 def get_registries_certs(lst, default_port=5000):
     '''
-    Given a list of "valid" items, return a dictionay of
```
MalloZup: :rofl:

salt/_modules/caasp_kubernetes_crds.py
MalloZup: i think personally that the saltstack way is to have `managed`, `absent` etc, logic.

I think yop, we we will have more rendundant code in SLS but if we gan simplicity is better to have it :rocket:

MalloZup:
This GitHub PR is unactive since 30. Is this GitHub PR still needed? Please close or update it accordingly.
         This reminder is autogenerated by https://github.com/MalloZup/blacktango

PR #682, opened 2018-11-06
[3.0] Enable `ExperimentalCriticalPodAnnotation` feature gate in the kubelet

ereslibre:
Fixes: bsc#1114812
Fixes: bsc#1123650
(cherry picked from commit 7776b4b237d6bb3776a579d74e8294017b82f1a8)

Backport of https://github.com/kubic-project/salt/pull/681

PR #708, opened 2019-01-22
[k8s 1.13][bsc#1122783] update critical pod configuration

MaximilianMeister:
https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/

bsc#1122783

### DIFF
--- a/salt/addons/dex/manifests/20-deployment.yaml
+++ b/salt/addons/dex/manifests/20-deployment.yaml
@@ -19,7 +19,6 @@ spec:
       labels:
         app: dex
       annotations:
-        scheduler.alpha.kubernetes.io/critical-pod: ''
         # Kubernetes will not restart dex when the configmap or secret changes, and
         # dex will not notice anything has been changed either. By storing the checksum
         # within an annotation, we force Kubernetes to perform the rolling restart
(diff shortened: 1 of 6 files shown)


MaximilianMeister:
related https://github.com/kubic-project/caasp-container-manifests/pull/212

MaximilianMeister:
This can only be merged when we updated to kubernetes 1.11

ereslibre:
Needs a rebase on top of #710 for passing salt.tests.

ereslibre:
We would have to check, but the previous annotation wasn't actually used, and I'm not sure about the new annotation. I saw sometimes ago that these patches were necessary:

- https://github.com/kubic-project/salt/pull/681
- https://github.com/kubic-project/salt/pull/682

Since `ExperimentalCriticalPodAnnotation` is `false` by default as per https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/.

MaximilianMeister:
> We would have to check, but the previous annotation wasn't actually used, and I'm not sure about the new annotation. I saw sometimes ago that these patches were necessary:
> 
> * #681
> * #682
> 
> Since `ExperimentalCriticalPodAnnotation` is `false` by default as per https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/.

good catch. this seems very likely. i think the best for now is to get those PR's merged

MaximilianMeister:
now that we are on 1.13 we need to consider this

lcavajani:
@MaximilianMeister `kube-flannel` pods would also require this annotation

MaximilianMeister:
> `kube-flannel` pods would also require this annotation

@lcavajani thanks for the hint. updated

PR #709, opened 2019-01-24
[bsc#1116049] don't fail timing out on a drain

MaximilianMeister:
we want to ensure that the platform 'survives' even if an application
is not 'drainable'

it already happened that some applications are getting stuck in a
termination process and won't recover anymore. This can happen for
various reasons (volume not detachable, podDisruptionBudget ...)

as we are using the drain during every update, we are likely to run
into this issue multiple times per cluster

### DIFF
--- a/salt/kubelet/stop.sls
+++ b/salt/kubelet/stop.sls
@@ -10,7 +10,12 @@ include:
 drain-kubelet:
   cmd.run:
     - name: |
+        # give some time to drain the node, if it times out continue to ensure the update will continue
+        # this can cause downtime of applications, so ideally an application is drainable within the drain-timeout
+        # bsc#1116049
         kubectl --request-timeout=1m --kubeconfig={{ pillar['paths']['kubeconfig'] }} drain {{ grains['nodename'] }} --force --delete-local-data=true --ignore-daemonsets --timeout={{ pillar['kubelet']['drain-timeout'] }}s
+    - check_cmd:
+      - /bin/true
     - require:
       - file: {{ pillar['paths']['kubeconfig'] }}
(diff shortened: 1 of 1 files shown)


salt/kubelet/stop.sls
```diff
@@ -10,7 +10,9 @@ include:
 drain-kubelet:
   cmd.run:
     - name: |
-        kubectl --request-timeout=1m --kubeconfig={{ pillar['paths']['kubeconfig'] }} drain {{ grains['nodename'] }} --force --delete-local-data=true --ignore-daemonsets --timeout={{ pillar['kubelet']['drain-timeout'] }}s
+        # give some time to drain the node, if it times out continue to ensure the update will continue
+        # this can cause downtime of applications, so ideally an application is drainable within the drain-timeout
+        kubectl --request-timeout=1m --kubeconfig={{ pillar['paths']['kubeconfig'] }} drain {{ grains['nodename'] }} --force --delete-local-data=true --ignore-daemonsets --timeout={{ pillar['kubelet']['drain-timeout'] }}s || :
```
ereslibre: I think it would be more clear if we readd the:

```
    - check_cmd:
      - /bin/true
```

that was here before.

ereslibre:
Needs a rebase on top of https://github.com/kubic-project/salt/pull/710 for passing `salt.tests`.

salt/kubelet/stop.sls
MaximilianMeister: updated

MaximilianMeister:
> Needs a rebase on top of #710 for passing `salt.tests`.

done

PR #713, opened 2019-01-30
[k8s 1.13] Adapt configs to 1.13

MaximilianMeister:
kubelet config and flannel config needed some tweak

### DIFF
--- a/salt/cni/kube-flannel.yaml.jinja
+++ b/salt/cni/kube-flannel.yaml.jinja
@@ -152,8 +152,7 @@ spec:
       tolerations:
         # Allow the pod to run on the master.  This is required for
         # the master to communicate with pods.
-        - key: node-role.kubernetes.io/master
-          operator: Exists
+        - operator: Exists
           effect: NoSchedule
         # Mark the pod as a critical add-on for rescheduling.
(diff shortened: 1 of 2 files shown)


MaximilianMeister:
tested with k8s `1.13.2` - bootstrap was successful

salt/cni/kube-flannel.yaml.jinja
```diff
@@ -152,8 +152,7 @@ spec:
       tolerations:
         # Allow the pod to run on the master.  This is required for
         # the master to communicate with pods.
-        - key: node-role.kubernetes.io/master
-          operator: Exists
+        - operator: Exists
```
ereslibre: What about the rest of the manifests that were using this as well?

`tiller`, `dex`, `kubedns`... also `haproxy`, I wonder why if it's a static manifest :/

There are some results if you grep by `node-role.kubernetes.io/master`

The change in flannel upstream happened because it didn't work out of the box in Kubernetes 1.12, so their change was related to being able to tolerate everything:

- https://github.com/coreos/flannel/pull/1045
- https://github.com/coreos/flannel/issues/1044#issuecomment-425519107

I'm fine with this change, but I'll include it in the rest of the components as well.

PR #718, opened 2019-02-01
Added instructions for running unit tests

mssola:


MalloZup:
@mssola  i think  we should use tox for that maybe?

mssola:
> @mssola i think we should use tox for that maybe? it is on the top

I had no idea about this `tox` tool :smile: Updated it, hope it makes sense.

PR #719, opened 2019-02-04
[3.0][bsc#1116049] don't fail timing out on a drain

MaximilianMeister:
we want to ensure that the platform 'survives' even if an application
is not 'drainable'

it already happened that some applications are getting stuck in a
termination process and won't recover anymore. This can happen for
various reasons (volume not detachable, podDisruptionBudget ...)

as we are using the drain during every update, we are likely to run
into this issue multiple times per cluster

(cherry picked from commit 73b87c4f7bac62678a009aff73257aa4399b8d5c)

Backport of https://github.com/kubic-project/salt/pull/709

### DIFF
--- a/salt/kubelet/stop.sls
+++ b/salt/kubelet/stop.sls
@@ -10,7 +10,12 @@ include:
 drain-kubelet:
   cmd.run:
     - name: |
+        # give some time to drain the node, if it times out continue to ensure the update will continue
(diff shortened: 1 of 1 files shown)


jordimassaguerpla:
This seems a trivial fix plus it was already reviewed and accepted in master.

PR #720, opened 2019-02-04
caasp_filters: properly parse IP addresses

mssola:
As it appears, for Python 2 Salt is using a custom module for providing the
ipaddress from Python 3. This implementation turned out to have
bugs (e.g. accepting something70.domain.net as a valid IPv4 record). I've
re-implemented the is_ip* functions so they are using socket.inet_pton instead,
which should suffice in our use case.

bsc#1123291

### DIFF
--- a/salt/_modules/caasp_filters.py
+++ b/salt/_modules/caasp_filters.py
@@ -1,6 +1,12 @@
 from __future__ import absolute_import
 
-from salt._compat import ipaddress
+import socket
+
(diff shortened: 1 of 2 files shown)


mssola:
This is a backport of #717

jordimassaguerpla:
This was validated by support, so I am approving it.

jordimassaguerpla:
Failing tests don't seem related to this fix.

PR #721, opened 2019-02-04
User cert support

dannysauer:
All the salt backend changes to make Velum, Dex, and kube-API (the external-facing services) support user-provided Certs (and their corresponding CA).

There is a related PR in kubic-project/caasp-container-manifests#213 which is necessary for this to work properly.

Marked work-in-progress since it's still not completely verified as working.

### DIFF
--- a/salt/_macros/certs.jinja
+++ b/salt/_macros/certs.jinja
@@ -41,7 +41,7 @@
 
 #####################################################################
 
-{% macro certs(name, crt, key, cn='', o='', extra_alt_names=None) -%}
+{% macro certs(name, crt, key, cn='', o='', extra_alt_names=None, bundle=False) -%}
 
 {%- if extra_alt_names == None -%}
   {#- calculate automatically the altNames depending on the role -#}
@@ -77,7 +77,7 @@
     - Email: {{ pillar['certificate_information']['subject_properties']['Email']|yaml_dquote }}
     - GN: {{ pillar['certificate_information']['subject_properties']['GN']|yaml_dquote }}
     - L: {{ pillar['certificate_information']['subject_properties']['L']|yaml_dquote }}
-    {# "system:{{ name }}" is a kubernetes specific role identifying a {{ name }} in th system #}
+    {# "system:{{ name }}" is a kubernetes specific role identifying a {{ name }} in the system #}
   {%- if o %}
     - O: {{ o|yaml_dquote }}
   {%- else -%}
@@ -103,9 +103,13 @@
       - sls: crypto
       - x509: {{ key }}
 
-{{ crt }}-bundle:
+{{ bundle or crt + '-bundle' }}:
   file.managed:
+{% if bundle %}
+    - name: {{ bundle }}
+{% else %}
     - name: /etc/pki/private/{{ name }}-bundle.pem
+{% endif %}
     - source: salt://cert/cert-bundle.jinja
     - template: jinja
     - user: root
@@ -120,3 +124,38 @@
         - x509: {{ key }}
 
 {%- endmacro %}
+
+# externally-sourced certs; content in a pillar
+{% macro external_pillar_certs(crt, crt_pillar, key, key_pillar, bundle=False) -%}
+{{ crt }}:
+  file.managed:
+    - user:  root
+    - group: root
+    - mode: 0644
+    - contents_pillar: {{ crt_pillar }}
+
+{{ key }}:
+  file.managed:
+    - user:  root
+    - group: root
+    - mode: 0444
+    - contents_pillar: {{ key_pillar }}
+
+{% if bundle %}
+# TODO: move this out to a macro so it's not repeated in here?
+{{ bundle }}:
+  file.managed:
+    - source: salt://cert/cert-bundle.jinja
+    - template: jinja
+    - user: root
+    - group: root
+    - mode: 600
+    - makedirs: True
+    - context:
+        certificate: {{ crt }}
+        key:         {{ key }}
+    - require: # always verify -- in case this file was deleted/altered
+        - file: {{ crt }}
+        - file: {{ key }}
+{% endif %}
+{% endmacro %}
--- a/salt/addons/dex/init.sls
+++ b/salt/addons/dex/init.sls
@@ -3,8 +3,23 @@ include:
   - kubectl-config
   - kube-apiserver
 
+{% if salt.caasp_pillar.get('external_cert:dex:cert', False)
+  and salt.caasp_pillar.get('external_cert:dex:key',  False)
+%}
+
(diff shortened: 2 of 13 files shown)


NunoFilipeSantos:
@dannysauer Please review this PR since the tests are failing, this is not done.

dannysauer:
Rebased on current master, since it's still working fine on my system.  I guess let's see if that triggers a successful CI run...

dannysauer:
🤦‍♂️ 
This fails because it changes to using the system cert store for Velum cert validation instead of the SUSE cert.  That means it's never going to pass CI until kubic-project/caasp-container-manifests#213 is merged first.  I'm not sure why it took several weeks to realize something I told people before I opened the PR...  Sigh.

dannysauer:
The new CI failure happens because teh haproxy config is rendered invalid after a reboot.  I didn't test rebooting because, well, why would simply rebooting change stuff? 😕 Whoops.  Should be fixed by kubic-project/caasp-container-manifests#224

dannysauer:
Planning to rebase on master and force-push, as CI is now failing at rebase. :)

dannysauer:
Squashed commits down in preparation for merge.

salt/_macros/certs.jinja
```diff
@@ -41,7 +41,7 @@
 
 #####################################################################
 
-{% macro certs(name, crt, key, cn='', o='', extra_alt_names=None) -%}
+{% macro certs(name, crt, key, cn='', o='', extra_alt_names=None, bundle=False) -%}
```
nanoscopic: It isn't clear offhand here that bundle is the full path to the bundle file. It is though, as it is used that way below. Perhaps 'bundle_file' would be more clear?

I do get though that crt and key are also full pathnames, so it does match the existing style for the macro.

salt/_macros/certs.jinja
```diff
@@ -103,9 +103,13 @@
       - sls: crypto
       - x509: {{ key }}
 
-{{ crt }}-bundle:
+{{ bundle or crt + '-bundle' }}:
```
nanoscopic: It's a little strange here that the name of the managed file block is not a path in all cases. When bundle is available, it is. Despite that, when it is bundle, the 'name' attribute is still specified. Perhaps leave name blank since the name of the block is the filename in that case?

salt/_macros/certs.jinja
```diff
@@ -120,3 +124,38 @@
...
+
+{{ key }}:
+  file.managed:
+    - user:  root
+    - group: root
+    - mode: 0444
+    - contents_pillar: {{ key_pillar }}
+
+{% if bundle %}
+# TODO: move this out to a macro so it's not repeated in here?
```
nanoscopic: While I agree with the nicety of 'hey someday we might want to make this cleaner' I doubt it will ever happen, so you might as well remove this TODO. Or leave it in -shrug-

salt/addons/dex/init.sls
```diff
@@ -3,8 +3,24 @@ include:
...
+
+{% from '_macros/certs.jinja' import external_pillar_certs with context %}
+
+{{ external_pillar_certs(
+      pillar['ssl']['dex_crt'],
+      'external_cert:dex:cert',
+      pillar['ssl']['dex_key'],
+      'external_cert:dex:key',
+      False
+
```
nanoscopic: Does this extra blank line belong?

salt/haproxy/haproxy.yaml.jinja
```diff
@@ -55,10 +61,17 @@ spec:
         path: {{ pillar['ssl']['ca_file'] }}
 {% endif %}
         type: FileOrCreate
+    - name: ca-bundle-dir
+      hostPath:
+        path: {{ salt['file.dirname']( pillar['ssl']['sys_ca_bundle'] ) }}
```
nanoscopic: For consistency, shouldn't type: Directory be here after path?

salt/orch/refresh-certs.sls
```diff
@@ -0,0 +1,83 @@
...
+    - tgt: {{ updates_all_target + ' and P@roles:(kube-(master|minion))' }}
+    - tgt_type: compound
+    - kwarg:
+        queue: True
+    - sls:
+      - addons.dex
+    - require:
+      - salt: update_ca_list
+
+#update_certs_kubeapi:
```
nanoscopic: How come this commented out code is left here?

salt/_macros/certs.jinja
dannysauer: This was done for compatability with the legacy code, so as to reduce the number of changes necessary to make this work. If there were going to be a future salt cleanup, it'd be worth making prettier - but there won't be, so I felt like minimizing the change footprint was worth the slightly ugly code. :/

salt/_macros/certs.jinja
dannysauer: That consistency is is precisely why it's "bundle". :)

salt/haproxy/haproxy.yaml.jinja
dannysauer: The default behavior is directory if the "thing" doesn't exist, and they same type of it does already exist. I think the typical expectation is that Directory be left out, but I do like being explicit. If it's changed here, though, it'll also need changed for the other directories in the other static pod definitions, both here and in container manifests (which are already merged and kicked upstream).

I don't really like "ugh, that's a pain" as reasoning, but I think "three PRs and some history mangling" means that might be where I land here. I'm now disappointed in myself. Thanks a lot. :D

salt/addons/dex/init.sls
dannysauer: No, that's a missed edit after pasting stuff in from another block. Whoops.

salt/orch/refresh-certs.sls
dannysauer: Because this orchestration isn't finished or used, and I forgot about cleaning it up. :)

I guess I'll add a comment to that effect for future developers - who likely won't exist. Or maybe I'll finish the last part of the orchestration so it'll be useful. I'll have to think about it. I think it's all haproxy and dex, so I suspect it'll work to just remove this block.

salt/addons/dex/init.sls
dannysauer: Removed.

salt/orch/refresh-certs.sls
dannysauer: Unused code removed; better commented useful code.

salt/_macros/certs.jinja
dannysauer: Leaving it in there in case we end up in a weird dystopian future where this codebase is the best path forward. ;)

PR #724, opened 2019-02-08
bsc#1117942: fixes for volume plugins

inercia:
* Use a writable directory for volume plugins
* Use the same volumes plugins directory for the `controller-manager` and the `kubelet`

bsc#1117942

### DIFF
--- a/pillar/top.sls
+++ b/pillar/top.sls
@@ -10,6 +10,7 @@ base:
     - docker
     - registries
     - schedule
+    - volume
   'roles:ca':
     - match: grain
     - ca
--- a/pillar/volume.sls
+++ b/pillar/volume.sls
@@ -1,0 +1,5 @@
+
+volume:
+  dirs:
+    bin: '/var/lib/kubelet/plugins'
+
--- a/salt/kube-controller-manager/controller-manager.jinja
+++ b/salt/kube-controller-manager/controller-manager.jinja
@@ -18,10 +18,13 @@ KUBE_CONTROLLER_MANAGER_ARGS="\
     --allocate-node-cidrs=true \
     --node-cidr-mask-size={{ pillar['cluster_cidr_len'] }} \
     --root-ca-file={{ pillar['ssl']['ca_file'] }} \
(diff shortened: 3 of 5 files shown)


lcavajani:
@MalloZup this is not specific to cloud, it is used to configure [flexvolume drivers.](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-storage/flexvolume.md)

I'm currently testing it

lcavajani:
@r0ckarong we've fixed this issue but we're not using the default path because it points to a RO mountpoint, so this would need a point in documentation just to explain what is the good path ;)

* Install FlexVolume drivers on CaaSP

A FlexVolume driver is an external (out of tree) driver usually provided by a specific Vendor.

These volume drivers have to be installed on each node (worker/master) in a Kubernetes cluster.

On SUSE CaaSP the path to install these drivers is `/var/lib/kubelet/plugins`.


Doc: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-storage/flexvolume.md

lcavajani:
@inercia we need to backport this one ;)

r0ckarong:
@lcavajani Is the documentation you linked applicable to CaaSP like that? If not could someone please provide the CaaSP specific steps.

Also, which version of CaaSP will receive this functionality? Is this going to be backported to v3?

lcavajani:
@r0ckarong this will be backported to v3 yes.

There is no specific configuration to do for CaaSP afterwards.

These drivers are usually deployed with DaemonSets, what the user should do is to change the value of the flexolume dir path in the manifests which is usually in the form of an environment variable (e.g `FLEXVOLUME_DIR_PATH` for rook [1])

[1] https://rook.io/docs/rook/v0.9/flexvolume.html#configuring-the-flexvolume-path

r0ckarong:
> so this would need a point in documentation just to explain what is the good path ;)

What is this path?

lcavajani:
@r0ckarong the path is `/var/lib/kubelet/plugins`

PR #727, opened 2019-02-12
[bsc#1125095] deployment timeout not correctly configured

MaximilianMeister:
instead of setting the timeout we were setting the retries which
causes the timeout to be prolonged too much


MalloZup:
( bad numerical typo :smile_cat: )

MalloZup:
@MaximilianMeister backport it if needed thx

PR #728, opened 2019-02-12
Sort nodenames first when generating /etc/hosts entries

ereslibre:
Salt will pick the first name on the current default interface to determine
the hostname of the machine. Since we are sorting with all entries for each
machine there's a high change that a salt minion id will win the first position,
affecting certain grains that we use to determine the hostname of the node.

With this change we are not only sorting alphabetically by the name of each column
in each entry, but also adding weight to each column on each entry. This means that
the nodename will always win, as its primary weight will be lower than the rest,
while the second sorting strategy is purely alphabetical.

Fixes: bsc#1117339

### DIFF
--- a/salt/_modules/caasp_hosts.py
+++ b/salt/_modules/caasp_hosts.py
@@ -68,10 +68,22 @@ def __virtual__():
     return "caasp_hosts"
 
 
-def _concat(lst1, lst2):
+# returns a list resulting of appending `lst2` to `lst1`, removing duplicates on
+# both lists (not preserving order on any of them) and removing empty elements on
+# the result. The result will be sorted as well
+def _sorted_append(lst1, lst2):
     res = list(set(lst1) | set(lst2))  # join both lists (without dups)
     res = [x for x in res if x]  # remove empty strings
     res.sort()  # sort the result (for determinism)
+    return res
+
+
+# returns a list resulting of prepending `lst2` to `lst1`, not removing
+# duplicates on any of both lists (preserving order on both of them) and
+# removing empty elements on the result
+def _unsorted_prepend(lst1, lst2):
+    res = lst2 + lst1  # unsorted prepend of lst2 in lst1
+    res = [x for x in res if x]  # remove empty strings
     return res
 
 
@@ -141,7 +153,7 @@ def _load_hosts_file(hosts, filename, marker_start=None, marker_end=None):
 
 # add a (list of) name(s) to a (maybe existing) IP
 # it will remove duplicates, sort names, etc...
-def _add_names(hosts, ips, names):
+def _add_names(hosts, ips, names, insert_fun=_sorted_append):
     if not isinstance(names, list):
         names = [names]
     if not isinstance(ips, list):
@@ -150,20 +162,25 @@ def _add_names(hosts, ips, names):
     for ip in ips:
         __utils__['caasp_log.debug']('hosts: adding %s -> %s', ip, names)
         if ip not in hosts:
-            hosts[ip] = _concat([], names)
+            hosts[ip] = insert_fun([], names)
         else:
-            hosts[ip] = _concat(hosts[ip], names)
+            hosts[ip] = insert_fun(hosts[ip], names)
 
 
-def _add_names_for(hosts, nodes_dict, infra_domain):
+def _add_names_for(hosts, nodes_dict, infra_domain, insert_fun=_sorted_append):
     for id, ifaces in nodes_dict.items():
         ip = __salt__['caasp_net.get_primary_ip'](host=id, ifaces=ifaces)
         if ip:
-            _add_names(hosts, ip, [id, id + '.' + infra_domain])
+            _add_names(hosts, ip, [id, id + '.' + infra_domain], insert_fun)
 
+
+def _add_nodenames_for(hosts, nodes_dict, infra_domain):
+    for id, ifaces in nodes_dict.items():
+        ip = __salt__['caasp_net.get_primary_ip'](host=id, ifaces=ifaces)
+        if ip:
             nodename = __salt__['caasp_net.get_nodename'](host=id)
             if nodename:
-                _add_names(hosts, ip, [nodename, nodename + '.' + infra_domain])
+                _add_names(hosts, ip, [nodename, nodename + '.' + infra_domain], _unsorted_prepend)
 
 
 # note regarding node removals:
@@ -271,12 +288,15 @@ def managed(name=HOSTS_FILE,
     def get_with_expr(expr):
         return __salt__['caasp_nodes.get_with_expr'](expr, grain='network.interfaces')
 
+    admin_nodes = admin_nodes or get_with_expr(ADMIN_EXPR)
+    master_nodes = master_nodes or get_with_expr(MASTER_EXPR)
+    worker_nodes = worker_nodes or get_with_expr(WORKER_EXPR)
+    other_nodes = other_nodes or get_with_expr(OTHER_EXPR)
+
     # add all the entries
     try:
-        _add_names_for(hosts, admin_nodes or get_with_expr(ADMIN_EXPR), infra_domain)
-        _add_names_for(hosts, master_nodes or get_with_expr(MASTER_EXPR), infra_domain)
-        _add_names_for(hosts, worker_nodes or get_with_expr(WORKER_EXPR), infra_domain)
-        _add_names_for(hosts, other_nodes or get_with_expr(OTHER_EXPR), infra_domain)
+        for nodes in [admin_nodes, master_nodes, worker_nodes, other_nodes]:
+            _add_names_for(hosts, nodes, infra_domain)
     except Exception as e:
         raise EtcHostsRuntimeException(
             'Could not add entries for roles in /etc/hosts: {}'.format(e))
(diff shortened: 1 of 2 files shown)


mssola:
Could you add some tests to check that the resulting /etc/hosts entries are sorted as expected?

ereslibre:
Tests added. @inercia PTAL when you can, thanks!

salt/_modules/caasp_hosts.py
```diff
@@ -155,15 +154,34 @@ def _add_names(hosts, ips, names):
...
+        names = [names]
+    if not isinstance(ips, list):
+        ips = [ips]
+
+    for ip in ips:
+        __utils__['caasp_log.debug']('hosts: adding %s -> %s', ip, names)
+        if ip not in hosts:
+            hosts[ip] = names
+        else:
+            hosts[ip] = names + hosts[ip]
```
inercia:
```Python
def _add_names(hosts, ips, names, fun):
    if not isinstance(names, list):
        names = [names]
    if not isinstance(ips, list):
        ips = [ips]
    for ip in ips:
        __utils__['caasp_log.debug']('hosts: adding %s -> %s', ip, names)
        if ip not in hosts:
            hosts[ip] = fun([], names)
        else:
            hosts[ip] = fun(hosts[ip], names)
```

so we could define 

```Python
def _prepend(lst1, lst2):
    res = list(set(lst2) | set(lst1))  # join both lists (without dups)
    res = [x for x in res if x]  # remove empty strings
    return res
```

(we should even calle these functions `prepend` and `append` for coherency)

use `_add_names(hosts, ip, [id, id + '.' + infra_domain], _concat)` or `_add_names(hosts, ip, [nodename, nodename + '.' + infra_domain], _prepend)`...

salt/_modules/caasp_hosts.py
```diff
@@ -308,19 +331,32 @@ def get_with_expr(expr):
...
+    # sort the names for determinism
+    for ip, names in hosts.items():
+        names.sort()
+
+    # prepend the nodenames at the beginning of each entry
+    try:
+        _add_nodenames_for(hosts, admin_nodes, infra_domain)
+        _add_nodenames_for(hosts, master_nodes, infra_domain)
+        _add_nodenames_for(hosts, worker_nodes, infra_domain)
+        _add_nodenames_for(hosts, other_nodes, infra_domain)
```
inercia: I would apply the same pattern:

```suggestion
        for x in [admin_nodes, master_nodes, worker_nodes, other_nodes]:
            _add_nodenames_for(hosts, x, infra_domain)
```

salt/_modules/caasp_hosts.py
```diff
@@ -155,15 +154,34 @@ def _add_names(hosts, ips, names):
...
 
+
+def _add_nodenames_for(hosts, nodes_dict, infra_domain):
+    for id, ifaces in nodes_dict.items():
+        ip = __salt__['caasp_net.get_primary_ip'](host=id, ifaces=ifaces)
+        if ip:
             nodename = __salt__['caasp_net.get_nodename'](host=id)
             if nodename:
-                _add_names(hosts, ip, [nodename, nodename + '.' + infra_domain])
+                _prepend_names(hosts, ip, [nodename, nodename + '.' + infra_domain])
```
inercia: `_add_nodenames_for` and `_add_names_for` are basically the same function: they only differ in the way they obtain the IP. I would define only one function:

```Python
def _get_primary_ip(id, ifaces):
    return __salt__['caasp_net.get_primary_ip'](host=id, ifaces=ifaces)

def _get_nodename(id, ifaces):
    return __salt__['caasp_net.get_nodename'](host=id, ifaces=ifaces)

def _add_names_for(hosts, nodes_dict, infra_domain, getter):
    for id, ifaces in nodes_dict.items():
        ip = getter(host=id, ifaces=ifaces)
        if ip:
            _add_names(hosts, ip, [id, id + '.' + infra_domain])
```

salt/_modules/caasp_hosts.py
```diff
@@ -271,12 +289,17 @@ def fqdn(name):
     def get_with_expr(expr):
         return __salt__['caasp_nodes.get_with_expr'](expr, grain='network.interfaces')
 
+    admin_nodes = admin_nodes or get_with_expr(ADMIN_EXPR)
+    master_nodes = master_nodes or get_with_expr(MASTER_EXPR)
+    worker_nodes = worker_nodes or get_with_expr(WORKER_EXPR)
+    other_nodes = other_nodes or get_with_expr(OTHER_EXPR)
+
```
inercia: Using a common `add_names_for()` you could define here a new function:

```Python
    def _add_names_for_all(getter):
        for x in [admin_nodes, master_nodes, worker_nodes, other_nodes]:
            _add_names_for(hosts, x, infra_domain, getter)
```

so you could use a `_add_names_for_all(_get_primary_ip)` and a  `_add_names_for_all(_get_nodename)` later on...

salt/_modules/caasp_hosts.py
ereslibre: `res = list(set(lst2) | set(lst1))  # join both lists (without dups)` is itself problematic, because the order preservation in `lst2` is not guaranteed. The fact that we are using `set` to remove duplicates also makes the order inside `lst2` subject to change.

For our specific usecase it's not an issue, because `nodename` and `nodename.infra.caasp.local` is the only usage of this, and order will always come with `nodename` first (the desired result), but this is based on language implementation of `set` and this can break again. OTOH it's weird to change `_prepend` to not use a `set` while `_concat` or the first argument of `_prepend` does.

salt/_modules/caasp_hosts.py
```diff
@@ -141,7 +153,7 @@ def _load_hosts_file(hosts, filename, marker_start=None, marker_end=None):
 
 # add a (list of) name(s) to a (maybe existing) IP
 # it will remove duplicates, sort names, etc...
-def _add_names(hosts, ips, names):
+def _add_names(hosts, ips, names, insert_fun=_sorted_append):
```
MalloZup: Can we name it fn or func for function? Or are we inserting happyess? 😁

salt/_modules/caasp_hosts.py
```diff
@@ -150,20 +162,25 @@ def _add_names(hosts, ips, names):
...
         if ip:
-            _add_names(hosts, ip, [id, id + '.' + infra_domain])
+            _add_names(hosts, ip, [id, id + '.' + infra_domain], insert_fun)
 
+
+def _add_nodenames_for(hosts, nodes_dict, infra_domain):
+    for id, ifaces in nodes_dict.items():
+        ip = __salt__['caasp_net.get_primary_ip'](host=id, ifaces=ifaces)
+        if ip:
             nodename = __salt__['caasp_net.get_nodename'](host=id)
```
MalloZup: We could use guard clause here for simplifying, if not ip return nothing then continue (pseudo code)

inercia: I've never been really fond of those guard clauses. I prefer a style where you can clearly see that you are going deep down the rabit hole... but it is my personal taste.

salt/_modules/caasp_hosts.py
```diff
@@ -68,13 +68,25 @@ def __virtual__():
...
     return res
 
 
+# returns a list resulting of prepending `lst2` to `lst1`, not removing
+# duplicates on any of both lists (preserving order on both of them) and
+# removing empty elements on the result
+def _unsorted_prepend(lst1, lst2):
+    res = lst2 + lst1  # unsorted prepend of lst2 in lst1
+    res = [x for x in res if x]  # remove empty strings
+    return res
```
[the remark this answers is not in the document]
inercia: Yes, we could even reduce the whole function to `return [x for x in lst2 + lst1 if x]`, but I think we are leaving it as it is because it makes the comments more clear...

salt/_modules/caasp_hosts.py
MalloZup: yop i think is personal taste. To me guard clause simplify the `context` infos and nested things and you see basically when you go down the func you follow  the main logic, otherwise you exit with guardClause

salt/_modules/caasp_hosts.py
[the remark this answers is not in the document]
MalloZup: yop, i think 1 liner is to complicate but imho 2 lines could be ok  but is ok also 3

salt/_modules/caasp_hosts.py
```diff
@@ -68,13 +68,25 @@ def __virtual__():
     return "caasp_hosts"
 
 
-def _concat(lst1, lst2):
+# returns a list resulting of appending `lst2` to `lst1`, removing duplicates on
```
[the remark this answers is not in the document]
ereslibre: This is implementation detail and I'm following the pattern in the file for this kind of functions, the only one with proper documentation on this module is `managed`.

We can open a PR to fix all comments on this module later.

salt/_modules/caasp_hosts.py
ereslibre: Let's leave as it is right now, following the style of other parts of the file. If we change this to a guard clause other parts of this file would need adapting as well and it's out of the scope of this PR.

salt/_modules/caasp_hosts.py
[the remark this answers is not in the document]
ereslibre: Leaving to keep the style with the previous code, refactoring more things here is out of the scope for this PR.

MalloZup:
:rocket:

PR #733, opened 2019-02-14
Fix python3 iteration over dictionary.

bergmannf:
In python3 python prevents modifying the dictionary that is iterated over.

Instead of modifying the initial dictionary it will be copied first.

Also `iteritems` is no longer available in `python3`, so the code should use the `six` function instead to be cross-version compatible.

### DIFF
--- a/salt/_modules/caasp_net.py
+++ b/salt/_modules/caasp_net.py
@@ -23,10 +23,9 @@ def get_hosts():
     :return: Dict(ip_address -> [aliases])
     """
     hosts = __salt__['hosts.list_hosts']()  # type: dict
-    for key in hosts.keys():
-        if key.startswith("comment-"):
-            hosts.pop(key)
-    return hosts
+    # Python3 (correctly) prevents modifying a dictionary we are iterating
+    # over, so we have to copy it first, so we can remove comment entries.
+    return {k:v for k,v in hosts.items() if not k.startswith("comment-")}
 
 
 def get_aliases(hostname):
--- a/salt/_modules/caasp_net.py
+++ b/salt/_modules/caasp_net.py
@@ -4,6 +4,7 @@
 #
 
 from __future__ import absolute_import
+from salt.ext.six import iteritems
 
 
 DEFAULT_INTERFACE = 'eth0'
@@ -25,7 +26,7 @@ def get_hosts():
     hosts = __salt__['hosts.list_hosts']()  # type: dict
     # Python3 (correctly) prevents modifying a dictionary we are iterating
     # over, so we have to copy it first, so we can remove comment entries.
-    return {k:v for k,v in hosts.items() if not k.startswith("comment-")}
+    return {k: v for k, v in hosts.items() if not k.startswith("comment-")}
 
 
 def get_aliases(hostname):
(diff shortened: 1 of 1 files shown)


salt/_modules/caasp_net.py
```diff
@@ -23,10 +23,11 @@ def get_hosts():
     :return: Dict(ip_address -> [aliases])
     """
     hosts = __salt__['hosts.list_hosts']()  # type: dict
+    copy = hosts.copy()
```
MalloZup: `hosts_copy`?

MalloZup: can we add a comment here above : 
`In python3 python prevents modifying the dictionary that is iterated over.`

bergmannf:
@MalloZup Thanks for the review - updated the PR

MalloZup:
failure on CI was: 
`Error: Error applying plan:


1 error(s) occurred:


* libvirt_domain.admin: 1 error(s) occurred:


* libvirt_domain.admin: timeout while waiting for state to become 'all-addresses-obtained' (last state: 'waiting-addresses', timeout: 5m0s)


Terraform does not automatically rollback in the face of errors.
```

This is a terraform race condition afaik. retry

**Comment by MalloZup** [comment]
tox failure: 
```
+ tox --notest -e tests-salt-2018.3.0-py34

tests-salt-2018.3.0-py34 create: /srv/jenkins/workspace/salt.tests_PR-733/.tox/tests-salt-2018.3.0-py34

tests-salt-2018.3.0-py34 installdeps: git+https://github.com/opensuse/salt@openSUSE-2018.3.0, mock==2.0.0, nose==1.3.7

wrapper script does not seem to be touching the log file in /srv/jenkins/workspace/salt.tests_PR-733@tmp/durable-729564b2

(JENKINS-48300: if on an extremely laggy filesystem, consider -Dorg.jenkinsci.plugins.durabletask.BourneShellScript.HEARTBEAT_CHECK_INTERVAL=86400)

script returned exit code -1
```

rescheduling

**Comment by hwoarang** [comment]
@MalloZup please document these random failures on the confluence page

File: salt/_modules/caasp_net.py [hunk 2]
```diff
@@ -23,10 +24,13 @@ def get_hosts():
...
     hosts = __salt__['hosts.list_hosts']()  # type: dict
+    # Python3 (correctly) prevents modifying a dictionary we are iterating
+    # over, so we have to copy it first, so we can remove comment entries.
+    hosts_copy = hosts.copy()
     for key in hosts.keys():
         if key.startswith("comment-"):
-            hosts.pop(key)
-    return hosts
+            hosts_copy.pop(key)
+    return hosts_copy
```
inercia: Taking into account we are using Pyrthon 3, I think it would be much better to use a dictionary comprehension:

```suggestion
    return {k:v for k,v in hosts.items() if not k.startswith("comment-")}
```

bergmannf: I will change to this approach - especially since I don't intend to backport these changes, as they are not needed for `python2`.

bergmannf:
We might have to wait with merging this - i was told that this error actually also pops up on v3, so I'm investigating that.

I would be surprised if it does, but if so the code needs to be backwards compatible.

MalloZup:
@bergmannf take a look flake

bergmannf:
@MalloZup Flake8 errors are fixed - little pointer: the CI does *not* tell me what's wrong so I have to run it locally - maybe it shouldn't just pipe the output to an unreachable `xml` file?

MalloZup:
@bergmannf  you need to checkout the result in Webui and you have the results

PR #737, opened 2019-02-18
use caasp v4 images from SUSE Registry

jordimassaguerpla:


salt/_modules/caasp_registry.py
```diff
@@ -72,4 +72,4 @@ def base_image_url():
     if _use_registry_images():
         return "{0}/{1}".format(_registry(), _namespace())
     else:
-        return "sles12"
+        return "caasp/v4"
```
MaximilianMeister: the `v4` should ideally be returned by `caasp_version()` to avoid hardcoding this over every release

jordimassaguerpla: can I run "caasp_version" on a running cluster to see what it returns? Or can I get the value of a grain?

jordimassaguerpla:
I just realized this is wrong. We need to set "use_registry" to true into the c-c-m package , instead

PR #741, opened 2019-02-19
Jenkinsfile: Switch to dynamic library fetching and drop branch

hwoarang:
Instead of having the library hardcoded to Jenkins master, we can fetch
it dynamically. We also drop the usage of library branches since it does
not make sense to maintain such a thing in the CI. The master branch
should be able to handle both development and release branches.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,5 @@
-def targetBranch = env.getEnvironment().get('CHANGE_TARGET', env.BRANCH_NAME)
-
-library "kubic-jenkins-library@${targetBranch}"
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -1,5 +1,5 @@
-def targetBranch = env.getEnvironment().get('CHANGE_TARGET', env.BRANCH_NAME)
-
-library "kubic-jenkins-library@${targetBranch}"
+library identifier: "kubic-jenkins-library@master", retriever: modernSCM(
```
ereslibre: We should probably keep `targetBranch` variable and use:

`
library identifier: "kubic-jenkins-library@${targetBranch}"
`

?

PR #745, opened 2019-02-20
replace assert with try/raise

MalloZup:
Is better to use a try/raise mechanism then a generic ex without a
failure reason. This can be usefull for tracing down debug issues


inercia:
Take into account that assertions are not used with the same intention as `try: except:` blocks.

While `try: except:` blocks are used for catching errors that can be treated, silenced, or wrapped in better errors, assertions are intended to be used _when developing_ as a firewall that stops the execution of a program before things go utterly wrong and the bug becomes too difficult to be found. 

Assertions make a good team with unit tests, where we can check things but we want to quickly stop execution when something looks suspicious. That is the reason why assertions are just disabled when the Python interpreter is run with _optimizations_ (`python -O`).

inercia:
I would close this PR. The intention of that assertion is to be used at development, expressing a _you can be sure the file exists at this point_, like an invariant.

PR #746, opened 2019-02-20
Synchronize everything before starting an orchestration.

inercia:
* Synchronize everything before starting an orchestration.
* Replace all the `mine.get` calls by the more compact `get_with_expr` function.

bsc#1124784

### DIFF
--- a/salt/_modules/caasp_orch.py
+++ b/salt/_modules/caasp_orch.py
@@ -1,0 +1,16 @@
+from __future__ import absolute_import
+
+
+def __virtual__():
+    return "caasp_orch"
+
+
+def sync_all():
+    '''
+    Syncronize everything before starting a new orchestration
+    '''
+    __utils__['caasp_log.debug']('orch: refreshing all')
+    __salt__['saltutil.sync_all'](refresh=True)
+
+    __utils__['caasp_log.debug']('orch: synchronizing the mine')
+    __salt__['saltutil.runner']('mine.update', tgt='*', clear=True)
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -1,3 +1,6 @@
+{#- Make sure we start with an updated mine #}
+{%- set _ = salt.caasp_orch.sync_all() %}
+
 {#- Get a list of nodes seem to be down or unresponsive #}
 {#- This sends a "are you still there?" message to all #}
 {#- the nodes and wait for a response, so it takes some time. #}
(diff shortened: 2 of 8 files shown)


salt/orch/update.sls
```diff
@@ -350,8 +354,8 @@ all-workers-2.0-pre-clean-shutdown:
 {%- endfor %}
 # END NOTE: Remove me for 4.0
 
-{%- set workers = salt.saltutil.runner('mine.get', tgt=is_updateable_worker_tgt, fun='network.interfaces', tgt_type='compound') %}
-{%- for worker_id, ip in workers.items() %}
+{%- set workers = salt.caasp_nodes.get_with_expr(is_updateable_worker_tgt) %}
```
MalloZup: :+1:

salt/_modules/caasp_orch.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def sync_all():
+    '''
+    Syncronize everything before starting a new orchestration
+    '''
+    __utils__['caasp_log.debug']('orch: synchronizing the mine')
+    __salt__['saltutil.runner']('mine.update', tgt='*', clear=True)
+
+    # TODO: maybe sync modules and so...
```
MalloZup: i was thinking, how much is the effort to sync all modules? this could spare some error msg we have in minions, we could maybe do it in the same pr if it is 3 lines

salt/_modules/caasp_orch.py
```diff
@@ -0,0 +1,15 @@
...
+
+def __virtual__():
+    return "caasp_orch"
+
+
+def sync_all():
+    '''
+    Syncronize everything before starting a new orchestration
+    '''
+    __utils__['caasp_log.debug']('orch: synchronizing the mine')
```
MalloZup: we can refactor this 2 lines  in a `def sync_mine()` function

MalloZup: and calling it in the `all`

salt/_modules/caasp_orch.py
```diff
@@ -0,0 +1,15 @@
...
+    __salt__['saltutil.runner']('mine.update', tgt='*', clear=True)
```
MalloZup: can we do here:
```python
     try:
    __salt__['saltutil.runner']('mine.update', tgt='*', clear=True)
            except Exception as exc:
                WARN: YOUR BANK ACCOUNT HAS BEEN HACKED 

```

inercia:
> looks good to me, however i'm not honestly a fun of todo in Code and i think we could maybe try thing to fix it. Or we open an isssue in github

Well, I personally do not see anything bad in TODOs. It flags something that can be fixed in the future but it is not urgent, so maybe in a second refinement of the code we could investigate about adding some other syncs.

> Anoter thing i was looking for , was about `{%- set _ = salt.caasp_orch.sync_all() %}` the `set` method, i find it really cool your approach, but i wask asking me if it is the only method used for salt for doing such things? ( it is more open question)

I haven't found any other way of doing this. Function calls in Jinja templates must return a value that is going tot be printed, and we don't want to print anything, so I think this is the only solution...

salt/_modules/caasp_orch.py
inercia: I think there is no need to catch exceptions in every single line of our code. The standard exception handler is fine here.

salt/_modules/caasp_orch.py
inercia: I wouldn't refactor so early this code. If we need to have a `sync_mine()` at some moment we can refactor out this code and do as you suggest, but until then I think the only case we have is a `sync_all()`.

PR #753, opened 2019-02-27
Make sure we refresh modules synchronously after syncing everything

inercia:
bsc#1124784

### DIFF
--- a/salt/_modules/caasp_orch.py
+++ b/salt/_modules/caasp_orch.py
@@ -12,5 +12,8 @@ def sync_all():
     __utils__['caasp_log.debug']('orch: refreshing all')
     __salt__['saltutil.sync_all'](refresh=True)
 
+    # make sure we refresh modules synchronously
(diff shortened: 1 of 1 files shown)


MalloZup:
after ci green we can merge this

salt/_modules/caasp_orch.py
```diff
@@ -12,5 +12,8 @@ def sync_all():
     __utils__['caasp_log.debug']('orch: refreshing all')
     __salt__['saltutil.sync_all'](refresh=True)
 
+    # make sure we refresh modules synchronously
+    __salt__['saltutil.refresh_modules'](async=False)
```
MalloZup: why not  async? or better question what does bring syncro here?

MalloZup:
@inercia i don't think the PR is responsable of the failure. Can you retrigger? What was your finding about the RED?

MalloZup:
Red was race condition on ci

PR #755, opened 2019-03-04
Remove custom module caasp_http

MalloZup:
This module was required so we can pass options like `http_request_timeout` to
the final `query` call without having to modify the minion global configuration.

This has been implemented in Salt 2018.3.0 version:
https://github.com/saltstack/salt/commit/f72c2820f2d68f50e0919327a677f4ad8a5584b5

So we can remove our custom module.


MalloZup:
backporting to 30 thx

PR #756, opened 2019-03-04
Don't always reenable the transaction-update timer on update

mssola:
When executing the update orchestration, it's not always desired to reenable the
transactional-update. Moreover, sometimes the timer was on a weird state if the
update failed. This commit refines the update orchestration following this
logic:

1. Check first which nodes have explicitely disabled the timer, so they are not
   reenabled in the end.
2. Disable the timer at the very beginning.
3. If this is a migration, reenable it always.
4. Otherwise, reenable the timer only for these nodes which did not explicitely
   disable it before the update.

Last but not least, I've also tweaked the `transactional-update/init.sls` file
so the timer is not touched by this file when updating.

bsc#1113518

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -47,6 +47,11 @@
 {%- set migrated_nodes_tgt = is_responsive_node_tgt + ' and G@migration_in_progress:true' %}
 {% endif %}
 
+{#- Fetch all the nodes which have explicitely disabled the timer. #}
+{#- These won't be reenabled at the end of thi orchestration #}
+{%- set enabled_timer_nodes_list = salt.caasp_nodes.get_with_expected_ret_value('saltutil.cmd', True, tgt='*', fun='service.enabled', arg=['transactional-update.timer']) %}
+{%- do salt.caasp_log.debug('Nodes the timer of which should be reenabled at the end: %s', enabled_timer_nodes_list|join(',')) %}
+
 # Ensure all nodes with updates are marked as upgrading. This will reduce the time window in which
 # the update-etc-hosts orchestration can run in between machine restarts.
 set-progress-grain:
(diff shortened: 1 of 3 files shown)


MalloZup:
@mssola  about point 3) `If this is a migration, reenable it always.` since Caasp v4 doesn't support if migration from v3 to v4, i'm thinking we should simplify it here and add that logic in v3

mssola:
> @mssola about point 3) If this is a migration, reenable it always. since Caasp v4 doesn't support if migration from v3 to v4, i'm thinking we should simplify it here and add that logic in v3

I'm not sure I understand. This was already in place before my change, actually. My change tries to preserve the past behavior (point 3), and adds points 1, 2 and 4.

MalloZup:
> > @mssola about point 3) If this is a migration, reenable it always. since Caasp v4 doesn't support if migration from v3 to v4, i'm thinking we should simplify it here and add that logic in v3
> 
> I'm not sure I understand. This was already in place before my change, actually. My change tries to preserve the past behavior (point 3), and adds points 1, 2 and 4.

ah ok i was thinking you added (point 3) ok thx for comment. ( i could also not found the point 3 in the pr :smile: )
thx

mssola:
This PR will also need to be backported into 3.0 once this is merged (since the bug originally targets 3.0)

salt/orch/update.sls
```diff
@@ -593,6 +610,17 @@ unset-{{ grain }}-grain:
     - require:
       - reenable-transactional-update-timer
 {%- endfor %}
+{%- elif enabled_timer_nodes_list|length > 0 %}
+reenable-transactional-update-timer:
+  salt.function:
+    - tgt: '{{ enabled_timer_nodes_list|join(",") }}'
```
inercia: Why can't we just enable the transactional update timer in all the nodes in the cluster?

mssola: That's one of the requirements of the bug: we should only reenable the timer where it was already enabled before the update. In other words, if the admin manually disabled the timer, we shouldn't reenable it after the update.

PR #759, opened 2019-03-05
[bsc#1127326] Enable Kubelet token authentication

lcavajani:
Enabling Token authentication allows external
components like monitoring tools to fetch information
against kubelet on port 10250 over HTTPS and authenticate
with a Bearer token.

Fixes bsc#1127326

Signed-off-by: Ludovic Cavajani <lcavajani@suse.com>
(cherry picked from commit c4005e5288496ade4f5e50392538356ea5b40061)

### DIFF
--- a/salt/kubelet/kubelet-config.jinja
+++ b/salt/kubelet/kubelet-config.jinja
@@ -12,7 +12,7 @@ authentication:
   x509:
     clientCAFile: "{{ pillar['ssl']['ca_file'] }}"
   webhook:
-    enabled: false
+    enabled: true
     cacheTTL: 2m0s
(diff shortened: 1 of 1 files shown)


salt/kubelet/kubelet-config.jinja
```diff
@@ -12,7 +12,7 @@ authentication:
   x509:
     clientCAFile: "{{ pillar['ssl']['ca_file'] }}"
   webhook:
-    enabled: false
+    enabled: true
```
MalloZup: before we merge this, it might be worth to consider also if enabling `webhook` hasn't have security consequences.

PR #760, opened 2019-03-05
don't sync module during update. (fix  -  bsc#1125444)

MalloZup:
This pr is used for seeing how far we can go with the removal.

This call cause race condition and issues

### DIFF
--- a/salt/_modules/caasp_orch.py
+++ b/salt/_modules/caasp_orch.py
@@ -13,7 +13,8 @@ def sync_all():
     __salt__['saltutil.sync_all'](refresh=True)
 
     # make sure we refresh modules synchronously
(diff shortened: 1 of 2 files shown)


MalloZup:
Wow this passed, 😮😮😮😮🎸🎸🎶🎸🎶🎸

MalloZup:
so 1 run was ok, i want maybe to test with sles15.

MalloZup:
@inercia before merging this, i would like to see if it works on sles15 see ( https://github.com/SUSE/caasp-jenkins-library/pull/287)

MalloZup:
i'm testing it with sles15, let's wait with that. currenlty i need to use the right image

MalloZup:
i'm closing this since we will have soonish the patch for fixing the race condition merged

PR #767, opened 2019-03-11
Cleanup unneeded reactor code

MalloZup:
we shouldn't sync grains/update mine during start minions. ( a part of slowing down orchestration and minion we don't needed this)

### DIFF
--- a/config/master.d/50-reactor.conf
+++ b/config/master.d/50-reactor.conf
@@ -1,10 +1,7 @@
 reactor:
   - 'salt/presence/change':
-    - /usr/share/salt/kubernetes/reactor/presence.sls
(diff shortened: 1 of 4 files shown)


config/master.d/50-reactor.conf
```diff
@@ -1,10 +1,7 @@
 reactor:
   - 'salt/presence/change':
-    - /usr/share/salt/kubernetes/reactor/presence.sls
```
MalloZup: @inercia  i wonder what the empty state should do. imho is ok to remove it

inercia: :+1:

MalloZup:
i will merge this tomorrow so i don't forget to backport.. :grin:

PR #768, opened 2019-03-11
Remove the proxy config files when `proxy:systemwide` is not set

inercia:
Remove the proxy config files when `proxy:systemwide` is not set (or set to False).

[bsc#1116572](https://bugzilla.suse.com/show_bug.cgi?id=1116572)

### DIFF
--- a/salt/proxy/init.sls
+++ b/salt/proxy/init.sls
@@ -1,17 +1,18 @@
-{% if salt.caasp_pillar.get('proxy:systemwide') %}
-  {% set proxy_http  = salt.caasp_pillar.get('proxy:http') %}
-  {% set proxy_https = salt.caasp_pillar.get('proxy:https') %}
+{%- if salt.caasp_pillar.get('proxy:systemwide') %}
+  {%- set proxy_http  = salt.caasp_pillar.get('proxy:http') %}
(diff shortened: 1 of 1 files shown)


salt/proxy/init.sls
```diff
@@ -1,17 +1,18 @@
-{% if salt.caasp_pillar.get('proxy:systemwide') %}
-  {% set proxy_http  = salt.caasp_pillar.get('proxy:http') %}
-  {% set proxy_https = salt.caasp_pillar.get('proxy:https') %}
-
-  {% set no_proxy = [salt.caasp_pillar.get('dashboard'), '.infra.caasp.local', '.cluster.local'] %}
-  {% set extra_no_proxy = salt.caasp_pillar.get('proxy:no_proxy') %}
-  {% if extra_no_proxy %}
-    {% do no_proxy.append(extra_no_proxy) %}
-  {% endif %}
+{%- if salt.caasp_pillar.get('proxy:systemwide') %}
```
MalloZup: why the `-` added in this context

inercia: For removing the superfluous whitespaces in the rendered file: it makes logs really ugly. I know, it is a bit paranoic...

MalloZup: aha ok good :small_airplane:

PR #770, opened 2019-04-01
Adjust timeouts for the salt command and api (bsc1130242)

mjura:
By default, SUSE Manager globally sets timeout and gather_job_timeout to 120 seconds.
So, in the worst case, a Salt call targeting unreachable minions will end up with
240 seconds of waiting until getting a response.


mjura:
More about these timeouts
https://www.suse.com/documentation/suse-manager-3/book_suma_best_practices_3/data/bp_chap_salt_minion_scaleability_timeouts.html

PR #771, opened 2019-04-18
restart kubelet & salt-minion on cert change (bsc#1131568)

dannysauer:
Update kubernetes orchestration to restart kublet and salt-minion processes when the CA store changes.
Also fix cert-specific orchestration

This is "finishing" the User-provided SSL cert feature (SUSE/caasp-playground#170 and SUSE/caasp-playground#171). Once merged, will get added in to SUSE/kubic-salt-security-fixes#44.

### DIFF
--- a/salt/orch/kubernetes.sls
+++ b/salt/orch/kubernetes.sls
@@ -135,6 +135,40 @@ cert-setup:
     - require:
       - update-mine-again
 
+# restart salt minions if cert changes
+{%- set nodes_down = salt.saltutil.runner('manage.down') %}
+{%- if nodes_down|length >= 1 %}
+  {%- set is_responsive_node_tgt = 'not L@' + nodes_down|join(',') %}
+{%- else %}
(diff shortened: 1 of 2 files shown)


salt/orch/kubernetes.sls
```diff
@@ -135,6 +135,40 @@ cert-setup:
...
+    - name: service.restart
+    - arg:
+      - 'salt-minion'
+    - onchanges:
+      - cert-setup
+
+salt-minion-wait-for-start:
+  salt.wait_for_event:
+    # TODO: should this specify node_id instead of '*'?
+    - name: salt/minion/*/start
```
stefsuse: It might be a good idea to iterate over known minion ids, I'm not sure if this will work as expected after a node is removed from the cluster (CI does not include this step)

dannysauer: The ID List just below waits for the list of responsive nodes calculated just above, so it only waits for up nodes. :)

PR #773, opened 2019-08-07
update.sls: start transactional-update timer

jvanz:
In the update process we disable the transactional update timer and re
enabled it after update the machine. However, just re enable it is not
enough. We need to start the timer. Otherwise, the update service will
not run and consequently no updates is installed. This commit adds in
the update.sls a step to start the timer after re enabling it.

bsc#1141675

Signed-off-by: José Guilherme Vanz <jguilhermevanz@suse.com>

### DIFF
--- a/salt/orch/update.sls
+++ b/salt/orch/update.sls
@@ -524,9 +524,10 @@ reenable-transactional-update-timer:
     - tgt: '{{ enabled_timer_nodes_list|join(",") }}'
     - tgt_type: list
     - batch: 3
-    - name: service.enable
+    - name: service.running
     - arg:
         - transactional-update.timer
+        - enable: true
     - require:
       - remove-caasp-fqdn-grain
 {%- endif %}


jvanz:
> There is an identical state already, can we use that and fix the logic check? https://github.com/kubic-project/salt/blob/master/salt/orch/update.sls#L521

Yes, I've submitted the wrong patch. I've updated the branch with other patch. In this patch I did not add a step. I just changed the salt state to `service.running` instead of `service.enable`.

stefsuse:
FYI @ release-eng for backport. 
@ QA please take a look and determine if we need another test, thanks.

jordimassaguerpla:
I don't know about the CI, sorry. Maybe @pablochacin or @tdaines42 can help here.. can you?

jordimassaguerpla:
The CI is broken atm. Let's go ahead and merge this and we will do manual testing with the Maintenance Update.

PR #774, opened 2019-09-13
update.sls: using wrong module to start timer

jvanz:
A previous commit trying to start the transactional-update timer is
wrong. The right module is `service.start`. The current orchestration
step fails to start the timer, but does not fail the whole orchestration.
To fix that, a new step has been added to start the timer after re
enabling it. Unfortunately, the start/enable state functions do not
allow do both steps at once.

Signed-off-by: José Guilherme Vanz <jguilhermevanz@suse.com>


stefsuse (approved):
makes sense to me

